PR #5, opened 2018-10-12
docs/design/cincinnati: fix example config

steveej:
According to the document itself and to the current implementation the
`kind` attribute is required.

/cc @crawford

### DIFF
--- a/docs/design/cincinnati.md
+++ b/docs/design/cincinnati.md
@@ -39,6 +39,7 @@ An example of this metadata document can be seen here:
 
 ```json
 {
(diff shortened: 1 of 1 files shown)


docs/design/cincinnati.md
```diff
@@ -39,6 +39,7 @@ An example of this metadata document can be seen here:
 
 ```json
 {
+	"kind": "cincinnati-metadata-v0",
 	"schema": 1,
```
lucab, line 43: I think you want to remove `schema` then (here and below too).

steveej:
Closing this in favor of #7

Issue #6, opened 2018-10-15
design/cincinnati: clarify some details on the graph endpoint


lucab:
A spec clarification request to @crawford on https://github.com/openshift/cincinnati/blob/95db15a823b231636c7a3aaeef25b6c22e6c9a8b/docs/design/cincinnati.md#graph-api.

The current implementation in `cincinnati` is stricter than the design spec, and these are the points that I'd like to have explicitly noted regarding the JSON API of the `graph` endpoint:
 * is `version` a freeform string or a semver label?
 * is a `""` allowed as a `version` value?
 * must a `version` value be unique in the "nodes" array?
 * is the `payload` field required and is an empty-string allowed there?
 * is the client supposed to pass additional parameters somehow (e.g. as query parameters)?
 * is the edges "empty object for future use" still a thing? (it isn't in example nor in deserializer code)

crawford:
>  * is `version` a freeform string or a semver label?

Right now, it is semver. We might want to consider switching to the RPM versioning at some point.

>  * is a "" allowed as a version value?

No.

>  * must a version value be unique in the "nodes" array?

Yes. The version value is the identifier for a release. That cannot be ambiguous.

>  * is the payload field required and is an empty-string allowed there?

It is required. Right now we allow it to be an empty string, but we might want to disallow that.

>  * is the client supposed to pass additional parameters somehow (e.g. as query parameters)?

Yes, the client can pass query parameters in its request upstream. The contract for those parameters is defined by the client and Policy Engine implementation.

>  * is the edges "empty object for future use" still a thing? (it isn't in example nor in deserializer code)

Ah, I fixed that a while ago. I forgot to remove it from the documentation.

lucab:
Thanks, I'll update the docs with those details then.

Two followups:
 * `version`: it sounds like we may want to define that as a plain "required unique non-empty JSON string" and leave semantic validation (semver, NEVRA or whatever) to higher-level consumers, ack?
 * `payload`: I don't foresee a usecase for an empty string but I think it doesn't hurt allowing that (as it isn't a main key/id anyway)

crawford:
Both of those sound good.

steveej:
> `version`: it sounds like we may want to define that as a plain "required unique non-empty JSON string" and leave semantic validation (semver, NEVRA or whatever) to higher-level consumers, ack?

It seems like we're looking fro a `version_scheme` field or something similar which lets the higher-level consumer specify how they want to deal with versions.

[...]

PR #7, opened 2018-10-31
cincinnati: update for new spec details

lucab:
This updates docs and cincinnati logic (and its consumers too) according to
latest spec clarification from https://github.com/openshift/cincinnati/issues/6. In particular, this relaxes the "version"
field to non-semver keys, but introduces emptiness and uniqueness
check across the "nodes" array.

Closes: https://github.com/openshift/cincinnati/issues/6
Closes: https://github.com/openshift/cincinnati/pull/5

### DIFF
--- a/cincinnati/src/lib.rs
+++ b/cincinnati/src/lib.rs
@@ -15,7 +15,6 @@
 extern crate daggy;
 #[macro_use]
 extern crate failure;
-extern crate semver;
 extern crate serde;
 #[macro_use]
 extern crate serde_derive;
@@ -23,11 +22,10 @@ extern crate serde_derive;
 use daggy::petgraph::visit::{IntoNodeReferences, NodeRef};
 use daggy::{Dag, Walker};
 use failure::Error;
-use semver::Version;
 use serde::de::{self, Deserialize, Deserializer, MapAccess, Visitor};
 use serde::ser::{Serialize, SerializeStruct, Serializer};
 use std::collections::HashMap;
-use std::fmt;
+use std::{collections, fmt};
 
 pub const CONTENT_TYPE: &str = "application/json";
 
@@ -44,7 +42,7 @@ pub enum Release {
 }
 
 impl Release {
-    pub fn version(&self) -> &Version {
+    pub fn version(&self) -> &str {
         match self {
             Release::Abstract(release) => &release.version,
             Release::Concrete(release) => &release.version,
@@ -54,14 +52,14 @@ impl Release {
 
 #[derive(Debug, Deserialize, Serialize)]
 pub struct ConcreteRelease {
-    pub version: Version,
+    pub version: String,
     pub payload: String,
     pub metadata: HashMap<String, String>,
 }
 
 #[derive(Debug, Deserialize, Serialize)]
 pub struct AbstractRelease {
-    pub version: Version,
+    pub version: String,
 }
 
 pub struct ReleaseId(daggy::NodeIndex);
@@ -111,7 +109,7 @@ impl Graph {
         Ok(())
     }
 
-    pub fn find_by_version(&self, version: &Version) -> Option<ReleaseId> {
+    pub fn find_by_version(&self, version: &str) -> Option<ReleaseId> {
         self.dag
             .node_references()
             .find(|nr| nr.weight().version() == version)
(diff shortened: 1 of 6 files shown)


docs/design/cincinnati.md
```diff
@@ -128,10 +129,20 @@ HTTP GET requests are used to fetch the DAG from the Graph API endpoint. Request
...
 
 ### Response ###
 
-The response to the `/v1/graph` endpoint is a JSON representation of the release graph. Each of the releases are represented in an entry in the top-level "nodes" array. Each of these entries includes the release version and any metadata. The transitions between releases are represented as an array in the top-level "edges" array. Each of these arrays has three entries: the index of the starting node, the index of the ending node, and an empty object. This empty object is reserved for future use.
+The response to the `/v1/graph` endpoint is a JSON representation of the release graph. Each of the releases are represented in an entry in the top-level `nodes` array. Each of these entries includes the release version label, a payload identifier and any metadata according to the following schema:
+
+|   Key    | Optional | Description                                                                             |
+|:--------:|:--------:|:----------------------------------------------------------------------------------------|
+| version  | required | the version of the release, as a unique non-empty JSON string                           |
```
steveej: In which context must the string be unique? Also, are we assuming semver?

docs/design/cincinnati.md
```diff
@@ -93,21 +94,21 @@ The storage component is responsible for actually storing and hosting the releas
...
+The process of removing nodes from the graph is not defined by Cincinnati. Instead, it is up to the implementation to decide. For example, when using a Docker v2 Registry, a deleted node could be represented with a special label on the image.
 
 
 ### Policy Engine ###
 
-Policy Engines are in charge of altering a client’s view of the graph by applying a set of filters which are defined within the particular Policy Engine instance. Both the input to and the output from Policy Engines is a graph, allowing multiple Policy Engines to be chained together. The first Policy Engine in a chain will fetch its graph from the Graph Builder and the last Policy Engine in a chain will serve the modified graph to the client.
+Policy Engines are in charge of altering a client'ss view of the graph by applying a set of filters which are defined within the particular Policy Engine instance. Both the input to and the output from Policy Engines is a graph, allowing multiple Policy Engines to be chained together. The first Policy Engine in a chain will fetch its graph from the Graph Builder and the last Policy Engine in a chain will serve the modified graph to the client.
```
steveej: One `s` should do it here

docs/design/cincinnati.md
```diff
@@ -128,10 +129,20 @@ HTTP GET requests are used to fetch the DAG from the Graph API endpoint. Request
...
+The response to the `/v1/graph` endpoint is a JSON representation of the release graph. Each of the releases are represented in an entry in the top-level `nodes` array. Each of these entries includes the release version label, a payload identifier and any metadata according to the following schema:
+
+|   Key    | Optional | Description                                                                             |
+|:--------:|:--------:|:----------------------------------------------------------------------------------------|
+| version  | required | the version of the release, as a unique non-empty JSON string                           |
+| payload  | required | payload identifier, as a JSON string                                                    |
+| metadata | required | an opaque object that allows a release to convey arbitrary information to its consumers |
+
+
```
steveej: Not necessary empty line

cincinnati/src/lib.rs
```diff
@@ -44,7 +42,7 @@ pub enum Release {
 }
 
 impl Release {
-    pub fn version(&self) -> &Version {
+    pub fn version(&self) -> &str {
```
steveej, line 45: Why did you switch from `semver::Version` to `&str` here? Isn't the former much more helpful when comparing versions and such?

steveej (changes requested):
I'm not in favor if dropping the `semver::Version` here and would restrain from loosening up the restrictions if possible. Can't we have semver and the uniqueness and non-emptyness checks at the same time? ;-) 

Please see the few comments on the docs.

---

EDIT:

I noticed that @crawford and @lucab discussed the version loosening in I just noticed that this was discussed beforehand in https://github.com/openshift/cincinnati/issues/6#issuecomment-433952395 and above.

EDIT2:

I'd like to block this PR until https://github.com/openshift/cincinnati/issues/6#issuecomment-434698758 has been discussed.

cincinnati/src/lib.rs
steveej: I just noticed that this was discussed beforehand in https://github.com/openshift/cincinnati/issues/6#issuecomment-433952395.

lucab:
@steveeJ yup. I'll fix the immediate minor things here and then go back to #6 for more context/discussion.

crawford (approved):

steveej:
@lucab is there code depending on this change somewhere for me to have a look at?

lucab:
@steveeJ nope, this was mostly spec clarification.

PR #8, opened 2018-11-02
Add tests for `fetch_releases` pulling test images from quay.io

steveej:
This is in an early stage and experimentally adds a lib/bin split to the graph-builder crate to allow integration testing.

Depends on #4.


steveej:
@crawford I've changed this to use the unit test mechanism instead of the integration test one. This prevents us from having to make everything `pub` and is thus less invasive. We could cherry-pick this into #4 or just merge this as this is properly based on #4 and github will detect that.

steveej:
The tests are part of #4 now.

PR #9, opened 2018-11-08
[deprecated] graph-builder: support image manifest labels

steveej:
Closes #4 (obsoletes the same too)
Closes #8 (obsoletes the same too)


steveej:
All known bugs are resolved and have a test case. The code is also refactored to my own satisfaction.
If that's worth anything from the PR author:
/lgtm

steveej:
This PR is deprecated as we're not interested in the `LABEL`s which are in the manifest, but rather in the mutable label's which are managed via [the quay labels API][1].

[1]: https://docs.quay.io/api/swagger/#!/manifest/listManifestLabels

PR #11, opened 2018-11-12
add tests for `fetch_releases` pulling test images from quay.io

steveej:
Continuation of #8. Depends on #4.


graph-builder/tests/net/quay_credentials.json
```diff
@@ -0,0 +1,9 @@
+{
+  "_comment": "This is specifically generated by steveeJ for quay.io/steveej/cincinnati-test and merely has pull rights",
+  "auths": {
+    "quay.io": {
+      "auth": "c3RldmVlaitjaW5jaW5uYXRpY2k6WkhHTEpXS1pGRkZXNjRWSUgwS0kyUEhHNFEyWFpDRjgyNlBPNTJYUEZUSTg5TU1BSkNCT09ERThGSUpQSFFJWg==",
```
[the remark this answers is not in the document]
steveej: This is a token specifically generated for the cincinnati test repo. I haven't granted it access to any other repo. What do you think is the risk here?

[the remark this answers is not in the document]
riuvshin: I didn;t know what is that token for, but usually that's a not good practice to have any type of credentials in public repos. I would put this in to a vault and inject during CI build.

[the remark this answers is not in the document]
steveej: If we have a CI job available that is definitely better than having it in repo. Let's get the CI going and I'll rebase this PR on top of that, including removing this.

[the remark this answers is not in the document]
steveej: @riuvshin Can we move this into the CI environment somehow or should we leave it in-repo?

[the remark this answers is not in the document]
riuvshin: we can put this into vault and then inject in ci job runtime, we can setup this next week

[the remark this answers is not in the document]
steveej: Sounds good. I'll generate a new robot account for the CI then and we'll just leave this here until then.

graph-builder/src/registry.rs
```diff
@@ -293,3 +293,144 @@ fn assemble_metadata(blob: &[u8], metadata_filename: &str) -> Result<Metadata, E
...
+        use release::MetadataKind::V0;
+        use semver::Version;
+        fn init_logger() {
+            let _ = env_logger::try_init_from_env(env_logger::Env::default());
+        }
+
+        fn expected_releases(source_base: &str, count: usize, start: usize) -> Vec<Release> {
+            let mut releases = Vec::new();
+
+            let mut metadata: std::collections::HashMap<
```
crawford: I'd `use std::collections::HashMap` to simplify this.

graph-builder/src/registry.rs
```diff
@@ -293,3 +293,144 @@ fn assemble_metadata(blob: &[u8], metadata_filename: &str) -> Result<Metadata, E
     }
     .map_err(Into::into)
 }
+
+#[cfg(test)]
+mod tests {
+
+    #[cfg(feature = "test-net")]
+    mod tests_net {
```
crawford: These are integration tests and should be in the `tests` directory instead.

graph-builder/src/release.rs
```diff
@@ -17,9 +17,9 @@ use semver::Version;
 use std::collections::HashMap;
 use std::fmt;
 
-#[derive(Debug, Deserialize, Clone)]
+#[derive(Debug, Deserialize, Clone, PartialEq)]
 pub struct Metadata {
-    kind: MetadataKind,
+    pub kind: MetadataKind,
```
crawford, line 22: I'd prefer to keep this private. Nobody should be looking at this field.

crawford (changes requested):
I'm not really sure how all of this fits together. A readme would be helpful.

graph-builder/src/release.rs
steveej: The issue here is that we can't create an instance of it manually if it's private:

```
error[E0451]: field `kind` of struct `net::quay_io::graph_builder::release::Metadata` is private                                             
  --> graph-builder/tests/net/quay_io/mod.rs:36:17                                                                                           
   |                                                                                                                                         
36 |                 kind: V0,                                                                                                               
   |                 ^^^^^^^^ field `kind` is private
```

What would be the simplest way to solve this? E.g., should we make a `new()` method for constructing one?

graph-builder/src/registry.rs
steveej: @crawford PTAL at the last commit which implements your suggestion. It's quite invasive but that's fine with me.

steveej:
> I'm not really sure how all of this fits together. A readme would be helpful.

I'll deliver the README to all these files here once we've figured out the Rust tests.

graph-builder/src/release.rs
crawford: We should just be able to implement the Default trait.

crawford:
> I'll deliver the README to all these files here once we've figured out the Rust tests.

I'm having trouble reviewing this without an overview. There seems to be a bunch of unrelated things in this PR, which leads me to believe that I don't understand how this all fits together.

graph-builder/src/release.rs
crawford: That's what I was afraid of. Oh well, it's not the end of the world to make that public.

steveej: I implemented my previously mentioned idea with the `new(..)` method. PTAL at https://github.com/openshift/cincinnati/pull/11/commits/4010c8cdf32cedc1e75f62788b4c38a35179cc7e#diff-f0669f981e48dab0c10160890afd503bR34. I don't think it's actually much worse than the `default()` method.

steveej: Discussed this OOB and concluded that consumers of these objects need to be able to read this field too, so making this public is a requirement anyway.

steveej:
Yay the prow test is green :raised_hands:

crawford (approved):

PR #12, opened 2018-11-16
graph-builder: improve logging output

crawford:
Instead of logging the layer contents (which can be very large), it's
more useful to log the digest of the layer. Additionally, if no layers
contain the metadata, it's more clear to express that in the error
message instead of returning the last error (which indicates that a
particular layer didn't have the metadata).

### DIFF
--- a/graph-builder/src/registry.rs
+++ b/graph-builder/src/registry.rs
@@ -196,6 +196,7 @@ fn find_first_release(
     repo: String,
     tag: String,
 ) -> impl Future<Item = Release, Error = Error> {
+    let outer_tag = tag.clone();
     future::loop_fn(
         layer_digests.into_iter().peekable(),
         move |mut layer_digests_iter| {
@@ -250,8 +251,8 @@ fn find_first_release(
                             layer_digests_iter,
                             None,
                             Some(format_err!(
-                                "could not assemble metadata from blob '{:?}': {}",
-                                String::from_utf8_lossy(&blob),
+                                "could not assemble metadata from blob ({}): {}",
+                                layer_digest,
                                 e,
                             )),
                         )),
@@ -261,6 +262,7 @@ fn find_first_release(
         },
     )
     .flatten()
+    .map_err(move |_| format_err!("metadata not found in any layer of tag {}", outer_tag))
 }
 
 #[derive(Debug, Deserialize)]


graph-builder/src/registry.rs
```diff
@@ -190,6 +190,7 @@ fn find_first_release(
     repo: String,
     tag: String,
 ) -> impl futures::future::Future<Item = Release, Error = failure::Error> {
+    let err = format_err!("metadata not found in any layer of tag {}", tag);
```
steveej: It's a bit confusing to read this as the first line in the function.
Can the right value be inlined in the `map_err(..)` call?

steveej (commented):
The rationale behind this was that if something goes wrong here it might be caused by a 4XX response from the registry, so the body contains a message. I agree that this could potentially be spam the log in other cases. Ideally we'd distinguish the error types here but we don't get one from dkregistry either, so I'm fine with the change after the code comment is addressed.

graph-builder/src/registry.rs
crawford: `tag` is moved into the closure below, so we have to clone it before. It's a _little_ easier to do that with the call to `format_err!`, but I could just clone it instead. What do you think?

steveej: I think instead of shadowing the inner error we should return the information we get from inside, similar to:

```rust
diff --git a/graph-builder/src/registry.rs b/graph-builder/src/registry.rs
index c0d6759..6bb46ec 100644
--- a/graph-builder/src/registry.rs
+++ b/graph-builder/src/registry.rs
@@ -206,10 +206,7 @@ fn find_first_release(
                     let no_release_found = future::ok(continue_or_break_loop(
                         layer_digests_iter,
                         None,
-                        Some(format_err!(
-                            "no release found for tag '{}' and no more layers to examine",
-                            tag
-                        )),
+                        Some(format_err!("no more layers to examine for tag '{}'", tag)),
                     ));
                     return Either::A(no_release_found);
                 }
@@ -332,13 +329,10 @@ where
             trace!("Found release '{:?}'", r);
             Break(Ok(r))
         }
-        (_, Some(e)) => {
-            warn!("{}", e);
-            match layer_digests_iter.peek() {
-                Some(_) => Continue(layer_digests_iter),
-                None => Break(Err(e)),
-            }
-        }
+        (_, Some(e)) => match layer_digests_iter.peek() {
+            Some(_) => Continue(layer_digests_iter),
+            None => Break(Err(format_err!("unable to find a release: {}", e))),
+        },
         _ => continue_or_break_loop(
             layer_digests_iter,
             None,
```
*(not tested or rustfmt'd)*

The condition of all layers being exhausted is already covered.  With the removal of that `warn!(..)` we can log the error higher in the call chain.

crawford: Before your patch:

```
DEBUG 2018-11-21T19:54:54Z: graph_builder::graph: Updating graph...
TRACE 2018-11-21T19:54:55Z: graph_builder::registry: processing: openshift/origin-release:v4.0
TRACE 2018-11-21T19:54:55Z: graph_builder::registry: Downloading layer sha256:b5f0369eef1572c8086ad2c8145b2901d3d7b735e2eea14cb628bed698a428e5...
TRACE 2018-11-21T19:54:56Z: graph_builder::registry: v4.0: Looking for cincinnati.json in archive sha256:b5f0369eef1572c8086ad2c8145b2901d3d7b735e2eea14cb628bed698a428e5 with 135513 bytes
 WARN 2018-11-21T19:54:56Z: graph_builder::registry: could not assemble metadata from blob (sha256:b5f0369eef1572c8086ad2c8145b2901d3d7b735e2eea14cb628bed698a428e5): 'cincinnati.json' not found
TRACE 2018-11-21T19:54:56Z: graph_builder::registry: Downloading layer sha256:53628feae132723d9a41f96f686a69b197485e6ad5792642e8ba8ec049f9ade4...
TRACE 2018-11-21T19:55:01Z: graph_builder::registry: v4.0: Looking for cincinnati.json in archive sha256:53628feae132723d9a41f96f686a69b197485e6ad5792642e8ba8ec049f9ade4 with 10845963 bytes
 WARN 2018-11-21T19:55:01Z: graph_builder::registry: could not assemble metadata from blob (sha256:53628feae132723d9a41f96f686a69b197485e6ad5792642e8ba8ec049f9ade4): 'cincinnati.json' not found
TRACE 2018-11-21T19:55:01Z: graph_builder::registry: Downloading layer sha256:f990075ff5a03f332dbd4134569418ea45e5ac1afbf9ab2b4d3e75c5d7c7561b...
TRACE 2018-11-21T19:55:05Z: graph_builder::registry: v4.0: Looking for cincinnati.json in archive sha256:f990075ff5a03f332dbd4134569418ea45e5ac1afbf9ab2b4d3e75c5d7c7561b with 9240722 bytes
 WARN 2018-11-21T19:55:06Z: graph_builder::registry: could not assemble metadata from blob (sha256:f990075ff5a03f332dbd4134569418ea45e5ac1afbf9ab2b4d3e75c5d7c7561b): 'cincinnati.json' not found
TRACE 2018-11-21T19:55:06Z: graph_builder::registry: Downloading layer sha256:aeb7866da422acc7e93dcf7323f38d7646f6269af33bcdb6647f2094fc4b3bf7...
TRACE 2018-11-21T19:55:32Z: graph_builder::registry: v4.0: Looking for cincinnati.json in archive sha256:aeb7866da422acc7e93dcf7323f38d7646f6269af33bcdb6647f2094fc4b3bf7 with 74703002 bytes
 WARN 2018-11-21T19:55:35Z: graph_builder::registry: could not assemble metadata from blob (sha256:aeb7866da422acc7e93dcf7323f38d7646f6269af33bcdb6647f2094fc4b3bf7): 'cincinnati.json' not found
ERROR 2018-11-21T19:55:35Z: graph_builder::graph: failed to fetch all release metadata
ERROR 2018-11-21T19:55:35Z: graph_builder::graph: metadata not found in any layer of tag v4.0
```

After:

```
DEBUG 2018-11-21T20:00:02Z: graph_builder::graph: Updating graph...
TRACE 2018-11-21T20:00:03Z: graph_builder::registry: processing: openshift/origin-release:v4.0
TRACE 2018-11-21T20:00:03Z: graph_builder::registry: Downloading layer sha256:b5f0369eef1572c8086ad2c8145b2901d3d7b735e2eea14cb628bed698a428e5...
TRACE 2018-11-21T20:00:04Z: graph_builder::registry: v4.0: Looking for cincinnati.json in archive sha256:b5f0369eef1572c8086ad2c8145b2901d3d7b735e2eea14cb628bed698a428e5 with 135513 bytes
TRACE 2018-11-21T20:00:04Z: graph_builder::registry: Downloading layer sha256:53628feae132723d9a41f96f686a69b197485e6ad5792642e8ba8ec049f9ade4...
TRACE 2018-11-21T20:00:09Z: graph_builder::registry: v4.0: Looking for cincinnati.json in archive sha256:53628feae132723d9a41f96f686a69b197485e6ad5792642e8ba8ec049f9ade4 with 10845963 bytes
TRACE 2018-11-21T20:00:10Z: graph_builder::registry: Downloading layer sha256:f990075ff5a03f332dbd4134569418ea45e5ac1afbf9ab2b4d3e75c5d7c7561b...
TRACE 2018-11-21T20:00:14Z: graph_builder::registry: v4.0: Looking for cincinnati.json in archive sha256:f990075ff5a03f332dbd4134569418ea45e5ac1afbf9ab2b4d3e75c5d7c7561b with 9240722 bytes
TRACE 2018-11-21T20:00:14Z: graph_builder::registry: Downloading layer sha256:aeb7866da422acc7e93dcf7323f38d7646f6269af33bcdb6647f2094fc4b3bf7...
TRACE 2018-11-21T20:00:39Z: graph_builder::registry: v4.0: Looking for cincinnati.json in archive sha256:aeb7866da422acc7e93dcf7323f38d7646f6269af33bcdb6647f2094fc4b3bf7 with 74703002 bytes
ERROR 2018-11-21T20:00:43Z: graph_builder::graph: failed to fetch all release metadata
ERROR 2018-11-21T20:00:43Z: graph_builder::graph: unable to find a release: could not assemble metadata from blob (sha256:aeb7866da422acc7e93dcf7323f38d7646f6269af33bcdb6647f2094fc4b3bf7): 'cincinnati.json' not found
```

I think your change is misleading. We didn't fail to find a release because we failed to assemble metadata - that was just the last error. That's why I was shadowing the error. We aren't actually interested in the last error.

steveej: > We aren't actually interested in the last error.

I see. In that case you can ditch my patch. The easiest seems to be to just swap the error message in https://github.com/openshift/cincinnati/pull/12/files#diff-991c40183d69ebfda48cb830e94be824R211 where the *tag* is available for the generic one you're currently shadowing it with. Then you don't have to do the stunt in the line we're currently commenting on.

crawford: `tag` is still referenced in that closure though, so the error message still has to be constructed before.

steveej: How about this on top of your changes?

```diff
diff --git a/graph-builder/src/registry.rs b/graph-builder/src/registry.rs
index d7a87e8..fdec333 100644
--- a/graph-builder/src/registry.rs
+++ b/graph-builder/src/registry.rs
@@ -196,7 +196,6 @@ fn find_first_release(
     repo: String,
     tag: String,
 ) -> impl Future<Item = Release, Error = Error> {
-    let err = format_err!("metadata not found in any layer of tag {}", tag);
     future::loop_fn(
         layer_digests.into_iter().peekable(),
         move |mut layer_digests_iter| {
@@ -209,8 +208,9 @@ fn find_first_release(
                         None,
                         Some(format_err!(
                             "no release found for tag '{}' and no more layers to examine",
-                            tag
+                            &tag
                         )),
+                        &tag,
                     ));
                     return Either::A(no_release_found);
                 }
@@ -246,6 +246,7 @@ fn find_first_release(
                                 metadata,
                             }),
                             None,
+                            &tag,
                         )),
                         Err(e) => future::ok(continue_or_break_loop(
                             layer_digests_iter,
@@ -255,6 +256,7 @@ fn find_first_release(
                                 layer_digest,
                                 e,
                             )),
+                            &tag,
                         )),
                     }
                 });
@@ -262,7 +264,6 @@ fn find_first_release(
         },
     )
     .flatten()
-    .map_err(|_| err)
 }
 
 #[derive(Debug, Deserialize)]
@@ -323,6 +324,7 @@ fn continue_or_break_loop<I>(
     mut layer_digests_iter: Peekable<I>,
     r: Option<Release>,
     e: Option<Error>,
+    tag: &str,
 ) -> future::Loop<Result<Release, Error>, Peekable<I>>
 where
     I: Iterator<Item = String>,
@@ -334,19 +336,24 @@ where
             trace!("Found release '{:?}'", r);
             Break(Ok(r))
         }
-        (_, Some(e)) => {
-            warn!("{}", e);
-            match layer_digests_iter.peek() {
-                Some(_) => Continue(layer_digests_iter),
-                None => Break(Err(e)),
+        (_, Some(e)) => match layer_digests_iter.peek() {
+            Some(_) => Continue(layer_digests_iter),
+
+            None => {
+                warn!("{}", e);
+                Break(Err(format_err!(
+                    "metadata not found in any layer of tag {}",
+                    tag
+                )))
             }
-        }
+        },
         _ => continue_or_break_loop(
             layer_digests_iter,
             None,
             Some(format_err!(
                 "continue_or_break called with unexpected condition"
             )),
+            tag,
         ),
     }
 }

```

steveej: > `tag` is moved into the closure below, so we have to clone it before. It's a _little_ easier to do that with the call to `format_err!`, but I could just clone it instead. What do you think?

I think cloning it is fine for now and we'll revisit this when we refactor.

steveej (approved):
/lgtm

PR #17, opened 2018-11-29
graph-builder/registry: use stream instead of loop

steveej:
This refactor intorduces the use of Stream throughout the whole chain in
which simplifies the termination of the execution in case a Release was
found for a set of layer digests.

Besides the technical change there is also the semantic change of not
returning any errors, and instead return an empty `Vec<Release>` in case
no release could be found.

### DIFF
--- a/graph-builder/src/registry.rs
+++ b/graph-builder/src/registry.rs
@@ -15,13 +15,12 @@
 use cincinnati;
 use failure::Error;
 use flate2::read::GzDecoder;
-use futures::future::{self, Either, Future};
 use futures::prelude::*;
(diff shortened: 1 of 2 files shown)


graph-builder/src/registry.rs
```diff
@@ -15,13 +15,13 @@
 use cincinnati;
 use failure::Error;
 use flate2::read::GzDecoder;
-use futures::future::{self, Either, Future};
+use futures::future::Future;
```
lucab: This should be part of the prelude, already imported in the next line.

PR #18, opened 2018-12-03
dist/Dockerfile: bump rustc and log dkregistry errors

steveej:
/cc @riuvshin


dist/Dockerfile
```diff
@@ -13,6 +13,6 @@ COPY . .
 RUN cargo build --release
 
 FROM centos
-ENV RUST_LOG=actix_web=error
+ENV RUST_LOG=actix_web=error,dkregistry=error
```
crawford, line 16: We should try to integrate this with the `-v` flag instead of relying on environment variables. This is okay for now though.

PR #19, opened 2018-12-03
add scripts for CI/CD integration

riuvshin:
add scripts for CI/CD integration

note I've used `ekidd/rust-musl-builder:1.30.1` for rust static builds in docker container.

this is needed for: https://jira.coreos.com/browse/APPSRE-252

cc: @crawford @steveeJ @pbergene

### DIFF
--- a/dist/Dockerfile
+++ b/dist/Dockerfile
@@ -1,18 +1,10 @@
-FROM centos
+FROM centos:7
 
 RUN yum -y groupinstall 'Development Tools'
 RUN yum -y install openssl-devel
 
-ADD https://static.rust-lang.org/dist/rust-1.30.1-x86_64-unknown-linux-gnu.tar.gz rust.tar.gz
-RUN tar -xf rust.tar.gz --strip 1
-RUN ./install.sh
-
-RUN mkdir app
-WORKDIR app
-COPY . .
-RUN cargo build --release
-
-FROM centos
 ENV RUST_LOG=actix_web=error,dkregistry=error
+
+COPY target/x86_64-unknown-linux-musl/release/graph-builder target/x86_64-unknown-linux-musl/release/policy-engine /usr/bin/
+
 ENTRYPOINT ["/usr/bin/graph-builder"]
-COPY --from=0 /app/target/release/graph-builder /app/target/release/policy-engine /usr/bin/
--- a/dist/build_deploy.sh
+++ b/dist/build_deploy.sh
@@ -1,0 +1,21 @@
+#!/usr/bin/env bash
+
+set -e
+
+ABSOLUTE_PATH="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+IMAGE_BUILD="${IMAGE_BUILD:-ekidd/rust-musl-builder:1.30.1}"
+IMAGE="quay.io/app-sre/cincinnati"
+IMAGE_TAG=$(git rev-parse --short=7 HEAD)
+PROJECT_PARENT_DIR=$ABSOLUTE_PATH/../
+DOCKERFILE_DEPLOY="$ABSOLUTE_PATH/Dockerfile"
+
+docker run --rm -v $PROJECT_PARENT_DIR:/home/rust/src $IMAGE_BUILD cargo build --release
+
+docker build -f $DOCKERFILE_DEPLOY -t "${IMAGE}:${IMAGE_TAG}" $PROJECT_PARENT_DIR
+
+if [[ -n "$QUAY_USER" && -n "$QUAY_TOKEN" ]]; then
+    DOCKER_CONF="$PWD/.docker"
+    mkdir -p "$DOCKER_CONF"
+    docker --config="$DOCKER_CONF" login -u="$QUAY_USER" -p="$QUAY_TOKEN" quay.io
+    docker --config="$DOCKER_CONF" push "${IMAGE}:${IMAGE_TAG}"
+fi
--- a/dist/pr_check.sh
+++ b/dist/pr_check.sh
@@ -1,0 +1,9 @@
+#!/usr/bin/env bash
+
+set -e
+
+ABSOLUTE_PATH="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+IMAGE_BUILD="${IMAGE_BUILD:-ekidd/rust-musl-builder:1.30.1}"
+PROJECT_PARENT_DIR=$ABSOLUTE_PATH/../
+
+docker run --rm -v $PROJECT_PARENT_DIR:/home/rust/src $IMAGE_BUILD cargo test


dist/Dockerfile
```diff
@@ -1,18 +1,10 @@
...
-RUN mkdir app
-WORKDIR app
-COPY . .
-RUN cargo build --release
+COPY target/x86_64-unknown-linux-musl/release/graph-builder target/x86_64-unknown-linux-musl/release/policy-engine /usr/bin/
 
-FROM centos
-ENV RUST_LOG=actix_web=error
 ENTRYPOINT ["/usr/bin/graph-builder"]
-COPY --from=0 /app/target/release/graph-builder /app/target/release/policy-engine /usr/bin/
```
riuvshin, line 18: on our CI infra we don't yet have docker with multi-stage build feature so this will not work.

build_deploy.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
```
steveej: can we use `/usr/bin/env bash` instead?

steveej (commented):
Can the scripts live in some sub-directory, e.g. *dist*? If so please move them there. Otherwise I understand this is a WIP and I'll revisit on ping.

build_deploy.sh
```diff
@@ -0,0 +1,23 @@
...
+DOCKERFILE_DEPLOY="dist/Dockerfile"
+
+# clean up
+rm -rf target
+
+docker run --rm -v $(pwd):/home/rust/src $IMAGE_BUILD cargo build --release
+
+docker build -f $DOCKERFILE_DEPLOY -t "${IMAGE}:${IMAGE_TAG}" .
+
+if [[ -n "$QUAY_USER" && -n "$QUAY_TOKEN" ]]; then
```
riuvshin: this will be injected by CI so you can assume it is always set on CI

riuvshin:
@steveeJ at the moment it is a requirement to have these scripts in the root of a repo. 
https://gitlab.cee.redhat.com/service/jjb/blob/master/ci-ext/00-common.yaml#L68
we may improve that later but for now all projects are following this requirement.


update: i;ve added possibility to override script locations https://gitlab.cee.redhat.com/service/jjb/merge_requests/89 so once it will be merged we can move scripts to `dist`

build_deploy.sh
riuvshin: done.

build_deploy.sh
steveej: TOKEN should be PASSWORD here

build_deploy.sh
```diff
@@ -0,0 +1,23 @@
...
+    DOCKER_CONF="$PWD/.docker"
+    mkdir -p "$DOCKER_CONF"
+    docker --config="$DOCKER_CONF" login -u="$QUAY_USER" -p="$QUAY_TOKEN" quay.io
```
steveej: s/TOKEN/PASSWORD

dist/Dockerfile
```diff
@@ -1,18 +1,10 @@
-FROM centos
+FROM centos:7
 
 RUN yum -y groupinstall 'Development Tools'
 RUN yum -y install openssl-devel
 
-ADD https://static.rust-lang.org/dist/rust-1.29.0-x86_64-unknown-linux-gnu.tar.gz rust.tar.gz
-RUN tar -xf rust.tar.gz --strip 1
-RUN ./install.sh
+ENV RUST_LOG=actix_web=error
```
steveej: We can obsolete #18 if you add `,dkregistry=error` here

build_deploy.sh
```diff
@@ -0,0 +1,23 @@
...
+IMAGE_BUILD="ekidd/rust-musl-builder:1.30.1"
+
+IMAGE="quay.io/app-sre/cincinnati"
+IMAGE_TAG=$(git rev-parse --short=7 HEAD)
+DOCKERFILE_DEPLOY="dist/Dockerfile"
+
+# clean up
+rm -rf target
+
+docker run --rm -v $(pwd):/home/rust/src $IMAGE_BUILD cargo build --release
```
steveej: is there a way not to hardcode this path, or do you assume this is never run locally?

pr_check.sh
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env bash
+
+set -e
+
+#TODO add logic for PR testing
```
steveej: We can start with `cargo test` for now

steveej (changes requested):

build_deploy.sh
```diff
@@ -0,0 +1,23 @@
...
+set -e
+
+IMAGE_BUILD="ekidd/rust-musl-builder:1.30.1"
+
+IMAGE="quay.io/app-sre/cincinnati"
+IMAGE_TAG=$(git rev-parse --short=7 HEAD)
+DOCKERFILE_DEPLOY="dist/Dockerfile"
+
+# clean up
+rm -rf target
```
steveej: Is this necessary in a CI environment? Need to clarify if this is also meant to be used on developer machines.

build_deploy.sh
riuvshin: no, we do not have a password, since this is a robot acc we have only token. all our jobs are using this method to auth against quay and it perfectly works.

pr_check.sh
riuvshin: done

build_deploy.sh
steveej: It works fine. It's just a bit confusing to read this misnomer since the terms are mixed up and it's technically a password:

```console
$ docker login --help | grep -- ' -p'
  -p, --password string   Password
```

steveej:
> update: i;ve added possibility to override script locations https://gitlab.cee.redhat.com/service/jjb/merge_requests/89 so once it will be merged we can move scripts to `dist`

That's great, thank you!

build_deploy.sh
crawford: `cargo clean` can do that, but why do we need to manually clean this up?

build_deploy.sh
riuvshin: the thing is that actual creds are stored in vault and there is a TOKEN + username, and it is injected by CI for all jobs in the same way, we can't change ENV var name unfortunately here.

dist/Dockerfile
riuvshin: done

build_deploy.sh
riuvshin: @crawford I just don't know what will happen when I will rerun the build, will it automatically clean target?
I think of it as similar to `mvn clean install` like clean up first and then new target will be generated once build is complete.
should I replace this with `cargo clean`?

steveej: Rerunning the build will produce a new binary if the source has changed. This has worked reliably for me during development so I wouldn't worry about it for local development.
I'm still trying to figure out if this script is supposed to be used on a local development machine or not ;-) If that's the case we probably need to adapt it for a couple of things:
* Don't remove the bulid cache (which happens on the line we're currently commenting on)
* ~~Don't try to push the image~~ Edit: this is gated by the environment variables so it's fine 
* ~~Swtich from a hardcoded `/home/rust/src` (see above)~~ Edit: I overlooked that this is inside the container so that's fine

build_deploy.sh
riuvshin: sorry, which path?
`/home/rust/src` this is just a path inside  `ekidd/rust-musl-builder:1.30.1` image it should work locally same way as on CI, moreover you can build your project locally without this script since I assume you have rust installed, this neeed more for CI where we want to run some build in docker to avoid of installing alot of deps on CI slaves.

build_deploy.sh
steveej: I see, that's a pity but OTOH it's not critical. Thanks for explaining!

build_deploy.sh
steveej: Ah right. I was looking too quickly and didn't catch that. Sorry for the fuss

build_deploy.sh
riuvshin: it's up to you use these scripts for local dev or not, from my experience some teams are doing that.

riuvshin: I've removed clean up from the script

riuvshin:
@steveeJ scripts  are moved to `dist/` and I've rebased PR

dist/build_deploy.sh
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env bash
+
+set -e
+
+ABSOLUTE_PATH="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+IMAGE_BUILD="ekidd/rust-musl-builder:1.30.1"
```
steveej: We might want to make these overridable as in `IMAGE_BUILD="${IMAGE_BUILD:-ekidd/rust-musl-builder:1.30.1}"`

dist/build_deploy.sh
```diff
@@ -0,0 +1,21 @@
...
+IMAGE="quay.io/app-sre/cincinnati"
```
steveej, line 7: same as above

steveej (commented):
One comment and we should be good with this \o/

dist/build_deploy.sh
riuvshin: that's a nice to have, agree

riuvshin: done

dist/build_deploy.sh
riuvshin: done

dist/build_deploy.sh
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env bash
+
+set -e
+
+ABSOLUTE_PATH="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+IMAGE_BUILD="${IMAGE_BUILD:-ekidd/rust-musl-builder:1.30.1}"
+IMAGE="quay.io/app-sre/cincinnati"
```
steveej, line 7: I meant to ask you to make this overridable too, or does that make no sense?

riuvshin: I don't think this makes a lot of sense since image is public...

steveej (approved):

steveej:
Ah, I just saw that we don't have squash and merge enabled for this repo. Would you mind squashing this down into a single commit?

riuvshin:
@steveeJ yeah, will do ASAP

PR #20, opened 2018-12-03
hack,README: add quickstart-container script

steveej:
@riuvshin PTAL

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,8 +4,18 @@ Cincinnati is an update protocol designed to facilitate automatic updates. It de
 
 ## Quick Start
 
+Prepare custom environment variables
+
 ```console
-cargo run --package graph-builder -- --address 0.0.0.0 --registry https://quay.io --repository <namespace/reponame> &
+# Please change these accordingly
+export CINCINNATI_REGISTRY="https://quay.io"
+export CINCINNATI_REPO="redhat/openshift-cincinnati-test-public-manual"
+```
+
+### Executables on the build host
+
+```console
+cargo run --package graph-builder -- --address 0.0.0.0 --registry "${CINCINNATI_REGISTRY}" --repository "${CINCINNATI_REPO}" &
 cargo run --package policy-engine -- --address 0.0.0.0 &
 curl --verbose --header 'Accept:application/json' http://localhost:8081/v1/graph
 ```
--- a/hack/quickstart-container.sh
+++ b/hack/quickstart-container.sh
@@ -1,0 +1,81 @@
+#!/usr/bin/env bash
+set -e
+
+CINCINNATI_REPO="${CINCINNATI_REPO:?Please set CINCINNATI_REPO}"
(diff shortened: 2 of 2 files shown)


steveej:
I reduced the scope of this PR to only add the script and mention it in the README. Extending the tests will go into another PR.

steveej:
@crawford PTAL

README.md
```diff
@@ -4,8 +4,24 @@ Cincinnati is an update protocol designed to facilitate automatic updates. It de
 
 ## Quick Start
 
+Prepare custom environment variables
+
+```console
+# Please change these accordingly
+CINCINNATI_REGISTRY="https://quay.io"
+CINCINNATI_REPO="steveej/cincinnati-test-public"
```
crawford: I don't want to point people at a personal repo.

hack/quickstart-container.sh
```diff
@@ -0,0 +1,81 @@
...
+    pkill "docker logs -f ${CINCINNATI_GB}"
+}
+trap cleanup EXIT
+
+# Run the graph builder
+echo Spawning graph-builder container...
+docker run -d --rm \
+  --label="${CINCINNATI_LABEL}" \
+  --name "${CINCINNATI_GB}" \
+  --env "RUST_LOG=actix_web=error,dkregistry=trace" \
```
crawford: trace seems aggressive.

crawford (commented):
Who is the intended consumer of this script?

steveej:
> Who is the intended consumer of this script?

I see it as a example towards we can point for a hands-on demonstration of how a cincinnati stack is set up. It's also useful for quick integration tests.

hack/quickstart-container.sh
steveej: This is meant for demo purposes as it shows nicely what requests are being made. I'm fine to change that to whatever you suggest

steveej: I lowered it to error too

README.md
steveej: I was able to set up the test repositories under the redhat organization :raised_hands:

steveej:
@lucab what's your opinion on this? Do you see this being helpful?

lucab:
@steveeJ sorry, I missed your earlier ping on this. For demo purpose, I think it makes sense to keep this at the documentation level. It is more helpful for people to see a quick example than having to dig through the variables in a bash script. My preference would be for something short and clean like https://github.com/coreos/matchbox/blob/master/Documentation/deployment.md#docker.

lucab:
> I'd prefer to have this script in the repository, even without a reference from the README. How about that?

I am not in strong opposition here, so let's go for that. Please keep the other change in README (variables for test repo), as I think it's valuable there.

steveej:
> I am not in strong opposition here, so let's go for that. Please keep the other change in README (variables for test repo), as I think it's valuable there.

ACK and done.

PR #22, opened 2018-12-04
dist: optimize build script and Dockerfile

steveej:
This change removes the overhead of sending the whole target directory
to the Docker daemon as the build context. Instead we can create a
mininum build context which contains only the desired static binaries.

Also removes the obsolete runtime deps.

/cc @riuvshin

### DIFF
--- a/dist/Dockerfile
+++ b/dist/Dockerfile
@@ -1,8 +1,5 @@
 FROM centos:7
 
-RUN yum -y groupinstall 'Development Tools'
-RUN yum -y install openssl-devel
-
 ENV RUST_LOG=actix_web=error,dkregistry=error
 
 COPY graph-builder policy-engine /usr/bin/
--- a/dist/Dockerfile
+++ b/dist/Dockerfile
@@ -5,6 +5,6 @@ RUN yum -y install openssl-devel
 
 ENV RUST_LOG=actix_web=error,dkregistry=error
 
-COPY target/x86_64-unknown-linux-musl/release/graph-builder target/x86_64-unknown-linux-musl/release/policy-engine /usr/bin/
+COPY graph-builder policy-engine /usr/bin/
 
 ENTRYPOINT ["/usr/bin/graph-builder"]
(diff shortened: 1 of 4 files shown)


dist/build_deploy.sh
```diff
@@ -8,10 +8,16 @@ IMAGE="quay.io/app-sre/cincinnati"
 IMAGE_TAG=$(git rev-parse --short=7 HEAD)
 PROJECT_PARENT_DIR=$ABSOLUTE_PATH/../
 DOCKERFILE_DEPLOY="$ABSOLUTE_PATH/Dockerfile"
+RELEASE_DIR="${PROJECT_PARENT_DIR}/target/x86_64-unknown-linux-musl/release"
+RELEASE_OUTPUT_DIR="${PROJECT_PARENT_DIR}/$(date +'%Y%m%d.%H%M%S')"
```
riuvshin: this will lead to workspace growing with every new build, I would like to overwrite all generated files or clean-up after the build

steveej: Ah, isn't the CI job spawned with a clean workspace?

riuvshin: no, this one has static slaves.

steveej:
[test]

steveej:
triggering another build to test if subsequent builds now work. previously it failed to cleanup the workspace cause some leftover files were owned by root.

[test]

dist/Dockerfile
```diff
@@ -1,10 +1,7 @@
 FROM centos:7
 
-RUN yum -y groupinstall 'Development Tools'
-RUN yum -y install openssl-devel
```
riuvshin, line 4: have you tested resulting image with this?

steveej: I tested the steps from the docker-based stack outlined in #20 and it works as expected

steveej: Ideally our CI would test that ;-)

steveej:
The two consecutive builds ([6](https://ci.ext.devshift.net/job/openshift-cincinnati-gh-pr-check/6/console) and [7](https://ci.ext.devshift.net/job/openshift-cincinnati-gh-pr-check/7/console)) were successful, I think this is good to go.

riuvshin:
awesome! 
/lgtm

steveej:
I probably can't do it myself either ;-) but let's see

/lgtm

PR #23, opened 2018-12-04
introduce openshift template

riuvshin:
introduce openshift template


riuvshin:
[test]

riuvshin:
@steveeJ this can be changed, I've moved it to dist

dist/openshift/cincinnati.yaml
```diff
@@ -0,0 +1,103 @@
...
+                  key: registry
+                  name: cincinnati
+            - name: REPOSITORY
+              valueFrom:
+                configMapKeyRef:
+                  key: repository
+                  name: cincinnati
+          args: ["-vvv", "--address", "$(ADDRESS)", "--registry", "$(REGISTRY)", "--repository", "$(REPOSITORY)"]
+          imagePullPolicy: Always
+          name: cincinnati
```
steveej: I think we should differentiate between the graph-builder and the policy-engine in the name. The two should already be able to work together so we can add the policy-engine to the template too and expose either the latter or both.

steveej (changes requested):
I'm not confident about the syntax of the template but I recommend to add the policy-engine as a second container and chain the two like in the examples of #20

steveej:
Please rebase this onto current master so it doesn't mess up the CI permissions.

dist/openshift/cincinnati.yaml
riuvshin: done

riuvshin:
[test]

dist/openshift/cincinnati.yaml
```diff
@@ -0,0 +1,163 @@
...
+      deploymentconfig: cincinnati
+- apiVersion: v1
+  kind: Service
+  metadata:
+    name: cincinnati-policy-engine
+  spec:
+    ports:
+      - protocol: TCP
+        port: 80
+        targetPort: 8080
```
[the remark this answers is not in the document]
riuvshin: yeah, moved it to parameters

steveej (approved):

PR #27, opened 2018-12-07
move loglevel to the config map

riuvshin:
move loglevel to the config map
/cc @steveeJ


steveej:
How about "verbosity" instead of "level"? I'd expect loglevel to take an absolute name instead of something cumulative

riuvshin:
updated

steveej:
@riuvshin I agree we should remove the secret from the repo but it's not related to this commit/PR. Can we deal with it separately?

steveej (approved):

PR #30, opened 2018-12-12
graph-builder: use high-level tokio API

steveej:
The lower-level tokio_core API is leaking memory in the way we've been
using it.


steveej:
Merging this to test the fix on stage.

PR #35, opened 2019-01-03
graph-builder: cache release metadata lookups

lucab:
This adds a caching layer for release metadata lookups (both positive
and negative results).

Closes: https://jira.coreos.com/browse/CORS-950

### DIFF
--- a/graph-builder/src/registry.rs
+++ b/graph-builder/src/registry.rs
@@ -13,11 +13,12 @@
 // limitations under the License.
 
 use cincinnati;
-use failure::Error;
+use failure::{Error, Fallible, ResultExt};
 use flate2::read::GzDecoder;
 use futures::prelude::*;
 use release::Metadata;
 use serde_json;
+use std::collections::HashMap;
 use std::fs::File;
 use std::io::Read;
 use std::iter::Iterator;
@@ -94,6 +95,7 @@ pub fn fetch_releases(
     repo: &str,
     username: Option<&str>,
     password: Option<&str>,
+    cache: &mut HashMap<u64, Option<Release>>,
 ) -> Result<Vec<Release>, Error> {
     let mut thread_runtime = tokio::runtime::current_thread::Runtime::new()?;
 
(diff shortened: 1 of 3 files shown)


steveej (commented):
Please point me to the negative caching implementation as I can't find it, even though the comments state it's there.

Furthermore you've assumed that tags are immutable, but since they have multiple layers I would assume otherwise. 
Following that the `tag` can't be used as the key in the `HashMap`, and instead the identity (hash) of the `layer_digests` vector should be used, which will change if a tag receives another layer in subsequent runs.
Alternatively we could cache the metadata for each `layer_digest` and lookup there before fetching and investigating the layers.

graph-builder/src/registry.rs
```diff
@@ -114,21 +116,67 @@ pub fn fetch_releases(
...
+        registry_host,
+        repo,
+        tag,
+    );
+    let (tag, release) = thread_runtime
+        .block_on(tagged_release)
+        .context("failed to find first release")?;
+
+    trace!("Caching release metadata for new tag {}", &tag);
+    cache.insert(tag, release.clone());
```
lucab: If I didn't misread the `find_first_release` logic, a negative result gets here as a `None` release which is then cached for that tag.

lucab: I will add a docstring on the data-structure to make this more explicit.

steveej: I see, that makes sense. I missed that `cache.get(..)` wraps it so we're getting `Option<Option<Release>>`.

steveej:
> If tags are immutable, 

Well, they aren't ;-)

> caching them directly will cut on each scrape a number of API calls to quay linear in the number of tags (to fetch every single manifest).

I wouldn't expect too much traffic here as the number of tags per repository will go up slowly as they supposedly correspond to releases.

graph-builder/src/registry.rs
```diff
@@ -163,13 +224,13 @@ fn find_first_release(
     authenticated_client: dkregistry::v2::Client,
     registry_host: String,
     repo: String,
-    tag: String,
-) -> impl Future<Item = Option<Release>, Error = Error> {
-    let tag_for_error = tag.clone();
+    repo_tag: String,
```
steveej, line 227: this seems a leftover from your local changes. let's just keep the current names

steveej (changes requested):
Blockers:
* The tests for `fetch_releases` need to be updated
* CI is stuck, I'll look after that

Non-blockers:
* A namechange nitpick in the comments
* We're missing an explicit test for the cache. I'm fine adding that after this PR
* There's one leftover TODO in the code, did you want to get rid of that now or later? I'm fine with keeping it in

graph-builder/src/registry.rs
lucab: No, this is to disambiguate between input and output `tag`.

lucab:
@steveeJ I think you reviewed a stale version of this PR, because unless I'm missing something `fetch_releases` tests are already updated (and pass locally).

steveej:
> @steveeJ I think you reviewed a stale version of this PR, because unless I'm missing something `fetch_releases` tests are already updated (and pass locally).

Pardon me! :pray:

/lgtm

steveej:
[test]

lucab:
@steveeJ np. The TODO I left in may require some invasive changes, I'm still researching how to approach it.

PR #36, opened 2019-01-04
graph-builder/registry: make client authentication non-blocking

lucab:
This refactors release-fecthing logic and makes all helper functions
more futures-friendly. That allows the removal of a blocking call, and
will make further chain-combinations easier in the future.

### DIFF
--- a/graph-builder/src/registry.rs
+++ b/graph-builder/src/registry.rs
@@ -15,6 +15,7 @@
 use cincinnati;
 use failure::{Error, Fallible, ResultExt};
 use flate2::read::GzDecoder;
+use futures::future;
 use futures::prelude::*;
 use release::Metadata;
 use serde_json;
@@ -59,10 +60,10 @@ pub fn read_credentials(
     })
 }
 
-pub fn authenticate_client(
-    client: &mut dkregistry::v2::Client,
+fn authenticate_client(
+    client: dkregistry::v2::Client,
     login_scope: String,
-) -> impl Future<Item = &dkregistry::v2::Client, Error = dkregistry::errors::Error> {
+) -> impl Future<Item = dkregistry::v2::Client, Error = Error> {
     client
         .is_v2_supported()
         .and_then(move |v2_supported| {
@@ -72,19 +73,21 @@ pub fn authenticate_client(
                 Ok(client)
             }
         })
-        .and_then(move |dclient| {
-            dclient.login(&[&login_scope]).and_then(|token| {
+        .and_then(move |mut dclient| {
+            dclient.login(&[&login_scope]).and_then(move |token| {
                 dclient
                     .is_auth(Some(token.token()))
                     .and_then(move |is_auth| {
                         if !is_auth {
                             Err("login failed".into())
                         } else {
-                            Ok(dclient.set_token(Some(token.token())))
+                            dclient.set_token(Some(token.token()));
+                            Ok(dclient)
                         }
                     })
             })
         })
+        .map_err(|e| format_err!("{}", e))
 }
 
 /// Fetches a vector of all release metadata from the given repository, hosted on the given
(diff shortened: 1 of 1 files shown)


graph-builder/src/registry.rs
```diff
@@ -131,36 +135,44 @@ pub fn fetch_releases(
...
-        // According to https://docs.docker.com/registry/spec/api/#listing-image-tags
-        // the tags should be ordered lexically but they aren't
-        .get_tags(&repo, Some(20))
+/// Fetch all tags for a repository, as a stream.
+///
+/// Tags order depends on registry implementation.
+/// According to [specs](https://docs.docker.com/registry/spec/api/#listing-image-tags),
+/// remote API should return tags in lexicographic order.
+/// However on Quay 2.9 this is not true.
+fn fetch_tags(
```
steveej: Why the rename?

graph-builder/src/registry.rs
```diff
@@ -98,28 +101,29 @@ pub fn fetch_releases(
...
+            let tags_stream = fetch_tags(auth_client, repo);
+            future::ok(tags_stream)
+        })
+        .flatten_stream()
+        .and_then(|(auth_client, tag)| fetch_layers_digests(auth_client, tag, repo))
+        .and_then(|(auth_client, tag, digests)| {
             find_first_release(
-                layer_digests,
-                authenticated_client.to_owned(),
+                auth_client,
```
steveej: Why swap and rename here?

graph-builder/src/registry.rs
```diff
@@ -131,36 +135,44 @@ pub fn fetch_releases(
...
-            Ok((tag, get_layer_digests(&manifest_kind, &manifest)?))
-        })
+        .and_then(move |(manifest_kind, manifest)| get_layer_digests(&manifest_kind, &manifest))
+        .map(move |digests| (client, tag, digests))
 }
 
 fn find_first_release(
-    layer_digests: Vec<String>,
     authenticated_client: dkregistry::v2::Client,
+    layer_digests: Vec<String>,
```
steveej: Why swap here?

graph-builder/src/registry.rs
```diff
@@ -98,28 +101,29 @@ pub fn fetch_releases(
...
-            &mut client,
-            format!("repository:{}:pull", &repo),
-        ))
-        .map_err(|e| format_err!("{}", e))?;
+        .map_err(|e| format_err!("{}", e));
 
-    let releases = get_tags(repo.to_owned(), authenticated_client)
-        .and_then(|tag| get_manifest_and_layers(tag, repo.to_owned(), authenticated_client))
-        .and_then(|(tag, layer_digests)| {
+    let releases = future::result(client)
```
steveej: How about creating or refactoring a function which consumes the `client` and incorporates some of the following `and_then`'s, thus abstracting some of the future specific mechanics here? It would also help readability by giving the functionality a name.

steveej (changes requested):
I'm not 100% convinced what the improvement is here. Please separate the sync->async changes from the renames/refactoring into two separate commits, as the latter seem mostly opinions on names and argument orders.

graph-builder/src/registry.rs
lucab: To have some consistency in parameter order across functions. They all take a client, in an arbitrary local order.

graph-builder/src/registry.rs
lucab: Because most functions around here are `get_foo`, but in some cases those are "get, as network fetches" and some are "get, as extract from in-memory data".

graph-builder/src/registry.rs
lucab: I'm fine with leaving this chain as-is for now. We'll need to revisit this later to fit caching and parallel fetches anyway.

lucab:
@steveeJ I have expunged all renames/moves and rebased to master, PTAL.

graph-builder/src/registry.rs
steveej: Is there a timely need for this PR? I for one don't see that so we can wait until there's time to revisit this PR.

lucab: I'm experimenting with making the scanning logic asynchronous and encapsulated in actix actors, which should be easier to parallelize. Removing the `block_on` as done in this PR is a pre-requisite for that, and a good cleanup in general.

steveej:
This has been implemented upstream in https://github.com/camallo/dkregistry-rs/commit/785f23a36944cf4c9a1cf179c024ffd725631658, which will be included in #195.

steveej:
Since #195 implemented this we can close here.

PR #37, opened 2019-01-07
OWNERS*: fix typo and make Luca an approver

steveej:
/cc @crawford


steveej:
The tests aren't working properly on prow until #11 is merged, so this needs to be forced.

PR #38, opened 2019-01-08
*: introduce path-namespace CLI argument

steveej:
This gives the option to prefix all paths with a given namespace.

Contributes to https://jira.coreos.com/browse/CORS-957.

### DIFF
--- a/commons/Cargo.toml
+++ b/commons/Cargo.toml
@@ -1,0 +1,6 @@
+[package]
+name = "commons"
+version = "0.1.0"
+authors = ["Stefan Junker <mail@stefanjunker.de>"]
+
+[dependencies]
--- a/graph-builder/src/main.rs
+++ b/graph-builder/src/main.rs
@@ -43,6 +43,7 @@ fn main() -> Result<(), Error> {
 
     let state = graph::State::new();
     let addr = (opts.address, opts.port);
+    let graph_path = format!("/{}/v1/graph", opts.path_namespace);
 
     {
         let state = state.clone();
@@ -52,7 +53,7 @@ fn main() -> Result<(), Error> {
     server::new(move || {
         App::with_state(state.clone())
             .middleware(Logger::default())
-            .route("/v1/graph", Method::GET, graph::index)
+            .route(&graph_path, Method::GET, graph::index)
     })
     .bind(addr)?
     .run();
--- a/policy-engine/src/main.rs
+++ b/policy-engine/src/main.rs
@@ -3,6 +3,7 @@
 extern crate actix;
 extern crate actix_web;
 extern crate cincinnati;
+extern crate commons;
 extern crate env_logger;
 #[macro_use]
 extern crate failure;
@@ -56,10 +57,12 @@ fn main() -> Result<(), Error> {
     let state = graph::State {
         upstream: opts.upstream,
     };
+
+    let graph_path = format!("/{}/v1/graph", opts.path_namespace);
     server::new(move || {
         App::with_state(state.clone())
             .middleware(Logger::default())
-            .route("/v1/graph", Method::GET, graph::index)
+            .route(&graph_path, Method::GET, graph::index)
     })
     .bind((opts.address, opts.port))?
     .start();
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -640,6 +640,7 @@ version = "0.1.0"
 dependencies = [
  "actix-web 0.7.13 (registry+https://github.com/rust-lang/crates.io-index)",
  "cincinnati 0.1.0",
+ "commons 0.1.0",
  "dkregistry 0.2.2-alpha.0 (git+https://github.com/camallo/dkregistry-rs.git?rev=b6573b0e2c18da6f1e50ebfc1a0d6b7ec3c8accf)",
  "env_logger 0.6.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "failure 0.1.3 (registry+https://github.com/rust-lang/crates.io-index)",
@@ -1164,6 +1165,7 @@ dependencies = [
  "actix 0.7.6 (registry+https://github.com/rust-lang/crates.io-index)",
  "actix-web 0.7.13 (registry+https://github.com/rust-lang/crates.io-index)",
  "cincinnati 0.1.0",
+ "commons 0.1.0",
  "env_logger 0.6.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "failure 0.1.3 (registry+https://github.com/rust-lang/crates.io-index)",
  "futures 0.1.25 (registry+https://github.com/rust-lang/crates.io-index)",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -300,6 +300,10 @@ dependencies = [
 ]
 
 [[package]]
+name = "commons"
+version = "0.1.0"
+
+[[package]]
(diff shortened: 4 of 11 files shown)


cincinnati/src/lib.rs
```diff
@@ -19,6 +19,8 @@ extern crate serde;
 #[macro_use]
 extern crate serde_derive;
 
+pub mod common;
```
lucab: The `cincinnati` crate so far is a client library for graph manipulation (which I was hopefully planning to re-use in the fedora-coreos agent). I think this may not belong here.

policy-engine/src/main.rs
```diff
@@ -44,10 +44,11 @@ fn main() -> Result<(), Error> {
         .init();
 
     // Metrics service.
+    let metrics_path = opts.path_namespace.clone() + "/metrics";
     server::new(move || {
         App::new()
             .middleware(Logger::default())
-            .route("/metrics", Method::GET, metrics::serve)
+            .route(&metrics_path, Method::GET, metrics::serve)
```
lucab: This is on a different port, and not exposed as a public API. I think it shouldn't be prefix.

policy-engine/src/main.rs
```diff
@@ -56,10 +57,12 @@ fn main() -> Result<(), Error> {
     let state = graph::State {
         upstream: opts.upstream,
     };
+
+    let graph_path = opts.path_namespace.clone() + "/v1/graph";
```
lucab: If I'm reading the ticket correctly, this will result in `/api/updates_mgmt/v1/v1/graph` for both services. What's the semantic difference between the two `v1`? Is it fine that they are duplicated?

cincinnati/src/common.rs
```diff
@@ -0,0 +1,8 @@
+/// Adds "/" if the input is not empty
+pub fn parse_path_namespace(path_namespace: &str) -> String {
+    if path_namespace == "" {
+        path_namespace.to_string()
+    } else {
+        "/".to_owned() + path_namespace
```
lucab: You may want to take care of existing leading and trailing slashes, to avoid having (leading or trailing) `//` in the final string.

cincinnati/src/common.rs
```diff
@@ -0,0 +1,8 @@
+/// Adds "/" if the input is not empty
```
lucab: I think the actual logic you want is something like "make sure any (optional) namespace starts with a slash and ends without one".

policy-engine/src/main.rs
steveej: I *think* they would use `api/updates_mgmt` as the namespace but I don't know exactly. Anyway from Cincinnati perspective the namespace is opaque and we shouldn't care about the content.

cincinnati/src/common.rs
steveej: The docstring mentions that it must be provided without leading `/`, but I can handle it in the code as well if you think it's necessary.

cincinnati/src/common.rs
steveej: See above, I relied on the user to provide a correct string as this is only a compatibility feature.

policy-engine/src/main.rs
lucab: I see. I was just confused about version prefixes semantics.

cincinnati/src/common.rs
lucab: Not necessary (can be done at any later point if we still want), but if this become deployable by the general audience, I can foresee human mistakes (or plainly not reading the help docs). Feel free to leave it for now.

cincinnati/src/common.rs
lucab: Ack.

cincinnati/src/lib.rs
steveej: I wanted to discuss this before merging, thanks for bringing it up. Do we want a 4th crate in the repo named "common"?

cincinnati/src/common.rs
```diff
@@ -0,0 +1,17 @@
...
+use self::regex::Regex;
+use failure::Error;
+
+/// Strip any leading and trailing slashes
+pub fn parse_path_namespace(path_namespace: &str) -> Result<String, Error> {
+    if path_namespace == "" {
+        return Ok(path_namespace.to_string());
+    };
+
+    let stripped_leading = Regex::new(r"^/+")?.replace(path_namespace, "").to_string();
```
lucab: Just out of curiosity, why not https://doc.rust-lang.org/std/string/struct.String.html#method.trim_start_matches?

cincinnati/src/lib.rs
lucab: That looks cleaner (maybe a bit overkill for a single function?).

steveej: I moved it to a new crate anyway. I hope the single function doesn't feel too lonely :upside_down_face:

cincinnati/src/common.rs
steveej: Simply because I missed it while reading through the docs :-) even better there's one which works for prefixes and suffixes in the same call!

commons/Cargo.toml
```diff
@@ -0,0 +1,7 @@
+[package]
+name = "commons"
+version = "0.1.0"
+authors = ["Stefan Junker <mail@stefanjunker.de>"]
+edition = "2018"
```
lucab: Other crates are at an older edition (they should be updated at some point), I'm not sure if that's an issue.

graph-builder/src/main.rs
```diff
@@ -13,6 +13,8 @@
 // limitations under the License.
 
 extern crate actix_web;
+extern crate cincinnati;
+extern crate failure;
```
lucab: I think this may be a leftover, but I didn't check the actual warnings.

commons/Cargo.toml
steveej: This is the default now when you `cargo init ..`, so I didn't consciously set this. I'd take the chance to see if it causes trouble eventually, it hasn't as of now ;-)

steveej:
All comments addressed @lucab, PTAL.

commons/Cargo.toml
crawford: That's going to require Rust 1.31, which Red Hat doesn't support. We should drop back to 2015 for now.

steveej: Who can bump Rust in Red Hat? It's the latest stable so I don't expect much push back. Anyway, removing this doesn't hurt anything

lucab: I know we currently have a 1.29 requirement for coreos-metadata (see https://github.com/coreos/coreos-metadata/pull/127#issuecomment-447907284), because we need to build that as a normal rpm in a RHEL/fedora/centos buildroot.

However cincinnati is currently built in a custom container (now at 1.31) and there is no minimum version check in CI, so it is effectively already at 1.31.

lucab:
(for reference, my previous reviews were rejected due to the same permission mismatch as https://github.com/openshift/installer/pull/588#issuecomment-438583520)

commons/Cargo.toml
steveej: @lucab I'm pretending we actually depend on the RHEL build pipeline at one point. We only have CD from the app-sre system and not from the openshift system. AFAIU the latter will require to use official build images instead of custom ones, but we'll see ;-)

PR #39, opened 2019-01-10
dist: rm obsolete Dockerfile

steveej:


steveej:
/hold

@lucab Following the OOB conversation this is the answer that we do need that Dockerfile even though it's not directly referenced in any of the repo's files. I had a dark memory of this Dockerfile being used for the app-sre CI, but I don't know where this path is configured. @riuvshin please help me out here :-)

PR #29, opened 2018-12-12
bump dkregistry and adapt to API changes

steveej:
Blocked on https://github.com/camallo/dkregistry-rs/pull/76.

PR #40, opened 2019-01-10
policy-engine: add OpenAPI spec and endpoint

steveej:
Contains #41.
Closes https://jira.coreos.com/browse/CORS-955.


steveej:
The OpenAPI spec is still missing the input arguments on the graph endpoint. Therefore putting this on hold.
/hold

policy-engine/src/openapi.rs
```diff
@@ -0,0 +1,22 @@
+use actix_web::{HttpRequest, HttpResponse};
+use string_template::Template;
+use AppState;
+
+pub(crate) fn index(req: HttpRequest<AppState>) -> HttpResponse {
+    let path_prefix = &req.state().path_prefix;
+    let spec_template = include_str!("openapiv3.json");
```
lucab: I think this should be moved to a module constant instead.

policy-engine/src/openapi.rs
```diff
@@ -0,0 +1,22 @@
...
+    let spec = Template::new(spec_template).render_named(
+        &[("path_prefix", path_prefix.as_str())]
+            .iter()
+            .cloned()
+            .collect::<std::collections::HashMap<&str, &str>>(),
+    );
+
+    if spec_template == spec {
```
lucab: This seems to belong to a test, not to the serving endpoint.

policy-engine/src/openapi.rs
```diff
@@ -0,0 +1,22 @@
...
+
+pub(crate) fn index(req: HttpRequest<AppState>) -> HttpResponse {
+    let path_prefix = &req.state().path_prefix;
+    let spec_template = include_str!("openapiv3.json");
+    let spec = Template::new(spec_template).render_named(
+        &[("path_prefix", path_prefix.as_str())]
+            .iter()
+            .cloned()
+            .collect::<std::collections::HashMap<&str, &str>>(),
+    );
```
lucab: All of this seems a bit overkill to me for a single parameter fill in. Why not just a `format!()` which also does the check statically? (may need a wrapping macro if the template is not a literal)

policy-engine/src/openapiv3.json
```diff
@@ -0,0 +1,89 @@
...
+                        }
+                    }
+                }
+            }
+        }
+    },
+    "components": {
+        "schemas": {
+            "Graph": {
+                "type": "array",
```
lucab: I think this is wrong. A graph response is a top-level object containing two fields, `nodes` and `edges`, both of which are arrays with a homogeneous type.

policy-engine/src/openapi.rs
steveej: With `format!()` we'd have to escape every curly brace with another curly brace. I prefer to have the spec template as close to the original syntax as possible which is achieved with the current method. When you say overkill, which metric are you talking about? Since the development time is already spent anyway, do you see performance drawbacks compared to another method?

lucab: Logical complexity. As in, I would have expected the simpler, static approach with no external dependencies first. But your explanation makes sense, I'm ok with sticking to it.

This isn't a core hotpath and should be easily cacheable (both on our and on proxy side), so I'm not too concerned with performances here.

steveej: Another approach which I'm considering is to omit the template strings in the specification and use an OpenAPI parser which understands the spec and can rewrite the paths before serializing it again.

steveej:
@lucab
All comments are addressed and the alternative way of mangling the paths sits in a commit on top. PTAL and let me know if there's anything left to fix before squashing.

policy-engine/src/openapi.rs
```diff
@@ -0,0 +1,65 @@
...
+        Ok(s) => HttpResponse::from(s),
+        Err(e) => {
+            let e = format_err!("Could not serialize OpenAPI object: {}", e);
+            error!("{}", e);
+            HttpResponse::from_error(e.into())
+        }
+    }
+}
+
+fn rewrite_paths(paths: &openapiv3::Paths, path_prefix: &str) -> openapiv3::Paths {
```
lucab: This allocates a whole new `Paths`. Is it possible to re-arrange this to take ownership of it and consume the iterator into the result? Or alternatively, prefix the paths in place?

policy-engine/src/openapi.rs
```diff
@@ -0,0 +1,65 @@
...
+        .map(|(path, path_item)| {
+            let new_path = format!("{}{}", path_prefix, &path);
+            trace!("Rewrote path {} -> {} ", &path, &new_path);
+            (new_path, path_item.clone())
+        })
+        .collect()
+}
+
+#[cfg(test)]
+mod tests {
```
lucab, line 44: Can we maybe add a roundtrip test to make sure the library is not returning crazy results at runtime?

policy-engine/src/openapi.rs
lucab: That implementation looks nice. Comments I have on that are:
 * it goes through a deserialization+serialization roundtrip on each request, but as above I don't think performance is a concern here
 * runtime result depends on correctness of the library. I am not familiar with it, I hope you did a manual sanity check on the result

policy-engine/src/openapi.rs
steveej: I think that'd be redundantly testing the crates we're using which I'd avoid if possible. I did test it manually and it works as expected, and our unit test covers any OpenAPI related mistake which could be made in the spec itself.

policy-engine/src/openapi.rs
steveej: I wish we could rely on some CoW mechanism, or maybe we can? I'll look into your suggestion and report back.

policy-engine/src/openapi.rs
lucab: Ack, I guess I was being over-cautious there :)

steveej:
@lucab 

All the comments addressed in best-effort manner.

It could likely be optimized further but IMO that's not worth it for this code-path, as we're talking about the OpenAPI endpoint which I expect to have really low traffic.

policy-engine/src/openapi.rs
lucab: Current moving in and out seems clean and nice enough to me, thanks!

steveej:
Looks like there were changes on *quay.io* causing the following issues with the manifest: 

```
test net::quay_io::fetch_release_public_with_no_release_metadata_must_not_error ... FAILED
test net::quay_io::fetch_release_public_with_first_empty_tag_must_succeed ... FAILED
test net::quay_io::fetch_release_public_without_credentials_must_fail ... ok

failures:

---- net::quay_io::fetch_release_public_with_no_release_metadata_must_not_error stdout ----
thread 'net::quay_io::fetch_release_public_with_no_release_metadata_must_not_error' panicked at 'should not error on emtpy repo: ErrorMessage { msg: "unknown manifest_kind \'None\'" }', libcore/result.rs:1009:5
note: Run with `RUST_BACKTRACE=1` for a backtrace.

---- net::quay_io::fetch_release_public_with_first_empty_tag_must_succeed stdout ----
thread 'net::quay_io::fetch_release_public_with_first_empty_tag_must_succeed' panicked at 'fetch_releases failed: : ErrorMessage { msg: "unknown manifest_kind \'None\'" }', libcore/result.rs:1009:5


failures:
    net::quay_io::fetch_release_public_with_first_empty_tag_must_succeed
    net::quay_io::fetch_release_public_with_no_release_metadata_must_not_error

test result: FAILED. 1 passed; 2 failed; 0 ignored; 0 measured; 0 filtered out
```

There's [currently an incident on *quay.io*][quayincident] but I've yet to figure out if we're facing a backwards-compatibility issue or a temporary issue.

[quayincident]: http://status.quay.io/incidents/72jyq4322h85

steveej:
I tracked down the issue to a permanent quay.io change. I [proposed a band-aid for dkregistry-rs](https://github.com/camallo/dkregistry-rs/pull/82) with room for improvement.

steveej:
Rebased on master after #29 was merged which fixed the recent issue with quay.io.

@crawford or @lucab PTAL

steveej:
[test]

PR #41, opened 2019-01-10
*: bugfix and refactor prefix semantics

steveej:
The bugfix and semantic change is that the prefix keeps, and if
necessary adds, a leading slash. This simplifies the logic of dealing
with prefixes because it doesn't require special-casing for emtpy
namespaces which was buggy before.

Using actix-web's prefix method removes the need to specify the prefix
for every new route.

Fixes up #38.

### DIFF
--- a/graph-builder/src/main.rs
+++ b/graph-builder/src/main.rs
@@ -43,7 +43,7 @@ fn main() -> Result<(), Error> {
 
     let state = graph::State::new();
     let addr = (opts.address, opts.port);
-    let graph_path = format!("/{}/v1/graph", opts.path_namespace);
+    let app_prefix = opts.path_prefix.clone();
 
     {
         let state = state.clone();
@@ -51,9 +51,12 @@ fn main() -> Result<(), Error> {
     }
 
     server::new(move || {
-        App::with_state(state.clone())
+        let app_prefix = app_prefix.clone();
+        let state = state.clone();
+        App::with_state(state)
             .middleware(Logger::default())
-            .route(&graph_path, Method::GET, graph::index)
+            .prefix(app_prefix)
+            .route("/v1/graph", Method::GET, graph::index)
     })
     .bind(addr)?
     .run();
--- a/policy-engine/src/config.rs
+++ b/policy-engine/src/config.rs
@@ -1,6 +1,6 @@
 //! Command-line options for policy-engine.
 
-use commons::parse_path_namespace;
+use commons::parse_path_prefix;
 use hyper::Uri;
 use std::net::IpAddr;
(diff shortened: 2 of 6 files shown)


graph-builder/src/main.rs
```diff
@@ -43,17 +43,20 @@ fn main() -> Result<(), Error> {
...
+    let app_prefix = opts.path_prefix.clone();
 
     {
         let state = state.clone();
         thread::spawn(move || graph::run(&opts, &state));
     }
 
     server::new(move || {
-        App::with_state(state.clone())
+        let app_prefix = app_prefix.clone();
```
lucab, line 54: Can this (and the `state` below too) be just moved in? it looks like there is nothing else holding references to those.

policy-engine/src/graph.rs
```diff
@@ -84,8 +85,3 @@ pub(crate) fn index(req: HttpRequest<State>) -> Box<Future<Item = HttpResponse,
         }
     }
 }
-
-#[derive(Clone)]
-pub struct State {
```
lucab, line 89: I would personally leave this here, instead of moving it up to `main.rs` (and derive `Debug`, while at it).

policy-engine/src/main.rs
```diff
@@ -54,19 +54,27 @@ fn main() -> Result<(), Error> {
     .start();
 
     // Main service.
-    let state = graph::State {
-        upstream: opts.upstream,
+    let state = AppState {
+        upstream: (&opts).upstream.clone(),
```
lucab: This syntax looks a bit weird, are you sure it's needed?

lucab (commented):
Nice! I didn't know about actix-web `.prefix()`.

I've left a few minor comments, looks fine in general.

policy-engine/src/graph.rs
steveej: The reason I moved it out is that it's no longer specific to `graph.rs`, as `openapi.rs` also makes use of it, so there would be a dependency on the former by the latter, which I wanted to avoid.

lucab: Ok, let's move it then.

policy-engine/src/config.rs
```diff
@@ -30,11 +30,11 @@ pub struct Options {
     #[structopt(long = "metrics_port", default_value = "9081")]
     pub metrics_port: u16,
 
-    /// Path namespace prefix for all paths. Must be given without leading and ending '/'
+    /// Path prefix prefix for all paths.
```
lucab: Double "prefix" here as well.

PR #42, opened 2019-01-14
graph-builder: log update completion

lucab:
This adds support for emitting a debug message once a graph update
completes, recording the total number of releases being served.

### DIFF
--- a/cincinnati/src/lib.rs
+++ b/cincinnati/src/lib.rs
@@ -122,6 +122,11 @@ impl Graph {
             dag: &self.dag,
         }
     }
+
+    /// Return the number of releases (nodes) in the graph.
+    pub fn releases_count(&self) -> u64 {
+        self.dag.node_count() as u64
+    }
 }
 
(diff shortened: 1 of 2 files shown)


steveej:
I don't mind adding this but I don't see it as particularly useful. We already have a test which counts the number of releases. What could be useful is adding this as a metric which can be scraped by the monitoring tools.

lucab:
@steveeJ the first commit should be useful for metrics too (once that is in place). The second commit just surfaces the completion plus count to logs. I definitely see some value in it right now, but agree that once graph-builder metrics are in place it can be subsumed there.

steveej:
> I definitely see some value in it right now

I'd like to learn more about the use-case where someone would check the logs for that number.

lucab:
> I'd like to learn more about the use-case where someone would check the logs for that number.

It's a debug log, so meant for development/debugging. At the moment as a developer, I'd like to have this in place to get a quick-glance idea/correlation between 1) number of releases in the graph 2) update time 3) steady memory consumption.

PR #44, opened 2019-01-14
openshift: fix liveness and readyness probes

riuvshin:
openshift: fix liveness and readyness probes
/cc @steveeJ


steveej:
Thanks again!

/lgtm

PR #45, opened 2019-01-16
cincinnati: check for mandatory parameters in client query

lucab:
Cincinnati instances may require by specs a mandatory set of parameters
that must be part of client request. This adds configuration knobs to
all components so that the mandatory set can be configured and enforced
at runtime.

Ref: https://jira.coreos.com/projects/CORS/issues/CORS-966

/cc @steveeJ

### DIFF
--- a/commons/src/lib.rs
+++ b/commons/src/lib.rs
@@ -5,11 +5,13 @@ pub fn parse_path_prefix(path_prefix: &str) -> String {
 
 #[cfg(test)]
 mod tests {
+    use super::*;
+
     #[test]
     fn test_parse_path_prefix() {
-        assert_eq!(super::parse_path_prefix("//a/b/c//"), "/a/b/c");
-        assert_eq!(super::parse_path_prefix("/a/b/c/"), "/a/b/c");
-        assert_eq!(super::parse_path_prefix("/a/b/c"), "/a/b/c");
-        assert_eq!(super::parse_path_prefix("a/b/c"), "/a/b/c");
+        assert_eq!(parse_path_prefix("//a/b/c//"), "/a/b/c");
+        assert_eq!(parse_path_prefix("/a/b/c/"), "/a/b/c");
+        assert_eq!(parse_path_prefix("/a/b/c"), "/a/b/c");
+        assert_eq!(parse_path_prefix("a/b/c"), "/a/b/c");
     }
 }
--- a/commons/src/lib.rs
+++ b/commons/src/lib.rs
@@ -1,6 +1,51 @@
+#[macro_use]
+extern crate failure;
+extern crate url;
+
+use failure::Fallible;
+use std::collections::HashSet;
+use url::form_urlencoded;
+
 /// Strip all but one leading slash and all trailing slashes
 pub fn parse_path_prefix(path_prefix: &str) -> String {
     format!("/{}", path_prefix.to_string().trim_matches('/'))
+}
+
+/// Parse a comma-separated set of client parameters keys.
+pub fn parse_params_set(params: &str) -> HashSet<String> {
+    params
+        .split(',')
+        .filter_map(|key| {
+            let trimmed = key.trim().to_string();
+            if trimmed.is_empty() {
+                None
+            } else {
+                Some(trimmed)
+            }
+        })
+        .collect()
+}
+
+/// Make sure `query` string contains all `params` keys.
+pub fn ensure_query_params(params: &HashSet<String>, query: &str) -> Fallible<()> {
+    // No mandatory parameters, always fine.
+    if params.is_empty() {
+        return Ok(());
+    }
+
+    // Extract and de-duplicate keys from input query.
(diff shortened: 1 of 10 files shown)


graph-builder/src/graph.rs
```diff
@@ -19,11 +19,17 @@ use config;
...
-use std::collections::HashMap;
+use std::collections::{HashMap, HashSet};
 use std::sync::{Arc, RwLock};
 use std::thread;
 
 pub fn index(req: HttpRequest<State>) -> HttpResponse {
+    // Check for required client parameters.
+    let mandatory_params = &req.state().mandatory_params;
+    if commons::ensure_query_params(mandatory_params, req.query_string()).is_err() {
+        return HttpResponse::NotFound().finish();
```
steveej: I think that "Bad Request" instead of "Not Found" would be more suitable

steveej (commented):
* If these parameters are configurable and we want to keep the OpenAPI in sync with the running service they have to be injected into the OpenAPI endpoint too.
* A minor comment

commons/src/lib.rs
```diff
@@ -1,15 +1,94 @@
...
+    // Make sure all mandatory parameters are present.
+    ensure!(
+        params.is_subset(&query_keys),
+        "mandatory parameters missing"
+    );
+
+    Ok(())
+}
+
+#[cfg(test)]
```
lucab, line 51: Note to self: spurious line.

lucab:
Indeed I forgot to touch the openapi endpoint, will add that logic.

graph-builder/src/graph.rs
steveej: 404 is more specific but it doesn't fit this case. Unfortunately there's no more specific way than 400 to say "Bad Arguments", so I think "Bad Request" comes closest. On Linux I would compare this to EINVAL (not existent in HTTP protocol but 400 comes closest) vs ENOENT (compared to 404).

lucab:
Added openapi logic too. Some fields in `openapiv3` objects are private, so they have to be constructed via deserializing a JSON template.

/hold cancel

commons/src/lib.rs
```diff
@@ -1,15 +1,93 @@
...
+    Ok(())
+}
+
 #[cfg(test)]
 mod tests {
+    use super::*;
+
     #[test]
     fn test_parse_path_prefix() {
-        assert_eq!(super::parse_path_prefix("//a/b/c//"), "/a/b/c");
```
steveej, line 10: Would you mind separating this refactor into a preceding commit? Just in the case of reverting the new feature we can keep the other.

commons/src/lib.rs
```diff
@@ -1,15 +1,93 @@
...
 /// Strip all but one leading slash and all trailing slashes
 pub fn parse_path_prefix(path_prefix: &str) -> String {
     format!("/{}", path_prefix.to_string().trim_matches('/'))
 }
 
+/// Parse a comma-separated set of client parameters keys.
+pub fn parse_params_set(params: &str) -> HashSet<String> {
+    params
+        .split(',')
+        .into_iter()
```
steveej: This line is redundant as [split already returns an iterator](https://doc.rust-lang.org/std/primitive.str.html#iterator-behavior)

policy-engine/src/openapi.rs
```diff
@@ -1,9 +1,23 @@
...
-use openapiv3::OpenAPI;
+use openapiv3::{OpenAPI, ReferenceOr};
+use std::collections::HashSet;
 use AppState;
 
+/// Template for policy-engine OpenAPIv3 document.
 const SPEC: &str = include_str!("openapiv3.json");
 
+/// Template for an OpenAPIv3 query parameter.
+static PARAM_TEMPLATE: &str = r#"
```
steveej: Asking for education since we now have both: please elaborate about using `static` over `const` for global variables since?

steveej: Also there's only one reference to this in one function, so you could move it there.

steveej (commented):
> Added openapi logic too. Some fields in `openapiv3` objects are private, so they have to be constructed via deserializing a JSON template.

Not the most elegant way but oh well. I added a few comments which should be resolved quickly.

policy-engine/src/openapi.rs
lucab: The difference is explained here: https://doc.rust-lang.org/book/ch19-01-unsafe-rust.html#accessing-or-modifying-a-mutable-static-variable.

I initially had it inline, but I thought it was adding noise when reading through the function, so moved it up.

commons/src/lib.rs
lucab: Oh, nice!

lucab:
@steveeJ last push contains all your suggestions. CI is happy. PTAL.

lucab:
@steveeJ no problems, I've split that one too (and I'll try to keep in mind for the future).

steveej:
> @steveeJ no problems, I've split that one too (and I'll try to keep in mind for the future).

Thanks!

/lgtm

PR #46, opened 2019-01-18
quay: introduce a quay.io client library

lucab:
This introduces an asynchronous quay.io client, based on reqwest.
Its feature set is currently limited to fetching tags and labels.

### DIFF
--- a/quay/src/v1/mod.rs
+++ b/quay/src/v1/mod.rs
@@ -1,0 +1,106 @@
+use failure::Fallible;
+use reqwest::{self, async};
+
+mod manifest;
+
+mod tag;
+pub use self::tag::Tag;
+
+static DEFAULT_API_BASE: &str = "https://quay.io/api/v1/";
+
+/// Client to make outgoing API requests to a quay instance.
+#[derive(Clone, Debug)]
+pub struct Client {
+    /// Base URL for API endpoint.
+    api_base: reqwest::Url,
+    /// Asynchronous reqwest client.
+    hclient: async::Client,
+    /// Authentication token.
+    token: Option<String>,
+}
+
+impl Client {
+    /// Return a client builder with default options.
+    pub fn builder() -> ClientBuilder {
+        ClientBuilder::default()
+    }
+
+    /// Return a request builder with base URL and parameters set.
+    pub(crate) fn new_request<S: AsRef<str>>(
+        &self,
+        method: reqwest::Method,
+        url_suffix: S,
+    ) -> Fallible<async::RequestBuilder> {
+        let url = self.api_base.clone().join(url_suffix.as_ref())?;
+        let builder = {
+            let plain = self.hclient.request(method, url);
+            match self.token {
+                None => plain,
+                Some(ref token) => {
+                    let bearer_token = format!("Bearer {}", token);
+                    plain.header("Authorization", bearer_token)
+                }
+            }
+        };
+        Ok(builder)
+    }
+}
+
+/// Client to make outgoing API requests to a quay instance.
+#[derive(Clone, Debug)]
+pub struct ClientBuilder {
+    api_base: Option<String>,
+    hclient: Option<async::Client>,
+    token: Option<String>,
+}
+
+impl ClientBuilder {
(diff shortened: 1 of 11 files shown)


lucab:
/cc @steveeJ 
This should already implement the minimal logic required by cincinnati to scrape labels on active tags. Authentication logic is as described [here](https://docs.quay.io/api/#generating-a-token-for-internal-application-use) but I haven't tested it yet. CI will need an access token for testing too.

cgwalters:
I'll just say I skimmed this, looks sane to me.  I found it useful to read Futures code as I don't have much experience with it.

lucab:
@steveeJ I fixed and validated the authentication part too, manually on a personal private organization. Test for private images are written but not enabled (pending a CI token), some of the assertions may be slightly off. I left a TODO for pagination. PTAL.

steveej:
This is quite some code, thanks for the effort! Due to the size of this I will start integrating it with the rest of the code and see how it performs there. I will add my review once that's done.

steveej:
Oh, I missed that functionality as I was looking for a "label" file. So the
*labels.rs* can be dropped from my commit. Do you mind if I force push to
this branch?

lucab:
@steveeJ feel free to force push to this, I locally duplicated my stuff in another private branch.

steveej:
@lucab 

> @steveeJ feel free to force push to this, I locally duplicated my stuff in another private branch.

I merged the additional code with the `manifest.rs`. I had also done a rebase but feel free to just cherry-pick my two commits on top of a local branch and re-force-push ;-)

quay/src/v1/mod.rs
```diff
@@ -0,0 +1,106 @@
...
+#[derive(Clone, Debug)]
+pub struct ClientBuilder {
+    api_base: Option<String>,
+    hclient: Option<async::Client>,
+    token: Option<String>,
+}
+
+impl ClientBuilder {
+    /// Set (or reset) the HTTP client to use.
+    pub fn http_client(self, hclient: Option<async::Client>) -> Self {
```
steveej, line 59: I would add a `set_` or `with_` prefix to this method name because it usually means to get the thing if it doesn't say otherwise.

quay/src/v1/mod.rs
```diff
@@ -0,0 +1,106 @@
...
+        let mut builder = self;
+        builder.hclient = hclient;
+        builder
+    }
+
+    /// Set (or reset) the access token to use.
+    pub fn access_token(self, token: Option<String>) -> Self {
```
steveej, line 66: I would add a `set_` or `with_` prefix to this method name because it usually means to get the thing if it doesn't say otherwise.

quay/src/v1/mod.rs
```diff
@@ -0,0 +1,106 @@
...
+        let mut builder = self;
+        builder.token = token;
+        builder
+    }
+
+    /// Set (or reset) the base API endpoint URL to use.
+    pub fn api_base(self, api_base: Option<String>) -> Self {
```
steveej, line 73: I would add a `set_` or `with_` prefix to this method name because it usually means to get the thing if it doesn't say otherwise.

steveej (approved):
Optionally you could add a `use tokio::runtime::Runtime` in the tests as it's used quite often.

steveej:
PTAL at the comments before we merge this. I consider them optional so just remove the hold or ping me back after you addressed them.

/lgtm
/hold

quay/src/v1/mod.rs
lucab: The advice is generally true. However in this specific case this a builder, and in my experience so far builders don't explicitly label methods as setters (I guess because they don't have getters):
 * https://doc.rust-lang.org/std/fs/struct.DirBuilder.html
 * https://doc.rust-lang.org/std/fs/struct.OpenOptions.html
 * https://docs.rs/reqwest/0.9.9/reqwest/struct.ClientBuilder.html

I'll leave these methods without the prefix (same for the following comments too).

lucab:
@steveeJ I've added use statements (but stuck with current names) and repushed. From my side this is ok to land.

/lgtm

PR #47, opened 2019-01-24
cincinnati: add Partial{,Eq} to Graph 

steveej:
This greatly simplifies testing of Graph related functions.

---

- [x] remove `unwraps()` from non-test code

### DIFF
--- a/cincinnati/src/lib.rs
+++ b/cincinnati/src/lib.rs
@@ -29,6 +29,8 @@ use std::{collections, fmt};
 
 pub const CONTENT_TYPE: &str = "application/json";
 
+const EXPECT_NODE_WEIGHT: &str = "all exisitng nodes to have a weight (release)";
+
 #[derive(Debug, Default)]
 pub struct Graph {
     dag: Dag<Release, Empty>,
@@ -75,7 +77,7 @@ impl<'a> Iterator for NextReleases<'a> {
     fn next(&mut self) -> Option<Self::Item> {
         self.children
             .walk_next(self.dag)
-            .map(|(_, i)| self.dag.node_weight(i).unwrap())
+            .map(|(_, i)| self.dag.node_weight(i).expect(EXPECT_NODE_WEIGHT))
     }
 }
 
@@ -90,7 +92,7 @@ impl Graph {
         let release = release.into();
         match self.find_by_version(&release.version()) {
             Some(id) => {
-                let mut node = self.dag.node_weight_mut(id.0).unwrap();
+                let mut node = self.dag.node_weight_mut(id.0).expect(EXPECT_NODE_WEIGHT);
                 if let Release::Concrete(_) = node {
                     bail!(
                         "Concrete release with the same version ({}) already exists",
--- a/cincinnati/src/lib.rs
+++ b/cincinnati/src/lib.rs
@@ -28,7 +28,6 @@ use std::collections::HashMap;
 use std::{collections, fmt};
 
 pub const CONTENT_TYPE: &str = "application/json";
-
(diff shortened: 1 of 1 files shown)


steveej:
I corrected the PR title (the original one will follow soon) and fixed up existing and newly introduced unwraps in that file. this is ready for review now!

/cc @lucab

cincinnati/src/lib.rs
```diff
@@ -237,14 +238,77 @@ impl Serialize for Graph {
...
+                        neighbors_releases_other.sort_by(asc_order_release_by_version);
+                        neighbor_releases
+                            .iter()
+                            .zip(neighbors_releases_other.iter())
+                            .all(|(neighbor, neighbor_other)| neighbor == neighbor_other)
+                    }
+                })
+                .count()
+                // Ensure each node in self has exactly one matching node in
+                // other , including its neighbors
```
lucab: Spurious space before the comma.

cincinnati/src/lib.rs
```diff
@@ -237,14 +238,77 @@ impl Serialize for Graph {
...
+                    let node_index_other = node_ref_other.0;
+                    let release_other = node_ref_other.1;
+
+                    // Node matches
+                    release == release_other && {
+                        let (mut neighbor_releases, mut neighbors_releases_other) = (vec![], vec![]);
+                        // Determine neighbors of own and matching node
+                        self.dag
+                            .neighbors(node_index)
+                            .zip(dag_other.neighbors(node_index_other))
```
lucab, line 286: I'm not sure about my understanding, so please double-check the following assertion. When `zip` combines the two iterators, the resulting length is the shortest one. Thus if there a matching on a strict subset, this may still think the two iterators are equals.

cincinnati/src/lib.rs
```diff
@@ -237,14 +238,77 @@ impl Serialize for Graph {
...
+
 #[cfg(test)]
 mod tests {
     extern crate serde_json;
 
     use super::*;
 
-    #[test]
-    fn serialize_graph() {
+    fn get_graph() -> Graph {
```
lucab: Minor nit: I usually try to avoid `get_` as it is very generic verb. I would call something like this `generate_graph`.

lucab (commented):
Few minor nits, and a more general question on zipping iterators.

cincinnati/src/lib.rs
steveej: Nice catch, your suspicion was correct on this one. The zipper ends as soon as one of the iterators returns `None`. I'll fix this up

steveej:
@lucab
> Few minor nits, and a more general question on zipping iterators.

Thanks for the review! I've addressed all the comments and simplified the code a bit.

cincinnati/src/lib.rs
steveej: I went with the trivial solution and prepended the iteration with a count assertion

cincinnati/src/lib.rs
```diff
@@ -237,14 +238,92 @@ impl Serialize for Graph {
...
+
+                        if neighbors_count != neighbors_other_count {
+                            return false;
+                        }
+
+                        let mut neighbors = self
+                            .dag
+                            .neighbors(node_index)
+                            .zip(dag_other.neighbors(node_index_other))
+                            .fold(
```
lucab, line 287: Side note: we are into three levels of loop-nesting at this point, all in serial logic. I think it is fine for now in order to start landing an `Eq` implementation. Longer term we may (or may not) have a need to investigate a more efficient DAG comparison algorithm.

lucab (approved):

PR #48, opened 2019-01-25
graph-builder: process the release/previous/next labels

steveej:
Depends on
- [x] #46 
- [x] #47
- [x] #51 

Fixes https://jira.coreos.com/browse/CORS-884.

### DIFF
--- a/cincinnati/src/lib.rs
+++ b/cincinnati/src/lib.rs
@@ -128,6 +128,28 @@ impl Graph {
     pub fn releases_count(&self) -> u64 {
         self.dag.node_count() as u64
     }
+
+    /// Prune the graph from all abstract releases
+    ///
+    /// Return the number of pruned releases
+    pub fn prune_abstract(&mut self) -> usize {
+        let to_remove: Vec<daggy::NodeIndex> = self
+            .dag
+            .node_references()
+            .filter_map(|nr| {
+                if let Release::Abstract(_) = nr.weight() {
+                    Some(nr.0)
+                } else {
+                    None
+                }
+            })
+            .collect();
+
+        to_remove
+            .iter()
+            .filter(|ni| self.dag.remove_node(**ni).is_some())
+            .count()
+    }
 }
 
 impl<'a> Deserialize<'a> for Graph {
--- a/graph-builder/src/graph.rs
+++ b/graph-builder/src/graph.rs
@@ -128,7 +128,7 @@ pub fn run<'a>(opts: &'a config::Options, state: &State) -> ! {
             releases = populated_releases;
         };
 
-        match create_graph(releases) {
+        match create_graph(releases, &opts.quay_label_filter) {
             Ok(graph) => match serde_json::to_string(&graph) {
                 Ok(json) => {
                     *state.json.write().expect("json lock has been poisoned") = json;
@@ -145,11 +145,16 @@ pub fn run<'a>(opts: &'a config::Options, state: &State) -> ! {
     }
 }
 
-pub fn create_graph(releases: Vec<registry::Release>) -> Result<Graph, Error> {
+pub fn create_graph(
+    releases: Vec<registry::Release>,
+    quay_label_filter: &str,
+) -> Result<Graph, Error> {
     let mut graph = Graph::default();
 
     releases
         .into_iter()
+        .filter_map(|release| process_release_remove_label(quay_label_filter, release))
+        .map(|release| process_neighbor_labels(quay_label_filter, release))
         .inspect(|release| trace!("Adding a release to the graph '{:?}'", release))
         .try_for_each(|release| {
             let previous = release.metadata.previous.clone();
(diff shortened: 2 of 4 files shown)


steveej:
@lucab 

I've thoroughly split this PR into many commits so we could split this into several PRs if you'd prefer that, and/or cherry-pick some of this into #46.

steveej:
> I've thoroughly split this PR into many commits so we could split this into several PRs if you'd prefer that, and/or cherry-pick some of this into #46.

I've done that. Putting this on 
/hold
until #46 is figured out.

graph-builder/src/registry.rs
```diff
@@ -121,17 +124,35 @@ pub fn fetch_releases(
...
+                &manifestref,
+                Some(&quay_label_filter),
+            ))?;
+
+            let release = Release {
+                metadata: Metadata {
+                    metadata: quay_labels
+                        .into_iter()
+                        .map(|label| (label.key, label.value))
+                        .chain(release.metadata.metadata)
```
lucab: I think these two iterators should be chained in the reverse order. I think with the current logic, if the same label key exists in both image manifest and quay API, the image one is inserted last and kept.

(This is from a quick reading of this code and my current understanding of hashmap collection. Please double-check my assertion as either one of those could be wrong.)

steveej: Nice catch. I meant to revisit this before merging but who knows if I would've remembered today since I forgot to add a TODO for it too ;-) If we want to the labels to have precedence -- which could be useful even though I can't come up with a immediate use case now -- your suggestion is correct and we need to reverse them. See [this playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=402ad76ed37ad002397bdaf00ced72b4).

graph-builder/src/registry.rs
```diff
@@ -121,17 +124,35 @@ pub fn fetch_releases(
...
             layer_digests,
             authenticated_client.to_owned(),
             registry_host.to_owned(),
             repo.to_owned(),
             tag.to_owned(),
             cache,
-        )?;
-        if let Some(metadata) = release {
-            releases.push(metadata);
+        )? {
```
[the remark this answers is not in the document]
steveej: I see your point. I made it less dense and introduced a match statement which allows to see what happens in a glimpse

graph-builder/src/registry.rs
lucab: I think it makes sense to state that mutable labels have precedence over image ones. The usecase I had in mind was "amend a mistake in an (immutable) image metadata label, via a (mutable) quay label with the same key-name".

steveej:
All comments addressed and rebased.

/hold off

cincinnati/src/lib.rs
```diff
@@ -128,6 +128,28 @@ impl Graph {
     pub fn releases_count(&self) -> u64 {
         self.dag.node_count() as u64
     }
+
+    /// Prune the graph from all abstract releases.ConcreteRelease
+    ///
+    /// Return the number of pruned releases
+    pub fn prune_abstract(&mut self) -> usize {
```
lucab, line 135: Side note (not for this PR): we should document what is an abstract release, why/how they can appear as part of graph, and what's the special logic around them.

cincinnati/src/lib.rs
```diff
@@ -128,6 +128,28 @@ impl Graph {
...
+            .dag
+            .node_references()
+            .filter_map(|nr| {
+                if let Release::Abstract(_) = nr.weight() {
+                    Some(nr.0)
+                } else {
+                    None
+                }
+            })
+            .collect();
```
lucab, line 146: Out of curiosity, what's the reason for collecting here and doing a further loop instead of filtering nodes during iteration?

steveeJ:
> The following test **failed**, say `/retest` to rerun them all:
> ci/prow/cargo-test 	[ce914e8](https://github.com/openshift/cincinnati/commit/ce914e8d8d828b7fd2f76523fc889df327aa7293) 	[link](https://openshift-gce-devel.appspot.com/build/origin-ci-test/pr-logs/pull/openshift_cincinnati/48/pull-ci-openshift-cincinnati-master-cargo-test/91) 	`/test cargo-test`

The error is due a lack of `protoc` on CI. Now we have two options:

1. Separate the code generation from the build and only include the generated files and the schema
2. Install `protoc` on the CI and either of:
    1. verify that the generated code is the same as checked in
    2. add the generated files to the gitignore

My preference here is 2.1.

@lucab what's yours?

graph-builder/src/graph.rs
```diff
@@ -93,28 +113,16 @@ pub fn run(opts: &config::Options, state: &State) -> ! {
     }
 }
 
-fn create_graph(
-    opts: &config::Options,
-    username: Option<&str>,
-    password: Option<&str>,
-    cache: &mut HashMap<u64, Option<registry::Release>>,
+pub fn create_graph(
+    releases: Vec<registry::Release>,
```
lucab, line 149: Few minor doubts on the choice of this `Vec` datastructure: are releases supposed to be unique in this vector? Are release supposed to be order-insensitive in this vector?

If the answer to both of the above is positive, then an `HashSet` is a better choice to encode the invariant.

graph-builder/src/registry.rs
```diff
@@ -108,6 +109,8 @@ pub fn fetch_releases(
         .build()
         .map_err(|e| format_err!("{}", e))?;
 
+    let quay_client: quay::v1::Client = quay::v1::Client::builder().build()?;
```
lucab: This probably needs an access token to be set on the builder, coming from config. Were you planning to add it later?

graph-builder/src/graph.rs
```diff
@@ -142,5 +150,101 @@ fn create_graph(
...
+
+    if let Some(s) = remove_prefixed_label(filter, &mut release, "release.remove") {
+        // Filter releases which have "{prefix}.release-remove=true"
+        let remove = std::str::FromStr::from_str(&s).unwrap_or_else(|e| {
+            debug!("could not parse '{}': {}", s, e);
+            true
+        });
+
+        if remove {
+            debug!("Removing release '{:?}'", release);
```
lucab, line 220: This sounds like a `trace` (unless you were planning to watch the labels live initially).

graph-builder/src/graph.rs
```diff
@@ -142,5 +150,101 @@ fn create_graph(
...
+fn process_release_remove_label(
+    filter: &str,
+    release: registry::Release,
+) -> Option<registry::Release> {
+    let mut release = release;
+
+    if let Some(s) = remove_prefixed_label(filter, &mut release, "release.remove") {
+        // Filter releases which have "{prefix}.release-remove=true"
+        let remove = std::str::FromStr::from_str(&s).unwrap_or_else(|e| {
+            debug!("could not parse '{}': {}", s, e);
```
lucab: I'd go for `error` here, in the hope that a human sees and fixes this.
And also make it more explicit like `could not parse {} as a boolean (assuming true): {}`

I also wonder if we should be more conservative and default to `false` instead on errors.

graph-builder/src/graph.rs
```diff
@@ -142,5 +150,101 @@ fn create_graph(
...
 
+    graph.prune_abstract();
+
     Ok(graph)
 }
+
+/// Removes and retrieves the metadata value of the given `key` which is prefixed
+/// by `filter`
+fn remove_prefixed_label(
+    filter: &str,
```
lucab: Nit: I would personally re-order arguments so that the `&mut` comes first and filter+key are next to each other.

graph-builder/src/graph.rs
steveej: I had the same thoughts and then I went down the rabbit whole, which I just did again to give you the error output :-)

```
error[E0277]: the trait bound `std::collections::HashMap<std::string::String, std::string::String>: std::hash::Hash` is not satisfied
  --> graph-builder/src/release.rs:30:5
   |
30 |     pub metadata: HashMap<String, String>,
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `std::hash::Hash` is not implemented for `std::collections::HashMap<std::string::String, std::string::String>`
   |
   = note: required by `std::hash::Hash::hash`
```

I'm happy to use a `HashSet` if you have a proper solution for the above.

graph-builder/src/graph.rs
steveej: Just for future reference: we can do that ordering for new methods but not when changing mutability

steveej:
@lucab I've rebased this and addressed all your comments. I've already planned on refactoring the node removal code from `create_graph()` method in the same fashion #54 removes the metadata fetching from `fetch_releases()`. I think we should merge this nonetheless and do the refactoring when #54 is in.

cincinnati/src/lib.rs
steveej: I couldn't find a way to remove the nodes while iterating over them. Removing them requires exclusive access to `self.dag` but it's borrowed by the iteration.

graph-builder/src/graph.rs
steveej: That was the idea yes. Do you think it gets too verbose?

steveej:
I think this is good to go.

/hold cancel

graph-builder/src/graph.rs
lucab: I think initially it would be fine to double-check the behavior; later we can probably lower this if it becomes too verbose.

steveej:
@lucab this is rebased and ready for another pass. PTAL

cincinnati/src/lib.rs
lucab: Ack, let's leave it as is for now then.

PR #50, opened 2019-01-29
cincinnati: add plugin-interface, code, and plugins; rewire the graph-builder to use plugins

steveej:
Introducing a well-defined Cincinnati plugin interface with a distinction between internal and externals plugins.
Internal plugins are mere language structs which implement the `InternalPlugin` trait and therefore run in the same process as the component they're plugged in to.
External plugins are a combination of an internal client which implements the `ExternalPlugin` trait, and an external service which implements the Protobuf scheme defined at [*cincinnati/plugins/src/interface.proto*](cincinnati/plugins/src/interface.proto).


steveej:
> The following test **failed**, say `/retest` to rerun them all:
> ci/prow/cargo-test 	[ce914e8](https://github.com/openshift/cincinnati/commit/ce914e8d8d828b7fd2f76523fc889df327aa7293) 	[link](https://openshift-gce-devel.appspot.com/build/origin-ci-test/pr-logs/pull/openshift_cincinnati/48/pull-ci-openshift-cincinnati-master-cargo-test/91) 	`/test cargo-test`

The error is due a lack of `protoc` on CI. Now we have two options:

1. Separate the code generation from the build and only include the generated files and the schema
2. Install `protoc` on the CI and either of:
    1. verify that the generated code is the same as checked in
    2. add the generated files to the gitignore

My preference here is 2.1.

@lucab what's yours?

cincinnati/build.rs
```diff
@@ -0,0 +1,13 @@
+extern crate protoc_rust;
+
+use std::env;
+
+fn main() {
+    println!("{}/{}", env::var("OUT_DIR").unwrap(),"src/");
```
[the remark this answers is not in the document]
steveej: I've tried your suggestion and it can be made to work. An annoyance is that features aren't passed to member crates from virtual crates. We're working around this in [dist/pr_check.sh](https://github.com/openshift/cincinnati/blob/master/dist/pr_check.sh#L16-L22).

Since your suggestion is to not enable this on CI we wouldn't have to go through that for the regular builds but we'll have to at least document it.

lucab:
@steveeJ how hard is to get `protoc` on CI? (I almost gave up on travis due to libraries and version mismatches).

I don't expect the protobuf schema to ever change after this PR, so I would be fine with just a documented feature flag for build.rs, and running it either manually or in CI (depending on the above question).

steveej:
> @steveeJ how hard is to get `protoc` on CI? (I almost gave up on travis due to libraries and version mismatches).

Likely too much effort for our frequency of schema change.



> I don't expect the protobuf schema to ever change after this PR, so I would be fine with just a documented feature flag for build.rs, and running it either manually or in CI (depending on the above question).

I wouldn't say "never" but likely not very often. The PR already contains your suggestion, but I've yet to address the other comments, will do that in a bit.

steveej:
@lucab I've reworked this with a proper git history. The protobuf schema and generated code now live in commons, but I've left the rest of the code in-place as it would've been very invasive to move.

lucab:
@steveeJ this looks good on the surface. I am experimenting with building a policy-plugin using this, I will report back and review once I get a better picture on the whole thing.

commons/protos/policy_interface.proto
```diff
@@ -0,0 +1,22 @@
...
+
+message Graph {
+  repeated Node nodes = 1;
+  repeated Edge edges = 2;
+}
+
+message PluginInput {
+  Graph graph = 1;
+  map<string, string> parameters = 2;
+}
```
steveej: Missing:
* PluginOutput containing:  with outgoing parameters and a PluginError
* PluginError containing a number and a string

commons/protos/policy_interface.proto
```diff
@@ -0,0 +1,39 @@
...
+message PluginOutput {
+  message Error {
+    enum Status {
+      SUCCESS = 0;
+      FAILURE = 1;
+    };
+
+    Status plugin_status = 1;
+    string description = 2;
+    uint64 http_status = 3;
```
lucab: I think this will go out-of-band if/when we use HTTP as a transport, not inside the protobuf.

commons/protos/policy_interface.proto
```diff
@@ -0,0 +1,39 @@
...
+  }
+
+  Error error = 1;
```
lucab: I'm not a protobuf expert, but would it be possible to model the `Result<GraphAndParams, Error>` more directly? I seem to remember `oneof` would allow that. I think we can even re-use (after renaming) the graph+params message we use for input.

commons/protos/policy_interface.proto
```diff
@@ -0,0 +1,39 @@
...
+
+message PluginInput {
+  Graph graph = 1;
+  map<string, string> parameters = 2;
+}
+
+message PluginOutput {
+  message Error {
+    enum Status {
+      SUCCESS = 0;
```
lucab: Two comments on this:
 * I think we don't strictly need the "success" bit here if we can model the two arms of Result directly (see my comment below)
 * for JSON side, I picked up a textual `kind` field: https://github.com/openshift/cincinnati/pull/58/files#diff-1bb65de438c6b2131604dcc53ca74722R50. Do we want to align on a single approach?

policy-engine/src/graph.rs
```diff
@@ -56,6 +56,7 @@ pub(crate) fn index(
         HTTP_GRAPH_BAD_REQS.inc();
         return Box::new(future::err(e));
     }
+    let plugin_params = req.query().to_owned();
```
lucab, line 65: Leftover?

steveej: Good catch. I haven't converted the policy-engine to using the plugins yet. We don't actually have any plugins which have to run in the policy-engine just yet, but I'll prepare the infrastructure for it within the scope of this PR.

steveej:
@lucab Thanks for having a look at this XXL PR :pray: I've addressed all but one comment which will be resolved by a stubbing in an empty plugin list for the policy-engine.

I suggest we sync on this PR in a real-time conversation, I'll get in touch shortly!

policy-engine/src/graph.rs
steveej: This one is in use again, even though the consume is a no-op.

lucab:
I did a further pass on this WIP, and the synchronous implementations seem sane. One major doubt on the traits and enum arrangement.

cincinnati/src/plugins/internal/metadata_fetch_quay.rs
```diff
@@ -0,0 +1,195 @@
...
+        Ok(Self {
+            client,
+            repo,
+            label_filter,
+            manifestref_key,
+        })
+    }
+}
+
+mod async {
```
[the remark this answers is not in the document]
steveej: would uppercase `Async` work?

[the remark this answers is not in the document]
steveej: Since I don't want to blow the scope of this PR let's get all of it in without async support. Therefore I've removed the mod altogether.

cincinnati/src/plugins/mod.rs
```diff
@@ -0,0 +1,285 @@
...
+/// Dummy converter to satisfy the Trait
+impl TryFrom<PluginIO> for PluginIO {
+    type Err = Error;
+
+    fn try_from(io: Self) -> Fallible<Self> {
+        Ok(io)
+    }
+}
+
+/// Dummy convertert to satisfy the Trait
```
lucab: Typo: `converter`. Also, you may want to align the docstrings for those three identity-converters.

cincinnati/src/plugins/mod.rs
```diff
@@ -0,0 +1,285 @@
...
+                .context("could not parse ExternalIO to PluginExchange")?;
+
+        Ok(Self {
+            graph: plugin_exchange.take_graph().into(),
+            parameters: Default::default(),
+        })
+    }
+}
+
+/// Convert from an IntenralIO to a PluginExchange
```
lucab: Typo: `InternalIO`.

cincinnati/src/plugins/mod.rs
```diff
@@ -0,0 +1,285 @@
...
+    fn try_from(plugin_io: PluginIO) -> Fallible<Self> {
+        match plugin_io {
+            PluginIO::InternalIO(internal_io) => Ok(internal_io),
+            PluginIO::ExternalIO(external_io) => external_io.try_into(),
+        }
+    }
+}
+
+/// Copmatibility impl to automatically convert the unvarianted PluginIO when the
+/// variant struct ExternalIO is expented.
```
lucab: Nit: I think that neither `unvarianted` nor `expented` are English word, but maybe it is just my dictionary lacking a bit.

cincinnati/src/plugins/mod.rs
```diff
@@ -0,0 +1,285 @@
...
+            PluginIO::InternalIO(internal_io) => internal_io.try_into()?,
+        };
+        Ok(final_io)
+    }
+}
+
+/// Wrapper struct for the implementation for implementing Plugin<PluginIO>
+pub struct InternalPluginWrapper<T>(pub T);
+
+/// Wrapper struct for the implementation for implementing Plugin<PluginIO>
```
lucab: Here and in struct above, is `for the implementation` a leftover?

cincinnati/src/plugins/mod.rs
```diff
@@ -0,0 +1,285 @@
...
+    T: ExternalPlugin,
+{
+    fn run(&self, plugin_io: PluginIO) -> Fallible<PluginIO> {
+        println!("running external plugin");
+        let external_io = self.0.run_external(plugin_io.try_into()?)?;
+        external_io.try_into()
+    }
+}
+
+/// Processes all given Plugins sequentially.InternalIO
```
lucab: Nit: spurious trailing `InternalIO`.

cincinnati/src/plugins/mod.rs
```diff
@@ -0,0 +1,285 @@
...
+///
+/// This function automatically converts between the different IO representations
+/// if necessary.
+pub fn process<T: 'static>(
```
lucab: What is this static lifetime for? Is it going to cause problems as soon as we move to runtime-configured list of plugins?

cincinnati/src/plugins/mod.rs
```diff
@@ -0,0 +1,285 @@
...
+    plugins: &[Box<T>],
+    initial_io: InternalIO,
```
lucab: Cosmetic/Opinion: I tend to think as this function being the external entrypoint, thus taking a graph+params and not the internal type. It doesn't effectively change anything except for my code-skimming approach, so feel free to disregard the comment if you think otherwise.

lucab:
@steveeJ left several comments but pretty much all of them are minor things (and a few questions). Plugin trait rework is much easier to follow and looks fine now, thanks for that. Whenever you want to drop the stubbed things and the WIP label, I think this is close to be mergeable.

cincinnati/src/plugins/internal/metadata_fetch_quay.rs
```diff
@@ -0,0 +1,128 @@
...
+    repo: String,
+    label_filter: String,
+    manifestref_key: String,
+}
+
+impl InternalPlugin for QuayMetadataFetchPlugin {
+    fn run_internal(&self, io: InternalIO) -> Fallible<InternalIO> {
+        let (mut graph, parameters) = (io.graph, io.parameters);
+
+        trace!("fetching metadata from quay labels..");
```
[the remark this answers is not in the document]
steveej: that was on purpose. should I use three instead?

cincinnati/src/plugins/mod.rs
steveej: whops, `s/expented/expected`. I don't know a better term for the former though :thinking:

cincinnati/src/plugins/internal/metadata_fetch_quay.rs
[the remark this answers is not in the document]
lucab: Ah, I didn't realize that. Then I'm personally used to three, yes.

cincinnati/src/plugins/mod.rs
steveej: It supposedly justifies the existence of these structs, as they're only used for that. Did it confuse you?

cincinnati/src/plugins/mod.rs
steveej: Fortunately this is leftover from an intermediate state.

cincinnati/src/plugins/mod.rs
steveej: In my mental model this is an internal entrypoint, since it's called internally by the components main loop. Eventually it won't need to take `InternalIO` which includes a `Graph` and the parameters `HashMap`, as the first plugin will produce the first `Graph`. In that case we can replace `InternalIO` here with just the `HashMap` which is provided by the client and/or component configuration.

cincinnati/src/lib.rs
```diff
@@ -30,7 +38,7 @@ use std::{collections, fmt};
 pub const CONTENT_TYPE: &str = "application/json";
 const EXPECT_NODE_WEIGHT: &str = "all exisitng nodes to have a weight (release)";
 
-#[derive(Debug, Default)]
+#[derive(Debug, Default, Clone)]
```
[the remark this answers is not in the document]
steveej: what do you think of

```rust
#[derive(Debug, Default)]
#[cfg_attr(test, derive(Clone))]
pub struct Graph {
    dag: Dag<Release, Empty>,
}
```
?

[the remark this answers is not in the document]
lucab: let's go for that.

cincinnati/src/plugins/mod.rs
steveej: > It supposedly justifies the existence of these structs, as they're only used for that. Did it confuse you?

Please ignore me ;-)

steveej:
@lucab I cleaned up the git history into hopefully supposedly small enough commits. I am not convinced about the current error conversion between the different representations (see the *plugins/external/web.rs* for a demo usage) , but I want to postpone working on that until we have an actual web plugin to integrate.

PTAL :eyes:

cincinnati/src/plugins/internal/metadata_fetch_quay.rs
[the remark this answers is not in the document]
lucab: Answering first question, I'm usually settling for `asynchro`.

We discussed out-of-band that we will likely need some invasive work for the "async plugins" topic, which is mostly relevant for external plugins. It will be tackled in a followup PR as this one is already big enough.

cincinnati/src/plugins/mod.rs
lucab: Maybe `convert the PluginIO wrapper`? But I'm also fine as-is, it's understandable now.

cincinnati/src/plugins/mod.rs
lucab: Ack, let's defer this to the "eventually" point.

cincinnati/src/plugins/mod.rs
steveej: Let's leave it as is for a lack of a better word

cincinnati/src/plugins/mod.rs
```diff
@@ -0,0 +1,406 @@
...
+            PluginIO::InternalIO(internal_io) => internal_io.try_into()?,
+        };
+        Ok(final_io)
+    }
+}
+
+/// Wrapper struct for a universal implementaion of Plugin<PluginIO> for all InternalPlugin implementors
+pub struct InternalPluginWrapper<T>(pub T);
+
+/// Wrapper struct for a universal implementaiton of Plugin<PluginIO> for all ExternalPlugin implementors
```
lucab: Typo: "implementation".

cincinnati/src/plugins/internal/metadata_fetch_quay.rs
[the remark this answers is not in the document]
lucab: This still applies from previous review round.

cincinnati/src/plugins/mod.rs
```diff
@@ -0,0 +1,406 @@
...
+#[derive(Debug, PartialEq)]
+#[cfg_attr(test, derive(Clone))]
+pub struct ExternalIO {
+    pub bytes: Vec<u8>,
+}
+
+/// Trait which fronts InternalPlugin and ExternalPlugin, allowing their trait objects to live in the same collection
+pub trait Plugin<T>
+where
+    // <T as TryInto<PluginIO>>::Err: std::error::Error + std::marker::Send + std::marker::Sync,
```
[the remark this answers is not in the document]
steveej: leftover :+1:

cincinnati/src/plugins/interface.proto
```diff
@@ -0,0 +1,34 @@
...
+message PluginError {
+  enum Kind {
+    GENERIC = 0;
+    INVALID_GRAPH = 1;
+    INVALID_PARAM = 2;
+    FAILED_DEPENDENCY = 3;
+    INTERNAL_FAILURE = 4;
+  }
+  Kind kind = 1;
+  string details = 2;
```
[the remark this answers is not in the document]
steveej: Let's go for `value` since it's already in the spec.

lucab:
/lgtm
/hold

I think it's ready to land, but leaving you some room for second thoughts.

steveej:
@lucab thanks for the many reviews on this! 

/hold cancel

PR #51, opened 2019-01-29
graph-builder: fetch and add labels to releases and refactor for testability

steveej:
Split from #48.

Closes https://jira.coreos.com/browse/CORS-974.

/cc @lucab

### DIFF
--- a/graph-builder/src/graph.rs
+++ b/graph-builder/src/graph.rs
@@ -75,6 +75,7 @@ pub fn run(opts: &config::Options, state: &State) -> ! {
         let releases = match registry::fetch_releases(
             &opts.registry,
             &opts.repository,
+            &opts.quay_label_filter,
             username.as_ref().map(String::as_ref),
             password.as_ref().map(String::as_ref),
             &mut cache,
--- a/graph-builder/src/graph.rs
+++ b/graph-builder/src/graph.rs
@@ -16,7 +16,7 @@ use actix_web::http::header::{self, HeaderValue};
 use actix_web::{HttpMessage, HttpRequest, HttpResponse};
 use cincinnati::{AbstractRelease, Graph, Release, CONTENT_TYPE};
 use config;
-use failure::{Error, ResultExt};
+use failure::Error;
 use registry;
 use serde_json;
 use std::collections::{HashMap, HashSet};
--- a/graph-builder/src/graph.rs
+++ b/graph-builder/src/graph.rs
@@ -67,7 +67,10 @@ pub fn run(opts: &config::Options, state: &State) -> ! {
     // Read the credentials outside the loop to avoid re-reading the file
     let (username, password) =
         registry::read_credentials(opts.credentials_path.as_ref(), &opts.registry)
-            .expect("could not read credentials");
+            .expect("could not read registry credentials");
+
+    let quay_api_token = quay::read_credentials(opts.quay_api_credentials_path.as_ref())
+        .expect("could not read quay API credentials");
 
     loop {
         debug!("graph update triggered");
@@ -76,6 +79,7 @@ pub fn run(opts: &config::Options, state: &State) -> ! {
             &opts.registry,
             &opts.repository,
             &opts.quay_label_filter,
+            quay_api_token.as_ref().map(String::as_str),
             username.as_ref().map(String::as_ref),
             password.as_ref().map(String::as_ref),
             &mut cache,
--- a/graph-builder/src/graph.rs
+++ b/graph-builder/src/graph.rs
@@ -71,12 +71,31 @@ pub fn run(opts: &config::Options, state: &State) -> ! {
 
     loop {
         debug!("graph update triggered");
-        match create_graph(
-            &opts,
+
+        let releases = match registry::fetch_releases(
+            &opts.registry,
+            &opts.repository,
             username.as_ref().map(String::as_ref),
             password.as_ref().map(String::as_ref),
             &mut cache,
         ) {
+            Ok(releases) => {
+                if releases.is_empty() {
+                    warn!(
+                        "could not find any releases in {}/{}",
+                        &opts.registry, &opts.repository
+                    );
+                };
+                releases
+            }
+            Err(err) => {
+                err.iter_chain()
+                    .for_each(|cause| error!("failed to fetch all release metadata: {}", cause));
+                vec![]
+            }
+        };
+
+        match create_graph(releases) {
             Ok(graph) => match serde_json::to_string(&graph) {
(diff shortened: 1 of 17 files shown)


graph-builder/src/graph.rs
```diff
@@ -93,26 +112,9 @@ pub fn run(opts: &config::Options, state: &State) -> ! {
     }
 }
 
-fn create_graph(
-    opts: &config::Options,
-    username: Option<&str>,
-    password: Option<&str>,
-    cache: &mut HashMap<u64, Option<registry::Release>>,
-) -> Result<Graph, Error> {
+pub fn create_graph(releases: Vec<registry::Release>) -> Result<Graph, Error> {
```
lucab, line 120: Carry-over comment from the other PR: if the input releases are both unique and order-insensitive, then the argument here is better typed as an `HashSet`.

graph-builder/src/registry.rs
```diff
@@ -108,6 +109,8 @@ pub fn fetch_releases(
         .build()
         .map_err(|e| format_err!("{}", e))?;
 
+    let quay_client: quay::v1::Client = quay::v1::Client::builder().build()?;
```
lucab: Can you please wire an `Option<String>` from main to the `.access_token()` on the builder here?

lucab (commented):
Just two carry-over comments from #48.

graph-builder/src/graph.rs
steveej: I had the same thoughts and then I went down the rabbit whole, which I just did again to give you the error output :-)

```
error[E0277]: the trait bound `std::collections::HashMap<std::string::String, std::string::String>: std::hash::Hash` is not satisfied
  --> graph-builder/src/release.rs:30:5
   |
30 |     pub metadata: HashMap<String, String>,
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `std::hash::Hash` is not implemented for `std::collections::HashMap<std::string::String, std::string::String>`
   |
   = note: required by `std::hash::Hash::hash`
```

I'm happy to use a `HashSet` if you have a proper solution for the above.

graph-builder/src/registry.rs
steveej: I would avoid to have this token provided as a CLI flag for security reasons, so I suggest to read it either from a file or an environment variable, whereas a file is preferable to me. I haven't tried the authentication against the quay API yet. What format do we expect this token to have?

graph-builder/src/graph.rs
lucab: Ah. I went through the same rabbit hole multiple times and still haven't found a satisfying answer. Leave it as it is then.

graph-builder/src/registry.rs
lucab: I do agree with all of the above (that is, storing it in a file). To be honest if we don't need it right I'm fine with having just a `None` at the top caller, but having the rest of the logic already wired-in for the time when that requirement arrives. The token itself is a plain 41-chars ASCII string, without any apparent structure.

steveej: I've added all the infrastructure for it, although none of the tests work as we're lacking an API. The only concern I'd address in this PR if we expect the read of the file to work as is, e.g. do we have to strip EOL or EOF?

lucab: I would suggest to use a line reader, pick the first line, and `trim_end` on it. That would help users to avoid invisible typos, and us to avoid sending weird characters over HTTP.

graph-builder/src/registry.rs
```diff
@@ -121,17 +127,35 @@ pub fn fetch_releases(
...
+                &manifestref,
+                Some(&quay_label_filter),
+            ))?;
+
+            let release = Release {
+                metadata: Metadata {
+                    metadata: quay_labels
+                        .into_iter()
+                        .map(|label| (label.key, label.value))
+                        .chain(release.metadata.metadata)
```
lucab: I think this chaining is still in the wrong order, see https://github.com/openshift/cincinnati/pull/48#discussion_r251196071.

Did you maybe pick up an older version of your commits? The other PR seems updated and fine.

quay/src/lib.rs
```diff
@@ -8,4 +9,20 @@ extern crate serde;
...
 extern crate serde_json;
 
+use std::io::BufRead;
+use std::path::PathBuf;
+
 pub mod v1;
+
+pub fn read_credentials(credentials_path: Option<&PathBuf>) -> failure::Fallible<Option<String>> {
+    match &credentials_path {
+        Some(pathbuf) => match std::io::BufReader::new(std::fs::File::open(pathbuf)?)
```
lucab: Nit: this is a quite terse line and the whole function is pretty heavy on `Option`s matching.
I would personally make it more straight and focused on the positive case by:
 * passing a pathbuf directly (non-optional)
 * splitting the file opening on its own, and adding context for the error
 * introduce some `use`s to reduce visual cluttering
 * reduce the amount of `match`es
 * taking care of a possibly empty token at the end

That is, something like:
```rust
let fp = fs::File::open(pathbuf).context(...)?;
let first_line = io::BufReader::new(fp).lines().nth(0)
  .ok_or_else(|| "empty credentials.")?;
let token = first_line?.trim_end().to_string();

if !token.is_empty() {
  Ok(Some(token))
} else {
  Ok(None)
}
```

steveej: That's way too many lets for my taste :smirk: I think we've arrived in opinion-land

graph-builder/src/registry.rs
steveej: Thanks, I must've lost that change along the way, it's back in.

quay/src/lib.rs
lucab: To be clear, this wasn't just an opinion. This function is missing context on file-opening error and a final check on empty string, but I was worried you would go on nesting things here. Current code is artificially hard to review and follow, because the positive flow is implicit and mixed/split across match arms (while the underlying logic is linear).

It is fine to have separate `let` to split apart logical entities. You could collapse all `let`s in my example above if really wanted, but the goal should be to have it easy to read/maintain/debug, not to write as much FP as possible (which could be done by sprinkling a few more `map`/`and_then`, but I wouldn't recommend that).

steveej: Thanks for the hint on the missing checks. I restructured the code based on your suggestion with more error context. I've also added an error for an empty token, since that's surely not going to work, and in case a file is provided it's probably not intended either.

lucab: Ack, I agree that's more sensible.

PR #53, opened 2019-01-31
policy-engine: add basic tests for graph endpoint

lucab:
This adds initial test infrastructure and some (very basic) negative tests for policy-engine graph endpoint.

/cc @steveeJ

### DIFF
--- a/policy-engine/src/config.rs
+++ b/policy-engine/src/config.rs
@@ -5,6 +5,9 @@ use hyper::Uri;
 use std::collections::HashSet;
 use std::net::IpAddr;
 
+/// Default URL to upstream graph provider.
+pub(crate) static DEFAULT_UPSTREAM_URL: &str = "http://localhost:8080/v1/graph";
+
 #[derive(Debug, StructOpt)]
 pub struct Options {
     /// Verbosity level
@@ -12,7 +15,7 @@ pub struct Options {
     pub verbosity: u64,
 
     /// URL for the upstream graph builder or policy engine
-    #[structopt(long = "upstream", default_value = "http://localhost:8080/v1/graph")]
+    #[structopt(long = "upstream", raw(default_value = "DEFAULT_UPSTREAM_URL"))]
     pub upstream: Uri,
 
     /// Address on which the server will listen
--- a/policy-engine/src/graph.rs
+++ b/policy-engine/src/graph.rs
@@ -94,3 +94,46 @@ pub(crate) fn index(req: HttpRequest<AppState>) -> Box<Future<Item = HttpRespons
         }
     }
 }
+
+#[cfg(test)]
+mod tests {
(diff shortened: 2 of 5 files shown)


policy-engine/src/config.rs
```diff
@@ -5,14 +5,17 @@ use hyper::Uri;
...
+
 #[derive(Debug, StructOpt)]
 pub struct Options {
     /// Verbosity level
     #[structopt(short = "v", parse(from_occurrences))]
     pub verbosity: u64,
 
     /// URL for the upstream graph builder or policy engine
-    #[structopt(long = "upstream", default_value = "http://localhost:8080/v1/graph")]
+    #[structopt(long = "upstream", raw(default_value = "DEFAULT_UPSTREAM_URL"))]
```
steveej, line 18: This is neat. I've been looking for this for quite some time this morning!

policy-engine/src/graph.rs
```diff
@@ -94,3 +94,50 @@ pub(crate) fn index(req: HttpRequest<AppState>) -> Box<Future<Item = HttpRespons
...
+
+#[cfg(test)]
+mod tests {
+    extern crate tokio;
+
+    use self::tokio::runtime::current_thread::Runtime;
+    use crate::graph;
+    use crate::AppState;
+    use actix_web::http;
+    use actix_web::test as atest;
```
steveej: this doesn't seem worth the obfuscation

steveej (commented):
I haven't tested it but it reads fine, with a minor naming nit. Did you want to add more tests or keep the scope as is?

policy-engine/src/graph.rs
lucab: Ack, it was just feeling strange to end up with `test::`. Do you prefer the full form (`actix_web::test::<foo>`) or the short one but without renaming (`test::<foo>`)?

lucab:
@steveeJ for the moment I'd keep the scope like this. I'm still trying to figure out what's the proper way to mock the upstream request in the chain.

policy-engine/src/graph.rs
steveej: I'd prefer `actix_web::test` for the tests, shouldn't be too much overhead

policy-engine/src/graph.rs
```diff
@@ -94,3 +94,50 @@ pub(crate) fn index(req: HttpRequest<AppState>) -> Box<Future<Item = HttpRespons
...
+
+    #[cfg(feature = "test-net-private")]
+    mod private;
```
lucab: Self note: spurious lines.

lucab:
@steveeJ fixed the minor nit and also caught a copy-paste error. PTAL.

steveej (approved):
This is a nice testing approach.

/lgtm

PR #54, opened 2019-01-31
graph-builder: fetch quay labels conditionally

steveej:
Incomplete list of things done:

graph-builder: add metadata module
graph-builder: add CLI flag to disable quay labels
graph-builder: add CLI flag for quay API base
graph-builder/config: use appropriate constants as default values
graph-builder: fetch metadata in the main loop

I will split this into respective commits once we've iterated on these changes.

/cc @lucab

### DIFF
--- a/graph-builder/src/graph.rs
+++ b/graph-builder/src/graph.rs
@@ -17,6 +17,7 @@ use actix_web::{HttpMessage, HttpRequest, HttpResponse};
 use cincinnati::{AbstractRelease, Graph, Release, CONTENT_TYPE};
 use config;
 use failure::Error;
+use metadata::{fetch_and_populate_dynamic_metadata, MetadataFetcher, QuayMetadataFetcher};
 use registry::{self, Registry};
 use serde_json;
 use std::collections::{HashMap, HashSet};
@@ -60,7 +61,7 @@ impl State {
     }
 }
 
-pub fn run(opts: &config::Options, state: &State) -> ! {
+pub fn run<'a>(opts: &'a config::Options, state: &State) -> ! {
     // Grow-only cache, mapping tag (hashed layers) to optional release metadata.
     let mut cache = HashMap::new();
 
@@ -72,17 +73,29 @@ pub fn run(opts: &config::Options, state: &State) -> ! {
         registry::read_credentials(opts.credentials_path.as_ref(), &registry.host)
             .expect("could not read registry credentials");
 
-    let quay_api_token = quay::read_credentials(opts.quay_api_credentials_path.as_ref())
-        .expect("could not read quay API credentials");
+    let metadata_fetcher: Option<MetadataFetcher> = if opts.disable_quay_api_metadata {
+        debug!("Disable fetching of quay metadata..");
+        None
+    } else {
+        Some(
+            QuayMetadataFetcher::try_new(
+                opts.quay_label_filter.clone(),
+                opts.quay_api_credentials_path.as_ref(),
+                opts.quay_api_base.clone(),
+                opts.repository.clone(),
+            )
+            .expect("try_new to yield a metadata fetcher"),
+        )
+    };
+
+    let mut runtime = tokio::runtime::current_thread::Runtime::new().unwrap();
 
     loop {
         debug!("graph update triggered");
 
-        let releases = match registry::fetch_releases(
+        let mut releases = match registry::fetch_releases(
             &registry,
             &opts.repository,
-            &opts.quay_label_filter,
-            quay_api_token.as_ref().map(String::as_str),
             username.as_ref().map(String::as_ref),
             password.as_ref().map(String::as_ref),
             &mut cache,
@@ -102,6 +115,17 @@ pub fn run(opts: &config::Options, state: &State) -> ! {
                     .for_each(|cause| error!("failed to fetch all release metadata: {}", cause));
                 vec![]
             }
+        };
+
+        if let Some(metadata_fetcher) = &metadata_fetcher {
+            let populated_releases: Vec<registry::Release> = runtime
+                .block_on(fetch_and_populate_dynamic_metadata(
+                    metadata_fetcher,
+                    releases,
+                ))
+                .expect("fetch and population of dynamic metadata to work");
+
+            releases = populated_releases;
         };
 
         match create_graph(releases) {
--- a/graph-builder/src/metadata.rs
+++ b/graph-builder/src/metadata.rs
@@ -1,0 +1,102 @@
+//! This module implements the fetching of dynamic metadata
+
+use failure::{Error, Fallible};
+use futures::future::Future;
+use quay;
+use registry;
+use std::path::PathBuf;
+
+pub static DEFAULT_QUAY_LABEL_FILTER: &str = "com.openshift.upgrades.graph";
+pub static MANIFESTREF_KEY: &str = "com.openshift.upgrades.graph.release.manifestref";
+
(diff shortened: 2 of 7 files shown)


graph-builder/src/graph.rs
```diff
@@ -69,17 +72,29 @@ pub fn run(opts: &config::Options, state: &State) -> ! {
         registry::read_credentials(opts.credentials_path.as_ref(), &opts.registry)
             .expect("could not read registry credentials");
 
-    let quay_api_token = quay::read_credentials(opts.quay_api_credentials_path.as_ref())
-        .expect("could not read quay API credentials");
+    let metadata_fetcher: Option<MetadataFetcher> = if opts.disable_quay_api_metadata {
+        debug!("Disable fetching of quay metadata..");
+        None
+    } else {
+        Some(Rc::new(
```
steveeJ: @lucab I'm curious about your thoughts on `Rc` vs `Box` here

graph-builder/src/metadata.rs
```diff
@@ -0,0 +1,113 @@
...
+
+pub trait AsyncMetadataFetcher<T, E> {
+    fn fetch_metadata(&self, fetch_args: &MetadataFetchArgs) -> Box<Future<Item = T, Error = E>>;
+}
+
+/// Convenience type for a metadata entry
+type MetadataEntry = (String, String);
+
+/// Convenience type for fetch_args()'s metadata_fetcher argument
+pub type MetadataFetcher<'a> = Rc<AsyncMetadataFetcher<Vec<MetadataEntry>, Error> + 'a>;
```
steveej: @lucab I'm curious about your thoughts on `Rc` vs `Box` here

steveej: self answer: Rc is not needed here as these instances aren't meant to be owned by multiple sites

graph-builder/tests/net/quay_io/mod.rs
```diff
@@ -73,6 +76,16 @@ fn expected_releases(
     releases
 }
 
+fn remove_metadata_by_key(releases: &mut Vec<Release>, key: &str) -> Vec<Release> {
```
lucab: I'm not sure I follow the idea behind this, but I think this will result in a new cloned vector. If that is the case, I think this would better take ownership of the input vector and return it back after manipulation. Or otherwise just mutate the input one in-place.

graph-builder/tests/net/quay_io/mod.rs
```diff
@@ -73,6 +76,16 @@ fn expected_releases(
...
+    releases
+        .iter_mut()
+        .map(|release| {
+            release.metadata.metadata.remove(key).unwrap();
+            release.clone()
```
lucab: If there is a way to avoid both the `unwrap` and the `clone` in this closure, that would be better.

graph-builder/src/metadata.rs
```diff
@@ -0,0 +1,107 @@
...
+use failure::{Error, Fallible};
+use futures::future::Future;
+use quay;
+use registry;
+use std::path::PathBuf;
+
+pub static DEFAULT_QUAY_LABEL_FILTER: &str = "com.openshift.upgrades.graph";
+pub static MANIFESTREF_KEY: &str = "com.openshift.upgrades.graph.release.manifestref";
+
+pub struct MetadataFetchArgs {
```
lucab: This is effectively just a string. Do you foresee some different usecase for the future? If not, I think it makes sense to switch to a plain `String`.

graph-builder/src/metadata.rs
```diff
@@ -0,0 +1,107 @@
...
+    fn fetch_metadata(
+        &self,
+        fetch_args: &MetadataFetchArgs,
+    ) -> Box<Future<Item = Vec<MetadataEntry>, Error = Error>> {
+        Box::new(
+            self.client
+                .get_labels(
+                    self.repo.to_string(),
+                    fetch_args.manifestref.to_string(),
+                    Some(self.label_filter.to_string()),
```
lucab: It looks like you had to wire lifetimes through everywhere, but eventually the future needs to take ownership of these three parameters, thus cloning them here. It may be worth just taking ownership (in function inputs and in struct fields) and break the chain of lifetimes.

Up for your consideration though, I don't know which approach makes more sense. I just noticed it when following the lifetimes through.

lucab (commented):
Only concern is an `unwrap` which probably slipped through. Rest are few minor comments, mostly on ownership and cloning. The logic behind the PR as a whole seems fine.

graph-builder/src/metadata.rs
steveej: Indeed it doesn't make sense to go through the lifetime declarations and ending up cloning it in the end, thanks for pointing this one out :smile:

graph-builder/src/metadata.rs
```diff
@@ -0,0 +1,95 @@
...
+/// Asynchronously fetches and populates the dynamic metadata for the given releases
+pub fn fetch_and_populate_dynamic_metadata<'a>(
+    metadata_fetcher: &'a MetadataFetcher,
+    releases: Vec<registry::Release>,
+) -> impl Future<Item = Vec<registry::Release>, Error = Error> + 'a {
+    let populated_releases = releases.into_iter().map(move |mut release| {
+        futures::future::ok(release.metadata.metadata.remove(MANIFESTREF_KEY))
+            .and_then(|manifestref_value| match manifestref_value {
+                Some(manifestref) => Ok(manifestref),
+                None => bail!("release metadata doesn't contain the manifestref"),
```
lucab: Should we maybe just do this filtering upfront (outside of the future) and iterate on a vector of `(release, ref)`? I'm worried that a single entry without the corresponding manifest-ref will fail the whole `join_all`, while we could instead just skip the specific one.

steveej: This was intentional and I'm happy to discuss it. I assume that we have an incomplete view on the graph if we can't get *all* metadata, so I'd rather report a failure than a graph which was not intended to be published. Similarly it will error out if a single fetching operation fails.

lucab: I'm fine with this intentional behavior then. I would only suggest to make that more explicit (and actionable) to both human operators and casual code readers by:
 * having a docstring to mention the "either all or none" failure mode
 * adding the version id of the faulty release to the bail message

steveej:
@lucab comments addressed and rebased

graph-builder/src/metadata.rs
lucab: Ack, I do appreciate that docstring, thanks!

graph-builder/src/metadata.rs
```diff
@@ -0,0 +1,102 @@
...
+    metadata_fetcher: &'a MetadataFetcher,
+    releases: Vec<registry::Release>,
+) -> impl Future<Item = Vec<registry::Release>, Error = Error> + 'a {
+    let populated_releases = releases.into_iter().map(move |mut release| {
+        futures::future::ok(release.metadata.metadata.remove(MANIFESTREF_KEY))
+            .and_then(|manifestref_value| match manifestref_value {
+                Some(manifestref) => Ok(manifestref),
+                None => bail!(
+                    "metadata of release '{}' doesn't contain the manifestref",
+                    release.source
```
lucab, line 82: I *think* the compiler is complaining about this, because you are referencing `release` which however is not kept inside the future.

steveej: Haha yeah, I relied on the recompilation of the vscode rls instance which sometimes doesn't yield errors. Didn't expect an issue with that change :innocent:

PR #55, opened 2019-01-31
graph-builder/registry: don't enforce HTTPS

steveej:
Blocked by https://github.com/camallo/dkregistry-rs/issues/89.

### DIFF
--- a/graph-builder/src/registry.rs
+++ b/graph-builder/src/registry.rs
@@ -42,55 +42,157 @@ impl Into<cincinnati::Release> for Release {
     }
 }
 
-fn trim_protocol(src: &str) -> &str {
-    src.trim_left_matches("https://")
-        .trim_left_matches("http://")
+#[derive(Debug, Clone, PartialEq, Default)]
+pub struct Registry {
+    pub(crate) scheme: String,
+    pub(crate) insecure: bool,
+    pub(crate) host: String,
+    pub(crate) port: Option<u16>,
+}
+
+impl Registry {
+    pub fn try_from_str(src: &str) -> Fallible<Self> {
+        macro_rules! process_regex {
+            ($a:ident, $r:expr, $b:tt) => {
+                if let Some($a) = regex::Regex::new($r)
+                    .context(format!("could not compile regex pattern {}", $r))?
+                    .captures(src)
+                {
+                    $b
+                };
+            };
+        }
+
+        process_regex!(
+            capture,
+            // match scheme://h.o.s.t:port
+            r"^(?P<scheme>[a-z]+)(:/{2})(?P<host>([0-9a-zA-Z]+\.)*([0-9a-zA-Z]+)):(?P<port>[0-9]+)$",
+            {
+                let scheme = capture["scheme"].to_string();
+                return Ok(Registry {
+                    insecure: Registry::insecure_scheme(&scheme)?,
+                    scheme,
+                    host: capture["host"].to_string(),
+                    port: Some(
+                        capture["port"]
+                            .parse()
+                            .expect("could not parse port as a number"),
+                    ),
+                });
+            }
+        );
+
+        process_regex!(
+            capture,
+            // match scheme://h.o.s.t
+            r"^(?P<scheme>[a-z]+)(:/{2})(?P<host>([0-9a-zA-Z]+\.)*([0-9a-zA-Z]+))$",
+            {
+                let scheme = capture["scheme"].to_string();
+                return Ok(Registry {
+                    insecure: Registry::insecure_scheme(&scheme)?,
+                    scheme,
+                    host: capture["host"].to_string(),
+                    ..Default::default()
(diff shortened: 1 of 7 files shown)


lucab:
@steveeJ for this to work, you need to adjust tests to use `https://quay.io` too. I think this logic would better switch to use a `Url` though.

steveej:
> for this to work, you need to adjust tests to use `https://quay.io` too. I think this logic would better switch to use a `Url` though.

It came to me this morning too that this defaults to `http` if it's not specified. Which `Url` type do you mean? There's non in `std`.

steveej:
> It came to me this morning too that this defaults to `http` if it's not specified. Which `Url` type do you mean? There's non in `std`.

I've started to refactor the protocol trimming using `Url` create. Will push in a bit.

steveej:
We have a bug somewhere down the registry chain when using *http* as a scheme:

This is from a modified testcase:

```console
thread 'net::quay_io::fetch_release_public_with_no_release_metadata_must_not_error' panicked at 'should not error on emtpy repo: ErrorMessage { msg: "API v2 not supported" }', src/libcore/result.rs:1009:5
```

steveej:
The last test failures are legitimate as I've added a test which explicitly uses the `http` scheme. The blocking upstream issue is linked in the first comment.

steveej:
@lucab I've added a registry string parser since none of the existing URI/URL parsers work for this use-case. There's quite some duplicated code in the parser function which can be fairly simply reduced with a macro. I'll add that if we agree on the approach in general.

graph-builder/src/registry.rs
```diff
@@ -42,55 +42,148 @@ impl Into<cincinnati::Release> for Release {
...
+            });
+        };
+
+        if let Some(capture) = regex::Regex::new(
+            // match h.o.s.t
+            r"^(?P<host>([0-9a-zA-Z\-]+\.)*([0-9a-zA-Z]+))$",
+        )
+        .context("could not compile regex pattern")?
+        .captures(src)
+        {
```
lucab: Personal opinion: I had a brainfart reading this, maybe because there are five lines between the `if let` and its `{`. Naming captures result and having a single line `if let ... {` would make it straightforward to read. (But maybe I just need more coffee :)

graph-builder/src/registry.rs
```diff
@@ -42,55 +42,148 @@ impl Into<cincinnati::Release> for Release {
...
+                format!(":{}", port)
+            } else {
+                "".to_string()
+            }
+        )
+    }
+
+    fn insecure_scheme(scheme: &str) -> Fallible<bool> {
+        match scheme {
+            "https" | "docker" => Ok(false),
```
lucab: I don't think we want to allow a `docker` reference here, as the authority part is optional in some cases there (e.g. `busybox`).

graph-builder/src/registry.rs
```diff
@@ -394,3 +496,56 @@ fn assemble_metadata(blob: &[u8], metadata_filename: &str) -> Result<Metadata, E
...
     .map_err(Into::into)
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn registry_try_parse_valid() {
+        let tests = vec![
```
lucab, line 515: Can you please add an IPv4 test for `127.0.0.1`? The regexps should be already matching that.

graph-builder/src/registry.rs
```diff
@@ -42,55 +42,148 @@ impl Into<cincinnati::Release> for Release {
...
     })
 }
 
 pub fn authenticate_client(
-    client: &mut dkregistry::v2::Client,
+    dclient: &mut dkregistry::v2::Client,
     login_scope: String,
 ) -> impl Future<Item = &dkregistry::v2::Client, Error = dkregistry::errors::Error> {
-    client
-        .is_v2_supported()
```
lucab, line 67: This is missing in the new revision, just checking if it got dropped on purpose.

graph-builder/src/registry.rs
steveej: Since we've agreed on the approach I've refactored the duplicate code into a macro which should further obfuscate yet address this ;-)

graph-builder/src/registry.rs
steveej: I originally dropped this on purpose as I think that the the v2 client should throw an error if it can't talk to the registry as expected. I do understand that this check might be in scope of the registry module here so I'll bring it back in.

graph-builder/src/registry.rs
steveej: This won't be passed to dkregistry or elsewhere, it's just a safetycheck that the user doesn't expect us to fetch via `ftp` for example. I can remove it if it bothers you too much.

steveej:
@lucab I've addressed or answered all comments. PTAL

PR #56, opened 2019-02-06
Revert "*: rename payload -> image"

crawford:
This reverts commit 5d226dab3ca1a45cea13b2b6753cb08f6f05cb0d.

After some discussion, we've decided that the term "payload" is more
appropriate here given that not all consumers of Cincinnati will specify
their updates as an image reference. The idea of expressing the payload
as an extendable structure was also discussed. We determined that this
structure would need to remain opaque to the Cincinnati stack (since
different consumers have different requirements) and would ultimately
complicate things too much at this stage in the project.

### DIFF
--- a/cincinnati/src/lib.rs
+++ b/cincinnati/src/lib.rs
@@ -54,7 +54,7 @@ impl Release {
 #[derive(Debug, Deserialize, Serialize, PartialEq, Clone)]
 pub struct ConcreteRelease {
     pub version: String,
-    pub image: String,
+    pub payload: String,
     pub metadata: HashMap<String, String>,
 }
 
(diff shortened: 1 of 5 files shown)


steveej:
I'm in favor of calling it payload as the more generic term, so this 
/lgtm

I want to note that it has been bothering me a bit that we call it payload (or image) when it's really a reference to that. I mildly suggest we add `_ref` to the name for correctness, but if y'all think that's not worth it feel free to cancel the following
/hold

crawford:
It's not necessarily a reference to the payload though and in our cases it isn't even a reference to the payload (OpenShift is a reference to the _image_).

/hold cancel

PR #57, opened 2019-02-08
Cargo: bump dkregistry-rs (respect system CA certs)

steveej:
With the [completion of switch to reqwest in dkregistry-rs](https://github.com/camallo/dkregistry-rs/pull/95) the library now respects the system CA certs in all methods, and therefor the graph-builder will too :tada:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -411,14 +411,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "ct-logs"
-version = "0.4.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-dependencies = [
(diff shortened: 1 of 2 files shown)


steveej:
Let's wait for https://github.com/camallo/dkregistry-rs/pull/96.

/hold

steveej:
Unless, IIUC then [camallo/dkregistry-rs#96](https://github.com/camallo/dkregistry-rs/pull/96) doesn't have any effect on us since we're using the defaults anyway. @lucab can you confirm and release the hold and possibly lgtm too? :-)

lucab:
Yup, that one shouldn't affect the default library behavior, which is what cincinnati implicitly uses.

/lgtm

PR #58, opened 2019-02-12
commons: add custom errors with JSON responses

lucab:
This adds a custom `GraphError` error used by both graph-builder and policy-engine "graph" endpoints.
It provides a kind+value JSON serialization so that informative errors can be returned to clients.

Fixes: https://jira.coreos.com/browse/CORS-913
/cc @steveeJ

### DIFF
--- a/commons/src/errors.rs
+++ b/commons/src/errors.rs
@@ -1,0 +1,66 @@
+use actix_web::http;
+use actix_web::HttpResponse;
+
+#[derive(Debug, Fail, Eq, PartialEq)]
+pub enum GraphError {
+    #[fail(display = "failed to deserialize JSON: {}", _0)]
+    FailedJsonIn(String),
+    #[fail(display = "failed to serialize JSON: {}", _0)]
+    FailedJsonOut(String),
+    #[fail(display = "failed to fetch upstream graph: {}", _0)]
+    FailedUpstreamFetch(String),
+    #[fail(display = "failed to assemble upstream request")]
+    FailedUpstreamRequest,
+    #[fail(display = "invalid Content-Type requested")]
+    InvalidContentType,
+    #[fail(display = "mandatory client parameters missing")]
+    MissingParams,
+}
+
+impl actix_web::error::ResponseError for GraphError {
+    fn error_response(&self) -> HttpResponse {
+        self.as_json_error()
+    }
+}
+
+impl GraphError {
+    // Return the HTTP JSON error response.
+    pub fn as_json_error(&self) -> HttpResponse {
+        let code = self.as_status_code();
+        let json_body = json!({
+            "kind": self.as_kind(),
+            "value": self.as_value(),
+        });
+        HttpResponse::build(code).json(json_body)
+    }
+
+    // Return the HTTP status code for the error.
+    fn as_status_code(&self) -> http::StatusCode {
+        match *self {
+            GraphError::FailedJsonIn(_) => http::StatusCode::INTERNAL_SERVER_ERROR,
+            GraphError::FailedJsonOut(_) => http::StatusCode::INTERNAL_SERVER_ERROR,
+            GraphError::FailedUpstreamFetch(_) => http::StatusCode::INTERNAL_SERVER_ERROR,
+            GraphError::FailedUpstreamRequest => http::StatusCode::INTERNAL_SERVER_ERROR,
+            GraphError::InvalidContentType => http::StatusCode::NOT_ACCEPTABLE,
+            GraphError::MissingParams => http::StatusCode::BAD_REQUEST,
+        }
+    }
+
+    // Return the kind for the error.
+    fn as_kind(&self) -> String {
+        let kind = match *self {
+            GraphError::FailedJsonIn(_) => "failed_json_in",
+            GraphError::FailedJsonOut(_) => "failed_json_out",
+            GraphError::FailedUpstreamFetch(_) => "failed_upstream_fetch",
+            GraphError::FailedUpstreamRequest => "failed_upstream_request",
+            GraphError::InvalidContentType => "invalid_content_type",
+            GraphError::MissingParams => "missing_params",
+        };
+        kind.to_string()
+    }
+
+    // Return the value for the error.
+    fn as_value(&self) -> String {
+        format!("{}", self)
+    }
+}
--- a/commons/src/errors.rs
+++ b/commons/src/errors.rs
@@ -26,16 +26,16 @@ impl actix_web::error::ResponseError for GraphError {
 impl GraphError {
     // Return the HTTP JSON error response.
     pub fn as_json_error(&self) -> HttpResponse {
-        let code = self.as_status_code();
+        let code = self.status_code();
         let json_body = json!({
-            "kind": self.as_kind(),
-            "value": self.as_value(),
+            "kind": self.kind(),
+            "value": self.value(),
         });
         HttpResponse::build(code).json(json_body)
     }
 
     // Return the HTTP status code for the error.
-    fn as_status_code(&self) -> http::StatusCode {
+    fn status_code(&self) -> http::StatusCode {
         match *self {
             GraphError::FailedJsonIn(_) => http::StatusCode::INTERNAL_SERVER_ERROR,
             GraphError::FailedJsonOut(_) => http::StatusCode::INTERNAL_SERVER_ERROR,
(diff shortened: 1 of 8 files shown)


commons/src/errors.rs
```diff
@@ -0,0 +1,66 @@
...
+        let code = self.as_status_code();
+        let json_body = json!({
+            "kind": self.as_kind(),
+            "value": self.as_value(),
+        });
+        HttpResponse::build(code).json(json_body)
+    }
+
+    // Return the HTTP status code for the error.
+    fn as_status_code(&self) -> http::StatusCode {
```
steveej: I'd remove the `as_` prefix as this conversion is lossy

commons/src/errors.rs
```diff
@@ -0,0 +1,66 @@
...
+            GraphError::FailedJsonOut(_) => http::StatusCode::INTERNAL_SERVER_ERROR,
+            GraphError::FailedUpstreamFetch(_) => http::StatusCode::INTERNAL_SERVER_ERROR,
+            GraphError::FailedUpstreamRequest => http::StatusCode::INTERNAL_SERVER_ERROR,
+            GraphError::InvalidContentType => http::StatusCode::NOT_ACCEPTABLE,
+            GraphError::MissingParams => http::StatusCode::BAD_REQUEST,
+        }
+    }
+
+    // Return the kind for the error.
+    fn as_kind(&self) -> String {
```
steveej: I'd remove the `as_` prefix as this conversion is lossy

policy-engine/src/graph.rs
```diff
@@ -40,59 +39,65 @@ lazy_static! {
...
 
 /// Serve Cincinnati graph requests.
-pub(crate) fn index(req: HttpRequest<AppState>) -> Box<Future<Item = HttpResponse, Error = Error>> {
+pub(crate) fn index(
+    req: HttpRequest<AppState>,
+) -> Box<Future<Item = HttpResponse, Error = GraphError>> {
     HTTP_GRAPH_REQS.inc();
 
+    // Check that the client can accept JSON media type.
+    if let Err(e) = commons::ensure_content_type(req.headers(), CONTENT_TYPE) {
```
steveej, line 48: opportunity: combine this into something declarative with the next error check

policy-engine/src/graph.rs
```diff
@@ -40,59 +39,65 @@ lazy_static! {
...
+        HTTP_GRAPH_BAD_REQS.inc();
+        return Box::new(future::err(e));
+    }
+
     // Check for required client parameters.
     let mandatory_params = &req.state().mandatory_params;
-    if commons::ensure_query_params(mandatory_params, req.query_string()).is_err() {
+    if let Err(e) = commons::ensure_query_params(mandatory_params, req.query_string()) {
```
steveej, line 55: opportunity: combine this into something declarative with the previous error check

commons/src/lib.rs
```diff
@@ -40,14 +49,31 @@ pub fn ensure_query_params(params: &HashSet<String>, query: &str) -> Fallible<()
...
+/// Make sure client requested the relevant content type.
+pub fn ensure_content_type(
+    headers: &actix_web::http::HeaderMap,
+    content_type: &'static str,
+) -> Result<(), GraphError> {
+    let content_json = header::HeaderValue::from_static(content_type);
+
+    if !headers
+        .get(header::ACCEPT)
+        .map(|accept| accept == content_json)
```
steveej, line 68: opportunity: pull this into a declared table/struct of the accepted content types

steveej (commented):
This looks fine except for some naming nitpicks.

The "opportunity" comments are merely informative and for future reference if the code gets more complex.

policy-engine/src/graph.rs
lucab: I'm not sure I got what you are suggesting here. Would you like a chain of `and_then` checks or something different? I won't mind if you could very very briefly sketch what you have in mind with "something declarative".

commons/src/lib.rs
```diff
@@ -40,14 +49,31 @@ pub fn ensure_query_params(params: &HashSet<String>, query: &str) -> Fallible<()
         .collect();
 
     // Make sure all mandatory parameters are present.
-    ensure!(
-        params.is_subset(&query_keys),
-        "mandatory parameters missing"
-    );
+    if !required_params.is_subset(&query_keys) {
```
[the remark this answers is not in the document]
lucab: There was an error-to-string coercion, if I remember correctly. But I will re-check, as I may have overlooked this midway.

policy-engine/src/graph.rs
steveej: I thought about something like:

```rust
    for checks in vec![
        || commons::ensure_content_type(req.headers(), CONTENT_TYPE),
        || commons::ensure_query_params(mandatory_params, req.query_string()),
    ] {
        if let Err(e) = check() {
            HTTP_GRAPH_BAD_REQS.inc();
        }
    }
```

I tried this quickly but it's broken due to closures not matching in type.

commons/src/lib.rs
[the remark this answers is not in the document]
lucab: More precisely, there was a type mismatch:
```
expected type `errors::GraphError`
found type `failure::Error`
```

commons/src/lib.rs
lucab: I started plumbing through an `HashSet` for this, but I'm not totally satisfied with that.
Sticking close to the current logic, we would be constructing that single-member set on each request (without any actual need to).
We could either move that into AppState, or tackle this later on only if we actually end up with a multi-content endpoint.

Any further feedback/preference from your side?

steveej: Let's stick to the current approach since we have no plans to handle different content types right now.

commons/src/lib.rs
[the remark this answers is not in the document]
steveej: That's a bit inconvenient. Have you research if there's a trait in ˋfailureˋ  you can implement on ˋGraphErrorˋ to make it compatible?

commons/src/lib.rs
lucab: Ack.

commons/src/lib.rs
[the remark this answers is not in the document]
lucab: Indeed. We already derive `Fail`, but the macro enforces `Error` which is a fixed struct instead. I'm still having a look around, but I think this is a long-standing bug in failure itself: https://github.com/rust-lang-nursery/failure/issues/110.

lucab:
@steveeJ naming was indeed quite off, thanks for highlighting that. On the `ensure` and declarative-checks sides I unfortunately don't have much more success to report, thus left them as-is but open to further ideas. PTAL.

commons/src/lib.rs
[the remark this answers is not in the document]
steveej: Ah that's unfortunate, thanks for looking into this though!

steveej (commented):
Considering all my previous comments as resolved.
Going over this again it occurred to me that these changes aren't reflected on the `openapi` endpoint in the policy-engine, as well as in the docs. It seems to fit in this PR too.

lucab:
@steveeJ good point, let me add those.

/hold

lucab:
Updated error type in openapi and added a dedicated section to describe it in `cincinnati.md`.

/hold cancel

steveej:
I think it would be nice to add the specific responses from https://github.com/openshift/cincinnati/pull/58/files#diff-1bb65de438c6b2131604dcc53ca74722R38 to the response object of the graph endpoint. 

@lucab what are your thoughts on that?

lucab:
@steveeJ I've piled an additional commit on top with well-known codes. I think it's nice to have but not strictly required. My only concern is that as the openapi is not autogenerated from the type, the two could drift apart in the future. I've left the `default` case in place to minimize the problems in that case.

steveej:
@lucab I see your concerns with the drift. In a perfect world we'd generate the code from the spec; switching to that model would be pretty invasive at this point though. Thanks for adding the errors to the spec. One last comment about that is that it's not yet helpful for a machine reading the spec, as the type is the same for all errors. I'm fine leaving it as is in this PR, as it already helps humans ;-)

lucab:
> type is the same for all errors

From my side, this is on purpose **at the protocol level**, and it models the `Result<Graph, GraphError>` which is the single output of the graph endpoint. The **transport** point of view is not concerned with inner variants.

steveej:
> From my side, this is on purpose **at the protocol level**, and it models the `Result<Graph, GraphError>` which is the single output of the graph endpoint. The **transport** point of view is not concerned with inner variants.

Understood, that matches with everything being loosely typed via a string value. Thanks for clarifying!

/lgtm

PR #59, opened 2019-02-12
graph-builder: untangle the run-loop

steveej:
This restructure makes the main loop to be `continue`'able and therefore
more easy to extend.

This is in preparation for pulling the label processing out of the
`create_graph()` function into the main loop.

/cc @lucab

### DIFF
--- a/graph-builder/src/graph.rs
+++ b/graph-builder/src/graph.rs
@@ -90,7 +90,16 @@ pub fn run<'a>(opts: &'a config::Options, state: &State) -> ! {
 
     let mut runtime = tokio::runtime::current_thread::Runtime::new().unwrap();
 
+    // Don't wait on the first iteration
+    let mut first_iteration = true;
+
     loop {
+        if first_iteration {
+            first_iteration = false;
+        } else {
+            thread::sleep(opts.period);
+        }
+
         debug!("graph update triggered");
 
         let mut releases = match registry::fetch_releases(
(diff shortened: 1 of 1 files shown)


graph-builder/src/graph.rs
```diff
@@ -90,9 +90,16 @@ pub fn run<'a>(opts: &'a config::Options, state: &State) -> ! {
...
 
+    // Don't wait on the first iteration
+    let mut period = std::time::Duration::from_secs(0);
+
     loop {
+        thread::sleep(period);
+        period = opts.period;
         debug!("graph update triggered");
 
+        let mut graph: Option<cincinnati::Graph> = None;
```
lucab: I think you don't need this one, as `graph` is not persisted/accessed across loops here. A `let graph = match create_graph()` should be enough.

steveej: I wanted to be explicit about this being possibly assigned by multiple statements in the loop. For now it's only one assignment but it'll be more soon. Anyway, I don't mind removing this and go for variable shadowing instead.

lucab: If that is the case, then feel free to keep it. The comment above was because I'm generally skeptical on long-living mutable variables as it is easy to introduce logic bugs when the business logic gets more complex.

steveej: It's already gone ;-) I'll just shadow it as many times as necessary

PR #62, opened 2019-02-27
Correction: Add labels field and move actual label inside it

aditya-konarde:
Add the labels field and add the label inside it. 

@steveeJ


steveej:
[test]

PR #63, opened 2019-02-28
Add prometheus port to the containers and services

aditya-konarde:
Add the port that the metrics will be exposed on.
The ports are named:
- prometheus-policy-engine (port 9081)
- prometheus-graph-builder (port 9080)

Since it is required that all ports must be named in a multi-port
service, add a name to the policy-engine and graph-builder ports.

Also expose the ports via the service, so that the endpoints are
added, and servicemonitors can select them

### DIFF
--- a/dist/openshift/cincinnati.yaml
+++ b/dist/openshift/cincinnati.yaml
@@ -58,7 +58,7 @@ objects:
           ports:
           - name: graph-builder
             containerPort: ${{GB_PORT}}
-          - name: prometheus-graph-builder
+          - name: status-graph-builder
             containerPort: 9080
           livenessProbe:
             httpGet:
@@ -111,7 +111,7 @@ objects:
           ports:
           - name: policy-engine 
             containerPort: ${{PE_PORT}}
-          - name: prometheus-policy-engine
+          - name: status-policy-engine
             containerPort: 9081
           livenessProbe:
             httpGet:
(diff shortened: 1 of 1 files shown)


aditya-konarde:
cc @riuvshin

riuvshin:
this looks good to me, @steveeJ was mentioning that there is an arg to specify metrics port, do we need this for custom port 9081?

steveej:
> do we need this for custom port 9081?

This is the default: https://github.com/openshift/cincinnati/blob/19b79bed69abf0a0d371dbd19ce6bb42c0178ae9/policy-engine/src/config.rs#L34

The PR should work as is, except that the graph-builder doesn't expose any metrics yet.

lucab:
@aditya-konarde we may add a few more status/health introspection endpoints on those ports, which aren't specifically tied to Prometheus. Can we call them `status-policy-engine` and `status-graph-builder` maybe?

Sorry for the bikeshedding. Except for that, it looks good.

aditya-konarde:
As of now: yes, sure. I can rename it to status-*

Future/longer-term: We'd like to somewhat standardize on the port names and numbers for metrics, as it works easier for an SRE team looking at multiple services to have a definitive place to look in case something goes wrong. 

I'll update the PR, but I hope you understand that we may need to change it in the future. Lets change it when we get there :)

aditya-konarde:
Done. 

Btw, we're good to merge it, even if nothing is exposed on the ports yet. We can at least test out the policy builder metrics that way.

steveej:
Thanks @aditya-konarde! 

/lgtm

PR #64, opened 2019-03-06
Fix long status port names

aditya-konarde:
There's a 15 characters limit on the port names.

This commit shortens the port names to fix the deployment from failing

status-policy-engine -> status-pe
status-graph-builder -> status-gb

### DIFF
--- a/dist/openshift/cincinnati.yaml
+++ b/dist/openshift/cincinnati.yaml
@@ -58,7 +58,7 @@ objects:
           ports:
           - name: graph-builder
             containerPort: ${{GB_PORT}}
-          - name: status-graph-builder
+          - name: status-gb
             containerPort: 9080
           livenessProbe:
             httpGet:
@@ -111,7 +111,7 @@ objects:
           ports:
           - name: policy-engine 
             containerPort: ${{PE_PORT}}
-          - name: status-policy-engine
+          - name: status-pe
             containerPort: 9081
           livenessProbe:
             httpGet:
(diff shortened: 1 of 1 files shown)


aditya-konarde:
Actual error message from build-master: https://ci.ext.devshift.net/job/openshift-cincinnati-gh-build-master/48/console

```
The DeploymentConfig "cincinnati" is invalid: 
* spec.template.spec.containers[0].ports[1].name: Invalid value: "status-graph-builder": must be no more than 15 characters
* spec.template.spec.containers[1].ports[1].name: Invalid value: "status-policy-engine": must be no more than 15 characters
```

steveej:
Oh, I didn't see that coming. Are there any validation tools for this file we could run in the CI?

aditya-konarde:
@steveeJ I'll check if we could've caught this with a `oc apply --dry-run`. We should indeed add it to all the jobs. CC @jmelis

aditya-konarde:
PS: It doesn't 
```
-          - name: status-pe
+          - name: status-ajsdlkjaslkdjkasjdlkjlkasjdlkj
```
After which I applied
```
[akonarde@katana openshift]$ oc process -f cincinnati.yaml | oc apply --dry-run -f -
deploymentconfig.apps.openshift.io/cincinnati created (dry run)
```

steveej:
Possibly the 
```
--validate=false: If true, use a schema to validate the input before sending it
```
flag to apply?

aditya-konarde:
Doesn't help: 
```
[akonarde@katana openshift]$ oc process -f cincinnati.yaml | oc apply --validate=true --dry-run -f -
deploymentconfig.apps.openshift.io/cincinnati created (dry run)
```

I'll add a bug on origin-tools against `oc` to see what they have to say. Meanwhile this PR should be good to go.

steveej:
> I'll add a bug on origin-tools against `oc` to see what they have to say.

Thanks for looking into this. Validating the templates on CI will be very useful for all teams!

/lgtm

PR #65, opened 2019-03-07
Use default port numbers, add specific app labels to services

aditya-konarde:
We initially got the status ports interchanged. The correct port mappings are:

status-graph-builder: 9081
status-policy-engine: 9080

Also adds a specific app label to the service, so that we can select
by that label in the servicemonitor.

### DIFF
--- a/dist/openshift/cincinnati.configmap.yaml
+++ b/dist/openshift/cincinnati.configmap.yaml
@@ -9,5 +9,5 @@ data:
   gb.repository: "steveej/cincinnati-test"
   gb.log.verbosity: "vvv"
   pe.address: "0.0.0.0"
-  pe.upstream: "http://cincinnati-graph-builder.cincinnati-staging.svc:8181/v1/graph"
+  pe.upstream: "http://cincinnati-graph-builder.cincinnati-staging.svc:8080/v1/graph"
   pe.log.verbosity: "vvv"
--- a/dist/openshift/cincinnati.yaml
+++ b/dist/openshift/cincinnati.yaml
@@ -147,7 +147,7 @@ objects:
   metadata:
     name: cincinnati-graph-builder
     labels:
-      app: cincinnati
+      app: cincinnati-graph-builder
   spec:
     ports:
       - name: graph-builder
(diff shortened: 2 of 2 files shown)


aditya-konarde:
@steveeJ @lucab Prometheus was able to discover cincinnati, but not scrape the metrics. After some troubleshooting I found out that the status ports were actually the other way. This PR fixes that. Please take a look and see if the ports are as expected now

steveej:
I think the current code in the repo is correct. Check out the default:

https://github.com/openshift/cincinnati/blob/d4d759286ffd3acd4119e55d698f63e82bbeb47b/policy-engine/src/config.rs#L34

What's the issue you're seeing?

steveej:
In an OOB meeting we concluded that we should use the default ports in the openshift templates instead of the currently rather arbitrary and confusing choices.

steveej:
> In an OOB meeting we concluded that we should use the default ports in the openshift templates instead of the currently rather arbitrary and confusing choices.

As a clarification, we decided to leave the ports in the application as is. We tune the `--port` arguments given in the openshift template to match the default ports and change the service configs accordingly.

@aditya-konarde do you want to change the template within this PR or should I open a new one?

/assign

aditya-konarde:
@steveeJ I'm on it now, will change the template to use the standard ports. We'll also need to change the service and route to go to the correct port.

dist/openshift/cincinnati.yaml
```diff
@@ -192,10 +192,10 @@ parameters:
   displayName: Memory Limit
   description: Maximum amount of memory the container can use. Defaults 256Mi
 - name: GB_PORT
-  value: "8181"
+  value: "8180"
```
steveej: did you mean `8080`?

dist/openshift/cincinnati.configmap.yaml
```diff
@@ -9,5 +9,5 @@ data:
   gb.repository: "steveej/cincinnati-test"
   gb.log.verbosity: "vvv"
   pe.address: "0.0.0.0"
-  pe.upstream: "http://cincinnati-graph-builder.cincinnati-staging.svc:8181/v1/graph"
+  pe.upstream: "http://cincinnati-graph-builder.cincinnati-staging.svc:8081/v1/graph"
```
steveej: This should be `8080` since the policy-engine (PE) considers the graph-builder (GB) as its upstream, which runs on `8080`

steveej:
This seems fine now, thanks!

/lgtm

PR #66, opened 2019-03-08
docs: add instructions for inspecting the graph

crawford:
This provides a jq expression that allows one to determine the valid
tranversals from a graph. This is useful for large graphs where it would
be too time consuming to do by hand.


steveej:
Thanks for this helpful example!

/lgtm

PR #68, opened 2019-03-11
Use template parameters for status ports

aditya-konarde:
In order to maintain consistency with how the server ports are specified,
make use of template parameters to specify the status ports


steveej:
Very nice, thanks for restoring consistency! :-)

/lgtm

PR #69, opened 2019-03-12
graph-builder: add support for configuration file 

lucab:
This reworks plugins setup and graph-builder CLI parser to allow for sourcing complex/structured options from configuration file(s). In particular, configuration logic now cares about the following entities:
 * "options": configuration fragments (CLI flags, file snippets), optional and stringly-typed.
 * "unified config": configuration document, result of merging all options and defaults.
 * "app settings": runtime settings, result of config validation.

Ref: https://jira.coreos.com/browse/CORS-967
/cc @steveeJ


lucab:
Rebased to master and added graph-builder configuration options for https://github.com/openshift/cincinnati/pull/72#issuecomment-473327021.

graph-builder/src/main.rs
```diff
@@ -16,54 +16,43 @@ extern crate actix;
...
-use log::LevelFilter;
-use std::{net, thread};
-use structopt::StructOpt;
+use std::{thread};
 
 fn main() -> Result<(), Error> {
     let sys = actix::System::new("graph-builder");
 
-    let opts = config::Options::from_args();
+    let opts = config::AppSettings::assemble()?;
```
lucab: Self note: s/opts/settings/.

cincinnati/src/plugins/registry.rs
```diff
@@ -0,0 +1,35 @@
+//! Plugin registry.
```
lucab: Self note: check if there is a better term instead of "registry" (which could be confusing due to docker-registry).

lucab:
@steveeJ I've dropped all the plugins logic from here as requested, and split that to https://github.com/openshift/cincinnati/pull/74.

graph-builder/src/config/unified.rs
```diff
@@ -0,0 +1,156 @@
...
+    ( $dst:expr, $src:expr ) => {{
+        if let Some(x) = $src {
+            $dst = x;
+        };
+    }};
+}
+
+/// Top-level configuration (before semantic validation).
+#[derive(Debug)]
+pub struct UnifiedConfig {
```
steveej: Is there a reason for why we can't have `Option`s in here?

graph-builder/src/config/unified.rs
```diff
@@ -0,0 +1,156 @@
...
+static DEFAULT_REGISTRY_PERIOD: u64 = 30;
+static DEFAULT_REGISTRY_CREDENTIALS_PATH: &str = "";
+static DEFAULT_REGISTRY_MANIFESTREF_KEY: &str = "com.openshift.upgrades.graph.release.manifestref";
+static DEFAULT_REGISTRY_URL: &str = "http://localhost:5000";
+static DEFAULT_REGISTRY_REPO: &str = "openshift";
+
+static DEFAULT_STATUS_ADDR: &str = "127.0.0.1";
+static DEFAULT_STATUS_PORT: u16 = 9080;
+
+macro_rules! maybe_assign {
```
steveej: how about `assign_if_some` here? the second `_` is emitable if you think it's too long. I think it would be a more intuitive name, as, at least for me, _maybe_ didn't tell me what the condition is.

graph-builder/src/config/unified.rs
```diff
@@ -0,0 +1,156 @@
...
+        maybe_assign!(merged_cfg.manifestref_key, cfg.registry.manifestref_key);
+
+        Ok(merged_cfg)
+    }
+
+    /// Merge TOML options into unified configuration.
+    pub(crate) fn merge_file_config(self, file_cfg: Option<FileOptions>) -> Fallible<Self> {
+        let mut merged_cfg = self;
+
+        // Eaarly-return without updates if there is no configuration file.
```
steveej: s/Eaarly/Early/

graph-builder/src/config/unified.rs
```diff
@@ -0,0 +1,156 @@
...
+            }
+        }
+
+        // Status service options.
+        if let Some(status) = cfg.status {
+            maybe_assign!(merged_cfg.status_address, status.address);
+            maybe_assign!(merged_cfg.status_port, status.port);
+        }
+
+        // Plugins options. Order is relevant too.
```
steveej: Are the subsequent plugin options leftover in this PR or required?

graph-builder/src/graph.rs
```diff
@@ -88,49 +88,47 @@ impl State {
...
-                    .expect("could not initialize the QuayMetadataPlugin"),
-                )),
-                Box::new(InternalPluginWrapper(NodeRemovePlugin {
-                    key_prefix: opts.quay_label_filter.clone(),
-                })),
-                Box::new(InternalPluginWrapper(EdgeAddRemovePlugin {
-                    key_prefix: opts.quay_label_filter.clone(),
-                })),
-            ]
+    let configured_plugins: Vec<Box<plugins::Plugin<plugins::PluginIO>>> = {
```
steveej: The condition for gating the plugins got lost here

graph-builder/src/config/cli.rs
```diff
@@ -0,0 +1,83 @@
...
+    pub path_prefix: Option<String>,
+
+    /// Comma-separated set of mandatory client parameters
+    #[structopt(long = "service.mandatory_client_parameters")]
+    pub mandatory_client_parameters: Option<String>,
+}
+
+/// CLI configuration flags, Docker-registry fetcher.
+#[derive(Debug, StructOpt)]
+pub struct UpstreamRegistryOptions {
```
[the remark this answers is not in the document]
lucab: I don't plan to reuse this; however, I'm bad at naming things :(

This is technically a "DockerRegistryV2" upstream, and there are likely a "[OciDistributionVx](https://github.com/opencontainers/distribution-spec)" and a "FedoraCoreos" coming up at some point in the future. I went for `registry` only because it was shorter for options.

graph-builder/src/config/file.rs
```diff
@@ -0,0 +1,99 @@
...
+            cfg_path.as_ref()
+        ))?;
+        let mut bufrd = io::BufReader::new(cfg_file);
+
+        let mut content = vec![];
+        bufrd.read_to_end(&mut content)?;
+        let cfg = toml::from_slice(&content).context(format!(
+            "failed to read config file {}",
+            cfg_path.as_ref().display()
+        ))?;
```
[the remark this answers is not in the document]
lucab: Not directly because the expected error type is a `failure::Error` while this is a `failure::Context`. I could chain some error-mapping to massage types, but I think it's more straightforward to read in its current form.

graph-builder/src/config/file.rs
```diff
@@ -0,0 +1,99 @@
...
+
+        Ok(cfg)
+    }
+}
+
+/// TOML configuration, upstream fetcher.
+#[derive(Debug, Deserialize)]
+pub struct UpstreamOptions {
```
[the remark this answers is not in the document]
lucab: For some sections I think we can, yes. Top-level and plugins will be separate, but the others can be shared and derive `Debug + Deserialize + StructOpt` in an common `options.rs`.

graph-builder/src/config/unified.rs
lucab: Is there anything specific that you think should be an `Option`?

This struct holds configuration entries, after merging together defaults+toml+cli. Unless something here requires a real `None`, most of the entries will have real values (either from user-options or from defaults).

graph-builder/src/graph.rs
lucab: This is currently unused in our deployment, and #74 takes care of that (plugins not in the config are not enabled). I dropped the CLI flag here.

lucab:
@steveeJ thanks for the review, I've taken most of your suggestions and amended the code. In particular, CLI and TOML are now sharing their types, and I've split a minor enhancement to https://github.com/openshift/cincinnati/pull/76. PTAL.

graph-builder/src/config/unified.rs
steveej: I would imagine that this struct could be a mere composition of the ones we already have for CLI and file, so this would resemble the same structure everywhere. I understand this places the burden of handling the `Option`s again in the code :thinking:

graph-builder/src/config/file.rs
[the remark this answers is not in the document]
steveej: Now `UpstreamOptions` is only available for file configuration, correct?

graph-builder/src/config/file.rs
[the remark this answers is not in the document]
steveej: So the `?` handles that conversion automatically?

lucab:
@steveeJ I did a major rework on the PR in the direction of https://github.com/openshift/cincinnati/pull/69#discussion_r268681483. This means that the `Unified` struct does not exist anymore, input parsing goes via structopt+serde, and most of the structs are re-used.

While this definitely still needs some polishing, can you please another pass to see if suits your taste better than the previous iteration?

graph-builder/src/config/options.rs
```diff
@@ -0,0 +1,128 @@
...
+    /// Name of the container image repository
+    #[structopt(long = "upstream.registry.repository", alias = "repository")]
+    pub repository: Option<String>,
+
+    /// Credentials file for authentication against the image registry
+    #[structopt(
+        long = "upstream.registry.credentials_path",
+        alias = "credentials-file"
+    )]
+    pub credentials_path: Option<PathBuf>,
```
lucab, line 72: Self-note: this only works on toolchain >= 1.32.

graph-builder/src/config/settings.rs
```diff
@@ -0,0 +1,92 @@
...
+    #[default(IpAddr::V4(Ipv4Addr::LOCALHOST))]
+    pub address: IpAddr,
+
+    /// Optional auth secrets for the registry scraper.
+    #[default(None)]
+    pub credentials_path: Option<PathBuf>,
+
+    /// Required client parameters for the main service.
+    #[default(_code = "HashSet::new()")]
+    pub mandatory_client_parameters: HashSet<String>,
```
steveej, line 21: Might it simplify/streamline the rest of the code if this is wrapped in an `Option` too?

graph-builder/src/config/settings.rs
```diff
@@ -0,0 +1,92 @@
...
+    /// Listening address for the main service.
+    #[default(IpAddr::V4(Ipv4Addr::LOCALHOST))]
+    pub address: IpAddr,
+
+    /// Optional auth secrets for the registry scraper.
+    #[default(None)]
+    pub credentials_path: Option<PathBuf>,
+
+    /// Required client parameters for the main service.
+    #[default(_code = "HashSet::new()")]
```
steveej: The comment below will probably influence this, but in the current stat we could use `Default::default()` here

graph-builder/src/config/settings.rs
```diff
@@ -0,0 +1,92 @@
...
+    pub mandatory_client_parameters: HashSet<String>,
+
+    /// Metadata key where to record the manifest-reference.
+    #[default("com.openshift.upgrades.graph.release.manifestref")]
+    pub manifestref_key: String,
+
+    /// Endpoints namespace for the main service.
+    #[default("")]
```
steveej: `Default::default()`?

graph-builder/src/config/file.rs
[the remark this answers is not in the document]
steveej: For reference: we've agreed that we don't want to complicate the CLI flags more and thus keep this only in the config file for now. We might revisit later.

graph-builder/src/graph.rs
steveej: I needed the flag for testing the integration with Satellite. That might come up again anytime soon so I'm in favor to leave something in to gate this.

steveej (commented):
The overall approach looks very clean to me, especially the simplicity of the `assemble` function.
I would've wished for some code generation (macros) for the merge implementations but I wouldn't block on that if there exists no ready-made solution right now.

I have added a couple of comments to be addressed and there are also some leftover from the previous review. Thanks for exploring this so thoroughly!

graph-builder/src/config/options.rs
lucab: Split to https://github.com/openshift/cincinnati/pull/77.

graph-builder/src/config/settings.rs
lucab: Here and above, `Default::default()` is implicit if I remove the annotation (but I liked it explicit). Do you prefer me to drop all the non-custom default annotations?

graph-builder/src/config/options.rs
```diff
@@ -0,0 +1,128 @@
...
+
+    /// Metadata key where to record the manifest-reference
+    #[structopt(long = "upstream.registry.manifestref_key")]
+    pub manifestref_key: Option<String>,
+}
+
```
steveeJ: need a docstring here

graph-builder/src/config/mod.rs
```diff
@@ -0,0 +1,26 @@
...
+}
+
+mod cli;
+mod file;
+mod options;
+mod settings;
+
+pub use self::settings::AppSettings;
+
+trait Merge<T> {
```
steveej: need a docstring for this

graph-builder/src/config/settings.rs
lucab: Discussed out-of-band: let's use the implicit form.

graph-builder/src/config/options.rs
steveej: So a rebase should fix the build failure :+1:

graph-builder/src/config/settings.rs
lucab: Not really, it will even make set-appending nested into an additional level of option-matching.

lucab:
I added the final tests and documentation, completing the last TODOs. Dropped the WIP label, this is ready for the last round of review.

docs/user/graph-builder-configuration.md
```diff
@@ -0,0 +1,25 @@
...
+   - `port` (unsigned integer): local port for the main service. Default: 8080.
+ - `status` (section): configuration options related to the HTTP status service.
+   - `address` (string): local IP for the status service. Default: "127.0.0.1".
+   - `port` (unsigned integer): local port for the status service. Default: 9080.
+ - `upstream` (section): configuration options related to upstream release-data provider.
+   - `method` (string): upstream provider selector. Allowed values: "registry". Default: "registry".
+   - `registry` (section): configuration for Docker-v2 registry provider.
+     - `credentials_path` (string): path to file containing registry credentials. Default: unset. 
+     - `manifestref_key` (string): metadata key where to record the manifest-reference. Default: "com.openshift.upgrades.graph.release.manifestref".
+     - `period` (unsigned integer): pause between repository scrapes, in seconds. Default: 30.
```
steveej: should we rename this now while we're at it? to either "pause" or "delay"?

steveej (dismissed):
thanks for the nice rework!

there's a small doc change I'd like to make and a couple of questions/suggestions I added (please feel free to dismiss the suggestions).

docs/user/graph-builder-configuration.md
lucab: If you are fine with mixing the rename in here, I'm happy to do that. I also wanted to mention the time unit here. What about `pause_secs`?

docs/user/graph-builder-configuration.md
```diff
@@ -0,0 +1,25 @@
+# Graph-builder configuration
```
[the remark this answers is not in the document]
lucab: I think it is technically feasible somehow (worst case, via sed), but I am not aware of a clean way to do it right now. I'll chalk this to a ticket for later.

docs/user/graph-builder-configuration.md
steveej: `pause_secs` seems fine

graph-builder/src/config/file.rs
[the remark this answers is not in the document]
lucab: Eventually everything converts to a `Result` and `std::error::Error` trait. There a few layers to go through in `failure`, though. I don't know the exact path that `?` goes through, but it is pretty effective at doing those conversions automatically, yes.

docs/user/graph-builder-configuration.md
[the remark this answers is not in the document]
lucab: Moved to https://jira.coreos.com/browse/CIN-21

lucab:
Pushed another update addressing last comments, PTAL

steveej (approved):
:tada: 

/lgtm

PR #70, opened 2019-03-13
graph-builder/tests: fix building net-private tests

lucab:
This fixes "net-private" tests, which are currently failing to build.

### DIFF
--- a/graph-builder/tests/net/quay_io/mod.rs
+++ b/graph-builder/tests/net/quay_io/mod.rs
@@ -8,12 +8,13 @@ extern crate url;
 use self::graph_builder::registry::{self, fetch_releases, Release};
 use self::graph_builder::release::{Metadata, MetadataKind::V0};
 use self::semver::Version;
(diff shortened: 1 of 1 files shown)


steveej:
AFAIR the tests are also semantically broken, can you confirm or deny that?

lucab:
@steveeJ on this PR, with a local `cargo test --all-features --no-fail-fast` I only see a failure on `net::quay_io::fetch_release_private_with_credentials_must_succeed` (because I don't have the secrets on disk).

steveej:
I'm pretty sure the test itself is currently broken, see the TODO in it. I created https://jira.coreos.com/browse/CORS-1018 for tracking the issue which we should tackle in the next sprint.

lucab:
@steveeJ as you have a fair point that the test per-se is likely broken, I've piled up an `ignored` tag on it for the moment.

steveej:
Thanks @lucab. We'll need to go over all the ignored tests soon to ensure we don't regress without knowing it.

/lgtm

PR #71, opened 2019-03-13
policy-engine: minor appstate cleanup

lucab:
This marks the appstate as private and adds docstrings.


steveej:
If the tests are green this should be fine.

/lgtm

PR #72, opened 2019-03-14
graph-builder: instrument and expose graph metrics

lucab:
This adds a status service to serve graph-builder `/metrics` endpoint.
It is initially hardcoded to bind to `0.0.0.0:9080` (pending configuration rework).
It also adds initial instrumentation to record basic metrics about incoming
requests, total and failed scrapes, fetched releases, and nodes recorded in the
processed DAG.

Ref: https://jira.coreos.com/browse/CORS-1015
/cc @steveeJ

### DIFF
--- a/graph-builder/src/graph.rs
+++ b/graph-builder/src/graph.rs
@@ -17,13 +17,44 @@ use cincinnati::{plugins, AbstractRelease, Graph, Release, CONTENT_TYPE};
 use commons::GraphError;
 use config;
 use failure::Error;
+use prometheus::{Counter, IntGauge};
 use registry::{self, Registry};
 use serde_json;
 use std::collections::{HashMap, HashSet};
 use std::sync::{Arc, RwLock};
 use std::thread;
 
+lazy_static! {
+    static ref GRAPH_FINAL_RELEASES: IntGauge = register_int_gauge!(
+        "cincinnati_gb_graph_final_releases",
+        "Number of releases in the final graph, after processing"
+    )
+    .unwrap();
+    static ref GRAPH_UPSTREAM_RAW_RELEASES: IntGauge = register_int_gauge!(
+        "cincinnati_gb_graph_upstream_raw_releases",
+        "Number of releases fetched from upstream, before processing"
+    )
+    .unwrap();
+    static ref UPSTREAM_SCRAPES: Counter = register_counter!(
+        "cincinnati_gb_graph_upstream_scrapes_total",
+        "Total number of upstream scrapes"
+    )
+    .unwrap();
+    static ref UPSTREAM_ERRORS: Counter = register_counter!(
+        "cincinnati_gb_graph_upstream_errors_total",
+        "Total number of upstream scraping errors"
+    )
+    .unwrap();
+    static ref V1_GRAPH_INCOMING_REQS: Counter = register_counter!(
+        "cincinnati_gb_v1_graph_incoming_requests_total",
+        "Total number of incoming HTTP client request to /v1/graph"
+    )
+    .unwrap();
+}
+
 pub fn index(req: HttpRequest<State>) -> Result<HttpResponse, GraphError> {
+    V1_GRAPH_INCOMING_REQS.inc();
+
     // Check that the client can accept JSON media type.
     commons::ensure_content_type(req.headers(), CONTENT_TYPE)?;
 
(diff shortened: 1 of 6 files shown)


graph-builder/src/graph.rs
```diff
@@ -17,13 +17,34 @@ use cincinnati::{plugins, AbstractRelease, Graph, Release, CONTENT_TYPE};
...
 use registry::{self, Registry};
 use serde_json;
 use std::collections::{HashMap, HashSet};
 use std::sync::{Arc, RwLock};
 use std::thread;
 
+lazy_static! {
+    static ref GRAPH_DAG_NODES: IntGauge = register_int_gauge!(
+        "cincinnati_gb_graph_dag_nodes",
+        "Number of nodes (releases) in the graph."
```
steveej: Let's make sure this unmisunderstandably mentions "after processing".
Maybe we can rename it to `cincinnati_gb_graph_releases_processed` and explain it shortly in the description.

graph-builder/src/graph.rs
```diff
@@ -17,13 +17,34 @@ use cincinnati::{plugins, AbstractRelease, Graph, Release, CONTENT_TYPE};
...
+    )
+    .unwrap();
+    static ref GRAPH_FETCHED_RELEASES: IntGauge = register_int_gauge!(
+        "cincinnati_gb_graph_upstream_fetched_releases",
```
steveej: With the above rename this could be `cincinnati_gb_graph_releases_upstream` or maybe `_raw`

steveej (changes requested):
* Some suggestions around naming of the metrics.
* For the initial metrics we should also have an error counter for failed requests to the upstream.
* Optionally you could explore counting plugin errors, but I'm fine deferring that

graph-builder/src/graph.rs
lucab: I agree with that. I was trying to find a concise name to describe the "final processed DAG that is then cached as JSON-string" thing. Perhaps "processed DAG"? This would then become `graph_processed_dag_nodes` or `graph_processed_dag_releases`.

graph-builder/src/graph.rs
lucab: Current form matches the verb and help disambiguation: `fetch_releases()` -> `fetched_releases`. I can drop the "fetch" part if you prefer, but I think it makes more sense to rename the one above and keep this.

lucab:
Ack on the renaming part.
I'm leaving the error counting out of this as I'm waiting for some offline feedback on the suggested way to record positive/negative scrapes in an aggregable way (but I don't have a clear ETA right now).

lucab:
I got the feedback I was looking for (thanks @mxinden!), so I also added metrics for total and failed scrapes.

graph-builder/src/graph.rs
steveej: I understand that motivation, but I think it's unnecessarily close to the names in the code. I was trying to find high-level names which are meaningful without knowing the details of the codebase.

graph-builder/src/graph.rs
steveej: With the same motivation as explained below, trying to find high-level names: `graph_releases_final` and for the below `graph_releases_raw`?

lucab:
@steveeJ I've amended them to `upstream_raw_releases` and `final_releases`.

graph-builder/src/graph.rs
```diff
@@ -17,13 +17,44 @@ use cincinnati::{plugins, AbstractRelease, Graph, Release, CONTENT_TYPE};
...
+    )
+    .unwrap();
+    static ref UPSTREAM_SCRAPES: Counter = register_counter!(
+        "cincinnati_gb_graph_upstream_scrapes_total",
+        "Total number of upstream scrapes"
+    )
+    .unwrap();
+    static ref UPSTREAM_ERRORS: Counter = register_counter!(
+        "cincinnati_gb_graph_upstream_errors_total",
+        "Total number of upstream scrapes"
```
steveej: The description should mention errors ;-)

steveej (changes requested):
Just one mismatching description to fix before this is good to go :raised_hands:

lucab:
@steveeJ good catch! Amended.

steveej:
Thanks!

/lgtm

Unsolicited reminder: please don't forget to make this address and port configurable in #69.

mxinden (commented):
A *bit* late, but this looks good to me. Happy to see more components instrumented!

PR #74, opened 2019-03-19
[WIP] graph-builder: add dynamic plugins configuration

lucab:


cincinnati/src/plugins/mod.rs
```diff
@@ -4,7 +4,9 @@
 pub mod external;
 pub mod interface;
 pub mod internal;
+mod registry;
```
steveej, line 7: we lost a note on the way discussing this name. I suggest one of:
* catalog
* index

lucab: I'm +1 on catalog, will rename on next rebase.

steveej (commented):
I didn't manage to rebase this one properly, so I went ahead and re-created the changes and finally made the switch to toml config in #243, which has landed by now.

Thanks a lot for the effort!

/close

PR #75, opened 2019-03-19
graph-builder: use sha256 in payload if available

steveej:
This reduces the work for the consumer which now gets the most accurate
pullspec directly instead of having to assemble it using the manifestref
in the metadata.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1686589.

TODO: 
- [x] Extend tests

### DIFF
--- a/graph-builder/src/registry.rs
+++ b/graph-builder/src/registry.rs
@@ -249,6 +249,14 @@ pub fn fetch_releases(
         };
 
         if let Some(manifestref) = manifestref {
+            // Replace the tag specifier with the manifestref
+            release.source = {
+                let mut source_split: Vec<&str> = release.source.split(':').collect();
+                let _ = source_split.pop();
+
+                format!("{}@{}", source_split.join(":"), manifestref)
+            };
+
             // Attach the manifestref this release was found in for further processing
             release
                 .metadata
--- a/graph-builder/tests/net/quay_io/mod.rs
+++ b/graph-builder/tests/net/quay_io/mod.rs
@@ -12,10 +12,6 @@ use net::quay_io::cincinnati::plugins::internal::metadata_fetch_quay::DEFAULT_QU
 use net::quay_io::graph_builder::registry::Registry;
 use std::collections::HashMap;
 
-// TODO(steveeJ): fix and/or move this test to the plugin network tests
-//use net::quay_io::quay::v1::DEFAULT_API_BASE;
-//use net::quay_io::cincinnati::plugins::internal::metadata_fetch_quay::DEFAULT_QUAY_LABEL_FILTER;
-
 #[cfg(feature = "test-net-private")]
 use self::graph_builder::registry::read_credentials;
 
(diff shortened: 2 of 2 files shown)


steveej:
Putting this on hold explicitly in case the tests aren't actually broken, as this case isn't explicitly covered yet.

/hold

lucab:
My reading of the Bugzilla ticket is that this new behavior is intended to be the cornerstone for a security feature (supply chain integrity). As such, I think it was requested as the mandatory format for payloads, not an optional `if let`.

graph-builder/src/registry.rs
```diff
@@ -249,6 +249,14 @@ pub fn fetch_releases(
         };
 
         if let Some(manifestref) = manifestref {
+            // Replace the tag specifier with the manifestref
+            release.source = {
+                let mut source_split: Vec<&str> = release.source.split(':').collect();
+                let _ = source_split.pop();
```
lucab, line 255: digest-ref also [has a :](https://github.com/docker/distribution/blob/a25b9ef0c9fe242ac04bb20d3a028442b7d266b6/reference/reference.go#L17) within. I don't think there is any way for this `pop` to mistakenly operate on a hex-digest instead of a tag, but I wanted to mention that and make sure I didn't overlook something.

steveej:
> My reading of the Bugzilla ticket is that this new behavior is intended to be the cornerstone for a security feature (supply chain integrity). As such, I think it was requested as the mandatory format for payloads, not an optional `if let`.

We can't guarantee that since we're not getting the manifestref from all registries. Specifically when I ran the tests against Satellite that information was not available, and I don't want to completely break that.

I thought about introducing a flag to control the enforcement behavior and treat it as an error if we can't get the manifestref, but it seemed too invasive. We should get feedback from the reporter.

lucab (dismissed):
My understanding is that the request was for a mandatory digest in payload.

graph-builder/src/registry.rs
steveej: this assumes that the current form is of `image:tag`, which I know it is at this point. Do you suggest to ensure that there's no `@` in the source?

steveej:
> My understanding is that the request was for a mandatory digest in payload.

After getting the lack of feedback from the bug reporters I'm leaning towards making this configurable in form of a plugin which rewrites the payload according to the manifestref metadata. If the plugin is activated (which will be configurable after #74) it error if the required metadata can't be found.

I don't mind postponing this PR and get the configuration dependencies in first, or merge this one as is and refactor into a plugin later. @lucab thoughts?

graph-builder/src/registry.rs
lucab: I think it's a fair assumption even without additional checks, we just need to remember and be careful in the future not to break it (as it is not captured by type-system checks).

lucab:
I think this logic doesn't belong to a rewriting plugin, but to the registry-scraper itself (maybe controlled by an unsecure flag). I'm fine to merge this for now as it takes care of the most-likely case, and then adjust once we get clarification on how to handle the insecure corner-case. PR looks fine, feel free to cancel-hold when ok.

/hold
/lgtm

steveej:
Thanks for the review!

> I think this logic doesn't belong to a rewriting plugin, but to the registry-scraper itself (maybe controlled by an unsecure flag).

I agree, but let's just revisit this point when we get the feedback. It might be an opportunity to move the registry-scraper code into a plugin altogether.

/hold cancel

PR #76, opened 2019-03-25
commons: make options parsers string-generic

lucab:
This changes options parser to use string traits, allowing more
flexibility when using different string/collection/option types
and structopt.


lucab:
This is a small generic-enhancement split off #69, to make easier moving to `Option` and sharing the same type between structopt and serde.

/cc @steveeJ

PR #78, opened 2019-04-05
commons: list missing parameters in error message

lucab:
This tells the client which parameters are missing in a request.

Ref: https://jira.coreos.com/browse/CIN-10
/cc @steveeJ

### DIFF
--- a/commons/src/lib.rs
+++ b/commons/src/lib.rs
@@ -55,9 +55,11 @@ pub fn ensure_query_params(
         .map(|(k, _)| k)
         .collect();
 
-    // Make sure all mandatory parameters are present.
-    if !required_params.is_subset(&query_keys) {
-        return Err(GraphError::MissingParams);
+    // Make sure no mandatory parameters are missing.
+    let mut missing: Vec<String> = required_params.difference(&query_keys).cloned().collect();
+    if !missing.is_empty() {
+        missing.sort();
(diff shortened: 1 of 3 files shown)


commons/src/lib.rs
```diff
@@ -55,9 +55,11 @@ pub fn ensure_query_params(
         .map(|(k, _)| k)
         .collect();
 
-    // Make sure all mandatory parameters are present.
-    if !required_params.is_subset(&query_keys) {
-        return Err(GraphError::MissingParams);
+    // Make sure no mandatory parameters are missing.
+    let mut missing: Vec<String> = required_params.difference(&query_keys).cloned().collect();
+    if !missing.is_empty() {
+        missing.sort();
```
steveej, line 61: Just double-checking my understanding here. This sort is only for cosmetic reasons, right?

steveej (approved):
/lgtm

Thanks!

I left one question in a comment so just holding this until that's answered. Please cancel it after answering the question.

/hold

commons/src/lib.rs
lucab: This makes the output deterministic, allowing to check the error message for the real comma-separated values. It is not in a hot path, so I think it doesn't hurt. Alternatively I can drop this and make the test a little less strict.

PR #79, opened 2019-04-05
cincinnati,policy-engine: add channel filter plugin and activate it

steveej:
/cc @lucab 

TODO:
- [x] Test case for invalid channel
- [x] Discuss if we want to hide other channels if a matching release has multiple
- [x] Discuss if we want to land the configuration PRs first, as this PR currently hardcodes all the things
- [x] Discuss error exposure e.g. on invalid "channel"
- [x] Make sure we can handle the channel validation everywhere else (prohibits "" as a channel parameter)
    - [x] CVO (we can't do much about this on the CVO, as the installer sets the channel)
    - [x] live/readiness probes on the deployments #83

Fixes https://jira.coreos.com/browse/CIN-13.

### DIFF
--- a/cincinnati/src/lib.rs
+++ b/cincinnati/src/lib.rs
@@ -24,6 +24,9 @@ extern crate log;
 extern crate protobuf;
 extern crate try_from;
 extern crate url;
+#[macro_use]
+extern crate lazy_static;
+extern crate regex;
 
 pub mod plugins;
 
--- a/cincinnati/src/lib.rs
+++ b/cincinnati/src/lib.rs
@@ -160,6 +160,30 @@ impl Graph {
             .try_fold((), |_, (from, to)| self.add_edge(&from, &to))
     }
 
+    /// Returns tuples of ReleaseId and its version String for releases for which
+    /// filter_fn returns true.
+    ///
+    /// filter_fn is able to mutate the release as it receives a mutable borrow.
+    pub fn find_by_fn_mut<F>(&mut self, mut filter_fn: F) -> Vec<(ReleaseId, String)>
+    where
+        F: FnMut(&mut Release) -> bool,
+    {
+        self.dag
+            .node_weights_mut()
+            .enumerate()
+            .filter_map(|(i, nw)| {
+                if filter_fn(nw) {
+                    Some((
+                        ReleaseId(daggy::NodeIndex::from(i as u32)),
+                        nw.version().to_string(),
+                    ))
+                } else {
+                    None
+                }
+            })
+            .collect()
+    }
+
     /// Returns tuples of ReleaseId and its version String for releases which
     /// match the given metadata key/value pair.
     pub fn find_by_metadata_pair(&self, key: &str, value: &str) -> Vec<(ReleaseId, String)> {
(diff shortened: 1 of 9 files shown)


steveej:
This failure doesn't make sense :thinking: 

```
/bin/sh: line 2: dist/pr_check.sh: No such file or directory
```

Assuming it's a flake.

/retest

cincinnati/src/plugins/internal/channel_filter.rs
```diff
@@ -0,0 +1,329 @@
...
+static CHANNEL_VALIDATION_REGEX_STR: &str = r"[0-9a-zA-Z]+";
+
+lazy_static! {
+    static ref CHANNEL_VALIDATION_REGEX_RE: regex::Regex =
+        regex::Regex::new(&CHANNEL_VALIDATION_REGEX_STR).expect("could not create regex");
+}
+
+impl InternalPlugin for ChannelFilterPlugin {
+    fn run_internal(&self, internal_io: InternalIO) -> Fallible<InternalIO> {
+        let (channel, _) = match get_multiple_values!(internal_io.parameters, "channel", "channel")
```
steveej: FIXME

cincinnati/src/plugins/internal/channel_filter.rs
```diff
@@ -0,0 +1,329 @@
+//! This plugin can be used to filter a graph by a specific ! channel.
+//! It reads the requested channel from the parameters value at key ! "channel",
```
lucab: I think there are some spare `!` (or I didn't understand what they mean).

cincinnati/src/plugins/internal/channel_filter.rs
```diff
@@ -0,0 +1,329 @@
...
+
+use failure::Fallible;
+use plugins::{InternalIO, InternalPlugin};
+
+pub struct ChannelFilterPlugin {
+    pub key_prefix: String,
+    pub key_suffix: String,
+}
+
+static CHANNEL_VALIDATION_REGEX_STR: &str = r"[0-9a-zA-Z]+";
```
lucab: Do we have a specification for these or did you just come up with some reasonable values?

cincinnati/src/lib.rs
```diff
@@ -157,6 +160,35 @@ impl Graph {
...
 
+    /// Returns tuples of ReleaseId and its version String for releases for which
+    /// filter_fn returns true.
+    ///
+    /// filter_fn is able to mutate the release as it receives a mutable borrow.
+    /// It will only get passed a ConcreteRelease or AbstractRelease for each
+    /// release.
+    pub fn find_by_fn_mut<F>(&mut self, mut filter_fn: F) -> Vec<(ReleaseId, String)>
+    where
+        F: FnMut(Option<&mut ConcreteRelease>, Option<&mut AbstractRelease>) -> bool,
```
lucab: This is a bit of a strange signature. Unless you explicitly need to handle the None/None and Some/Some cases, it would be better to pass the actual enum.

steveej: It's indeed a bit strange. I wanted to prevent exposing the enum but if you'd prefer that I'm fine going that route.

cincinnati/src/plugins/internal/channel_filter.rs
steveej: I came up with these. I adapted them again to match the channels we currently envision.

cincinnati/src/lib.rs
steveej: I changed the signature to use the enum and it's actually not as terrible as I thought it would be.

lucab:
I just have a minor docstring comment.
However, my understanding is that as soon as this is deployed every non-channel-annotated release will disappear (i.e. most of the nodes in the current graph). Is that ok?

steveej:
> However, my understanding is that as soon as this is deployed every non-channel-annotated release will disappear (i.e. most of the nodes in the current graph). Is that ok?

The signal I received about this has been that we don't care about this circumstance, so I consider it OK.
I think that we do have to make sure that before we bump to prod we propose a doc change for the install process to set up channels, otherwise no cluster will be able to receive updates as per now the default is `""`.

steveej:
> I just have a minor docstring comment.

Addressed, thanks!

steveej:
[test]

lucab:
Code LGTM. Putting a temporary hold to have some extra room for figuring out RelEng integration.

/hold
/lgtm

steveej:
> Code LGTM. Putting a temporary hold to have some extra room for figuring out RelEng integration.

I think we're good here for the OpenShift deployment:
* The installer [has switched to the _stable-4.0_ channel][1] a while ago
* The 3 latest releases at [the default repo][2] are labelled with _stable-4.0_ and we just need to make sure that the subsequent ones will be too
* We accept that the older clusters won't be able to find their version in the received update graphs

/hold cancel

[1]: https://github.com/openshift/installer/commit/0ab9b19f1e10c3b3976f9b5d030ba64f7a94e54b 
[2]: https://quay.io/repository/openshift-release-dev/ocp-release?tag=latest&tab=tags

PR #80, opened 2019-04-08
graph-builder: expose liveness status

lucab:
This exposes a `/liveness` endpoint on the status service, to check
that the JSON graph content can be served.

Ref: https://jira.coreos.com/browse/CIN-16
/cc @steveeJ

PR #81, opened 2019-04-08
graph-builder: do not cache empty graph on failed scrapes

lucab:
This fixes a logic mistake in the scraper, where a temporary failure in
release fetching will result in caching an empty graph.

/cc @steveeJ


lucab:
Spot while playing with https://github.com/openshift/cincinnati/pull/80.

steveej (approved):
/lgtm

PR #84, opened 2019-04-11
openshift/policy-engine: require "channel" as a mandatory client parameter

steveej:
- [x] Depends on https://gitlab.cee.redhat.com/service/app-interface/merge_requests/365

/assign @lucab


lucab:
ConfigMap got already deployed on staging.

/lgtm

steveej:
> ConfigMap got already deployed on staging.

FYI: The ConfigMap in this repo doesn't have any effect (unfortunately). The referenced MR didn't trigger a re-deployment, for that we have to commit to master or manually restart the pods.

PR #89, opened 2019-04-17
cincinnati: namespace all metrics according to application

lucab:
This reworks metrics registration, introducing per-application namespaced registries.
In turn, this unlocks moving error counting into common library logic and introducing error-kind partitioning via labels.

/cc @steveeJ 
Ref: https://jira.coreos.com/browse/CIN-24

### DIFF
--- a/policy-engine/src/metrics.rs
+++ b/policy-engine/src/metrics.rs
@@ -5,13 +5,23 @@ use futures::future;
 use futures::prelude::*;
 use prometheus;
 
+/// Common prefix for policy-engine metrics.
+static PE_METRICS_PREFIX: &str = "cincinnati_pe";
+
+lazy_static! {
+    /// Metrics registry.
+    pub(crate) static ref PROM_REGISTRY: prometheus::Registry =
+        prometheus::Registry::new_custom(Some(PE_METRICS_PREFIX.to_string()), None)
+            .expect("could not create metrics registry");
+}
+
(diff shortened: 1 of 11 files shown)


lucab:
After this PR, error metrics looks like this (example from policy-engine):
```
# HELP cincinnati_pe_v1_graph_errors_total Errors on /v1/graph
# TYPE cincinnati_pe_v1_graph_errors_total counter
cincinnati_pe_v1_graph_errors_total{kind="invalid_content_type"} 1
```

lucab:
https://github.com/openshift/cincinnati/pull/97 landed with latest `prometheus`. Rebased and removed the WIP label, this is ready for review.

policy-engine/src/metrics.rs
```diff
@@ -5,13 +5,22 @@ use futures::future;
 use futures::prelude::*;
 use prometheus;
 
+/// Common prefix for policy-engine metrics.
+static PE_METRICS_PREFIX: &str = "cincinnati_pe";
+
+lazy_static! {
+    pub(crate) static ref PROM_REGISTRY: prometheus::Registry =
```
steveej, line 13: since this is `pub` we should probably add a docstring

steveej (commented):
looks like the upstream work has paid off, looks pretty neat!
I just have one minor comment for a missing docstring.

lucab:
Docstrings added.

PR #85, opened 2019-04-12
dist/template: fix YAML indentation

lucab:
This fixes several YAML sections that are over or under indented.
It contains no functional changes.

/cc @steveeJ

PR #90, opened 2019-04-29
Cargo/deps: bump dkregistry

steveej:
Recent changes in dkregistry implement content verification for image
layers.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -235,6 +235,25 @@ dependencies = [
 ]
 
 [[package]]
+name = "block-buffer"
+version = "0.7.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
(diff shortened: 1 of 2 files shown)


lucab:
@steveeJ my bad, I broke the CI when touching the Dockerfile. https://github.com/openshift/cincinnati/pull/91 should fix it (but I'm not sure how https://github.com/openshift/cincinnati/pull/85 and everything after it was green).

steveej:
> my bad, I broke the CI when touching the Dockerfile. #91 should fix it (but I'm not sure how #85 and everything after it was green).

CI uses what's currently in master. It's a trap I would like to avoid but haven't had any conversation about yet. We need someone to force-merge this.

steveej:
> We need someone to force-merge this.

What I meant is that we need someone to force merge the fix in #91 which has happened by now.

PR #93, opened 2019-04-30
dist/cincinnati: fix status port variable

lucab:
This fixes a mismatch on the template parameter name (as seen failing in staging).

/cc @steveeJ @aditya-konarde


steveej:
Let's get this one in as a quick fix and then clean up the configuration separately.

/lgtm

steveej (approved):

PR #95, opened 2019-05-03
graph-builder: eliminate possible rwlock starvation

steveej:


lucab:
I don't think that `RwLock::try_write` is what we are looking for. In particular, after this PR it'd be easier to hit the would-block case (with just a bunch of well-timed clients) and never update the cached graph.

I think we'd rather want some way to timeout a write-lock attempt after some reasonably long time (similar to https://docs.rs/lock_api/0.2.0/lock_api/trait.RawRwLockTimed.html) or to consider paying the price of a fair lock (similar to https://docs.rs/lock_api/0.2.0/lock_api/trait.RawRwLockFair.html).

steveej:
I started from scratch and substituted the fair `parking_lot::RWLock` for `std::sync::RwLock`. It's lock-poison-free which has consequences on the rest of the code, specifically the status methods. I haven't worked out all the details but I'd like to get an opinion on the current state.

@lucab thanks for the previous input, PTAL

steveej:
I took care of the remaining TODO by indicating a panic of the upstream scraper thread in the liveness status.
I also changed the readiness status to simply reflect the count of successful upstream scrapes.

graph-builder/src/graph.rs
```diff
@@ -170,6 +195,7 @@ pub fn run<'a>(settings: &'a config::AppSettings, state: &State) -> ! {
             }
             Err(err) => {
                 UPSTREAM_ERRORS.inc();
+                UPSTREAM_ERRORS_SUBSEQUENT.inc();
```
lucab: Prometheus can derive this from the rate of increase of `UPSTREAM_ERRORS`, no need to have a separate metrics for it (and it is not used for anything else in our own logic).

graph-builder/src/graph.rs
```diff
@@ -77,15 +79,29 @@ pub struct State {
...
+            live,
         }
     }
+
+    pub fn is_live(&self) -> bool {
+        *self.live.read()
+    }
+
+    pub fn is_ready(&self) -> bool {
+        (UPSTREAM_SCRAPES.get() - UPSTREAM_ERRORS.get()) > 0.0
```
lucab: This logic looks a bit strange to me.
`UPSTREAM_SCRAPES` can normally go up quite a bit (it's ~15k on current prod), so it will take a huge amount of runs for `UPSTREAM_ERRORS` to balance that and signal not-ready (if it ever manages to). Similarly, after a long strike of errors it will take a proportional strike of non-errors to reach ready again.

steveej: in case of an error both will increase. What this does is simply see if there was *ever* one successful scrape.

graph-builder/src/graph.rs
```diff
@@ -45,6 +46,11 @@ lazy_static! {
         "Total number of upstream scraping errors"
     )
     .unwrap();
+    static ref UPSTREAM_ERRORS_SUBSEQUENT: IntGauge = register_int_gauge!(
```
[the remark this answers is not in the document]
steveej: That word fits much better! I removed the metric altogether though as per your next comment. ;-) I was aware that this can be calculated in a range query but I thought it might be a nice instant metric to have.

steveej:
The devshift CI looks like a flake. It went through fine but somehow exited non-zero.

[test]

graph-builder/src/graph.rs
lucab: Ah, that's clever, I misread the logic!
Let's keep this then, just add a docstring or a comment to make that logic explicit.

graph-builder/src/status.rs
```diff
@@ -76,7 +39,7 @@ pub fn serve_metrics(
 ///  * Live (200 code): JSON graph is accessible (lock not poisoned).
```
lucab: This needs an update.

graph-builder/src/graph.rs
```diff
@@ -89,15 +86,29 @@ pub struct State {
...
+        live: Arc<RwLock<bool>>,
+    ) -> State {
         State {
             json,
             mandatory_params,
+            live,
         }
     }
+
+    pub fn is_live(&self) -> bool {
```
lucab, line 110: A docstring would be nice.

graph-builder/src/graph.rs
```diff
@@ -89,15 +86,29 @@ pub struct State {
...
+        *self.live.read()
+    }
+
+    pub fn is_ready(&self) -> bool {
```
lucab, line 115: A docstring would be nice.

graph-builder/src/graph.rs
```diff
@@ -214,7 +233,9 @@ pub fn run<'a>(settings: &'a config::AppSettings, state: &State) -> ! {
 
         match serde_json::to_string(&graph) {
             Ok(json) => {
-                *state.json.write().expect("json lock has been poisoned") = json;
+                *state.json.write() = json;
+                *state.live.write() = true;
```
lucab: I fear that we should signal "live" way before this.
Otherwise once we start having a large number of releases to scrape, we may enter in a restart-loop where the pod takes too much time to become live and is continuously killed and rescheduled.

steveej: I thought about that too, but we have https://github.com/openshift/cincinnati/blob/40e5956b0fbc5459bc0b762d6cd84a8942f1f75c/dist/openshift/cincinnati.yaml#L81 in place. I'm fine removing that delay and setting live earlier.

steveej: > I'm fine removing that delay and setting live earlier.

The delay is now 3 seconds and live is set before the scraping loop.

PR #96, opened 2019-05-06
graph-builder: test cyclic metadata

steveej:
This test is expected to fail with the according to the Cincinnati specification and implementation, which make use of a directed acyclic graph.

### DIFF
--- a/graph-builder/tests/images/test-cyclic-public-manual-0.0.2/release-metadata
+++ b/graph-builder/tests/images/test-cyclic-public-manual-0.0.2/release-metadata
@@ -1,6 +1,5 @@
 {
 	"kind": "cincinnati-metadata-v0",
-	"schema": 1,
 	"version": "0.0.2",
 	"previous": ["0.0.1"],
 	"next": ["0.0.1"],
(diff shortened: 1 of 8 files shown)


graph-builder/tests/net/quay_io/mod.rs
```diff
@@ -324,3 +328,36 @@ fn fetch_release_public_must_succeed_with_schemes_missing_http_https() {
     .map(|url| registry::Registry::try_from_str(url).unwrap())
     .for_each(test);
 }
+
+#[test]
+fn fetch_relesae_with_cyclic_metadata_fails() -> Fallible<()> {
```
lucab: typo: s/relesae/release/

graph-builder/tests/images/test-cyclic-public-manual-0.0.2/release-metadata
```diff
@@ -0,0 +1,10 @@
+{
+	"kind": "cincinnati-metadata-v0",
+	"schema": 1,
```
lucab: This field does not exist according to docs and codebase.

lucab (commented):
Just a couple of minor things, test logic is fine otherwise.

graph-builder/tests/net/quay_io/mod.rs
steveej: _relesae_ would be a nice plural form though ;-)

steveej:
@lucab all comments addressed

steveej:
> @steveeJ: you cannot LGTM your own PR.

my bad. @lucab mind putting it back? :-)

PR #92, opened 2019-04-30
policy-engine: add support for TOML configuration

lucab:
This adds TOML-based config file support for policy-engine, and unifies
CLI flags and TOML parameters names.

Ref: https://jira.coreos.com/browse/CIN-3
/cc @steveeJ

PR #98, opened 2019-05-08
commons: add docstrings and enable missing_docs lint

lucab:
This adds all missing docstrings and enable a lint to catch missing
ones in the future.

Ref: https://github.com/openshift/cincinnati/pull/92#discussion_r279713835
/cc @steveeJ


steveej:
thanks!

/lgtm

PR #101, opened 2019-05-10
commons: rename error metrics

lucab:
This renames the error counter in order to make more explicit that
it counts non-positive HTTP responses.

/cc @steveeJ @aditya-konarde


aditya-konarde:
Looks like I need to be added to reviewers? https://github.com/openshift/cincinnati/blob/master/OWNERS#L7

But LGTM

steveej:
Since https://gitlab.cee.redhat.com/service/app-interface/merge_requests/545 was already merged 3 hours ago we should get this in too. :pray:  to the CI gods

[test]
/lgtm

PR #102, opened 2019-05-14
*: Move from com.openshift.upgrades to io.openshift.upgrades

wking:
Generated with:

```console
$ sed -i 's/com\.openshift\.upgrades/io.openshift.upgrades/g' $(git grep -l 'com\.openshift\.upgrades')
```

There are currently no known consumers of this metadata, so move to the right namespace before we grow some.  The cluster-version operator is the only in-cluster graph consumer, and it isn't using anything under this namespace:

```console
cluster-version-operator$ git describe
v1.0.0-24-g33079c7
cluster-version-operator$ git grep 'com\.openshift\.upgrades'
...no hits...
```

CC @smarterclayton

### DIFF
--- a/cincinnati/src/plugins/internal/channel_filter.rs
+++ b/cincinnati/src/plugins/internal/channel_filter.rs
@@ -7,7 +7,7 @@ use plugins::{
     InternalIO, InternalPlugin, InternalPluginWrapper, Plugin, PluginIO, PluginSettings,
 };
 
-static DEFAULT_KEY_FILTER: &str = "com.openshift.upgrades.graph";
+static DEFAULT_KEY_FILTER: &str = "io.openshift.upgrades.graph";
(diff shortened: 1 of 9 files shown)


lucab (dismissed):
Tagged releases in the production repository are already actively using the `com.` prefix (both prefixes, actually):
```
io.openshift.release = 4.0.0-0.9
com.openshift.upgrades.graph.release.channels = stable-4.0 
```

steveej:
I changed the tags on https://quay.io/repository/redhat/openshift-cincinnati-test-labels-public-manual?tab=tags which is used for some tests to unblock this.

/retest
[test]

wking:
@smarterclayton just dropped dual-label into Quay to get us through this pivot on the feeding-into-Cincy side.

steveej (approved):
/lgtm

PR #103, opened 2019-05-15
docs/design/cincinnati: Remove edges instead of removing nodes

wking:
We need to keep the degraded release node in the graph so existing clusters can upgrade away.  Removing incoming edges is sufficient to keep new clusters off the degraded release.

Also add 1.1.1 to the Alpha graph, because we're not going to put the patch release straight into Beta without testing in Alpha first.

~The WIP is because it's not clear to me why we can't burn in new edges that aren't known (as previous or next edges) at release-publish time.  Maybe sometimes there would be upgrade implementation issues (e.g. 1.0.0 -> 1.2.0 upgrades crash the cluster), but I don't see any reason that Cincinnati couldn't handle these.~ [edit: Cincinnati [can handle these now][1].]

CC @crawford, @lucab, @smarterclayton, @steveeJ

[1]: https://github.com/openshift/cincinnati/pull/103#discussion_r284625343


wking:
Unclear to me at the moment is how to coax folks happily running a broken 1.1.0 with manual upgrades up to 1.2.0 (or over to 1.1.1, should we draw an edge for that?).  In passing today, @smarterclayton floated the [support operator][1] monitoring... some endpoint for new errata?  (And sets itself... Degraded?... when it sees a sufficiently frightening issue?).  But it's not clear to me why that couldn't come back in the Cincinnati graph response:

```json
{
  "nodes": [
    {
      "version": "1.1.0",
      "payload": "quay.io/openshift-release-dev/ocp-release@sha256:123...",
      "metadata": {
        "io.openshift.upgrades.graph.release.channels": "alpha",
        "io.openshift.upgrades.graph.release.manifestref": "sha256:123...",
        "io.openshift.upgrades.graph.release.warning": "Vulnerable to CVE-123!  Flee!"
      }
    },
    {
      "version": "1.1.1",
      "payload": "quay.io/openshift-release-dev/ocp-release@sha256:456...",
      "metadata": {
        "io.openshift.upgrades.graph.release.channels": "alpha",
        "io.openshift.upgrades.graph.release.manifestref": "sha256:456..."
      }
    }
  ],
  "edges": [
    [0, 1]
  ]
}
```

[1]: https://github.com/openshift/support-operator

docs/design/cincinnati.md
```diff
@@ -49,11 +49,20 @@ An example of this metadata document can be seen here:
...
 <figure align="center">
-  <img src="figures/cincinnati-1.svg" alt="Figure 1: A visualized DAG with three groups and a broken release" />
-  <figcaption>Figure 1: A visualized DAG with three groups and a broken release</figcaption>
+  <img src="figures/cincinnati-initial.svg" alt="Figure 1: A visualized DAG with four releases" />
+  <figcaption>Figure 2: A visualized DAG with four releases</figcaption>
+</figure>
+
+After releasing `1.1.0`, it is discovered that there is a critical bug and this version is subsequently flagged as degraded.
+Edges which lead into the degraded release are removed to prevent clients from upgrading to the degraded release.
+Because of this, it will not be possible for clients who are currently running `1.0.0` to update to `1.2.0` (FIXME: can we burn in a new edge from 1.0.0 direct to 1.2.0 by updating 1.2.0's metadata document to include a new `previous` entry for `1.0.0`?).
```
steveej: This is possible today by adding specific labels to quay tags from/to which you want to add/remove edges: https://github.com/openshift/cincinnati/blob/bf130b71146bf1f335de64f97e4ff9c1c481b2d3/cincinnati/src/plugins/internal/edge_add_remove.rs#L37-L39

steveej:
> The WIP is because it's not clear to me why we can't burn in new edges that aren't known (as previous or next edges) at release-publish time. Maybe sometimes there would be upgrade implementation issues (e.g. 1.0.0 -> 1.2.0 upgrades crash the cluster), but I don't see any reason that Cincinnati couldn't handle these.

Assuming by "burning" you mean creating new edges, despite the literal meaning being something destructive. Stating it positively I think that Cincinnati can handle this situation. It's just a different mutation of the graph.

I think we should make the changes which are required or better suited to the situations we're facing now, and there's no strict need to stick to original design docs.

wking:
> Assuming by "burning" you mean creating new edges...

Yup, https://en.wikipedia.org/wiki/Burn-in

docs/design/cincinnati.md
wking: > This is possible today by adding specific labels to quay tags...

This means we could remove `next` and just have folks manipulate `previous` to add/remove edges.  But I've left that alone for now, and just mentioned the late-edge-addition possibility in 974dc6e -> b9cea22.  This lets me remove the WIP tag from the PR too.

steveej: True that we don't strictly need both directions for adding/removing edges but it might come in handy and it wasn't significantly more effort, so let's keep the functionality in.

docs/design/openshift.md
```diff
@@ -82,8 +82,10 @@ Every update in the graph is annotated with its approved channels. In order for
   <figcaption>Figure 5: The view of the update graph from a client in the Alpha channel</figcaption>
 </figure>
 
-Above, in Figure 4, the Beta channel does not include 1.1.0 in the update graph because it hasn’t been approved. For example, this may be because a bug was discovered in 1.1.0 and fixed in 1.1.1 before rolling out to the Beta channel.
-
+Above, in Figure 4, the Beta channel does not include 1.1.0 in the update graph because it hasn’t been approved.
+For example, this may be because a bug was discovered in 1.1.0 and fixed in 1.1.1 before rolling out to the Beta channel.
+For release issues serious enough to warrent exclusion from the Beta channel, the Alpha-channel edge will also be removed to avoid funneling Alpha clusters to the degraded release.
+However, the degraded release will be remain in the Alpha channel to allow existing clusters to upgrade away from it.
```
steveej:
```suggestion
However, the degraded release will remain in the Alpha channel to allow existing clusters to upgrade away from it.
```

docs/design/cincinnati.md
```diff
@@ -49,14 +49,24 @@ An example of this metadata document can be seen here:
...
 <figure align="center">
-  <img src="figures/cincinnati-1.svg" alt="Figure 1: A visualized DAG with three groups and a broken release" />
-  <figcaption>Figure 1: A visualized DAG with three groups and a broken release</figcaption>
+  <img src="figures/cincinnati-initial.svg" alt="Figure 1: A visualized DAG with four releases" />
+  <figcaption>Figure 2: A visualized DAG with four releases</figcaption>
 </figure>
 
-In order to allow the transition mentioned above, a new version, `1.1.1`, needs to be released. In order to insert itself into the graph, it will need to make use of both the previous and next fields in its metadata:
+After releasing `1.1.0`, it is discovered that there is a critical bug and this version is subsequently flagged as degraded.
+Edges which lead into the degraded release are removed to prevent clients from upgrading to the degraded release.
```
steveej: I would try not to mix different levels of abstractions here. "Edge" is a graph term and "release" is an application term. I suggest to either replace "edge" by "update path" or "release" by "node".

docs/design/openshift.md
```diff
@@ -70,7 +70,7 @@ The Policy Engine represents these channels as a series of pointers to particula
...
 
-Every update in the graph is annotated with its approved channels. In order for the update to be included in the graph for a particular channel, it must be approved for that channel. This is done to prevent an update destined for a less stable channel from being served to a cluster subscribed to a more stable channel. The following helps demonstrate this:
+Every release in the graph is annotated with its approved channels. In order for the update to be included in the graph for a particular channel, it must be approved for that channel. This is done to prevent a release destined for a less stable channel from being served to a cluster subscribed to a more stable channel. The following helps demonstrate this:
```
steveej, line 73: Explicit :+1: here :smile:

steveej (changes requested):
Overall I agree with the general approach of removing edges instead of nodes :+1: 

Do you want to discuss how we represent the red color of a removed release in its metadata? In case you do, I'll bring one of the comment discussions we had out-of-band in here. I suggested to include a reason for the node removal, however this is slightly different with edge removals as it's not a boolean value which could be replaced by a reason.

Please have a look at my comments on the doc changes.

docs/design/openshift.md
wking: Fixed in b9cea22 -> 1221d80.

docs/design/cincinnati.md
wking: Fixed in b9cea22 -> 1221d80.

steveej (commented):
This looks good to me. Let's wait and merge it together with #105 in case we realized we missed something.

steveej (approved):

steveej (approved):
/lgtm
/hold

steveej:
The doc changes in here and #105 match up and the latter is about to be merged. This can go in!

Thanks @wking!

/hold cancel

PR #104, opened 2019-05-16
policy-engine && graph-builder: fix and switch to 2018 edition

shiywang:
https://jira.coreos.com/browse/CIN-32
https://jira.coreos.com/browse/CIN-33

What've been doing just basically follow this doc: https://doc.rust-lang.org/edition-guide/editions/transitioning-an-existing-project-to-a-new-edition.html

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 target/
 **/*.rs.bk
 **/*.swp
+.idea
--- a/cincinnati/src/plugins/mod.rs
+++ b/cincinnati/src/plugins/mod.rs
@@ -12,7 +12,7 @@ pub mod internal;
 pub use self::catalog::{deserialize_config, PluginSettings};
 use crate as cincinnati;
 use failure::{Error, Fallible, ResultExt};
-use plugins::interface::{PluginError, PluginExchange};
+use crate::plugins::interface::{PluginError, PluginExchange};
 use std::collections::HashMap;
 use try_from::{TryFrom, TryInto};
 
@@ -61,7 +61,7 @@ where
 
 /// Trait to be implemented by internal plugins with their native IO type
 pub trait InternalPlugin {
-    fn run_internal(&self, InternalIO) -> Fallible<InternalIO>;
+    fn run_internal(&self, input: InternalIO) -> Fallible<InternalIO>;
 }
 
 /// Trait to be implemented by external plugins with its native IO type
@@ -69,7 +69,7 @@ pub trait InternalPlugin {
 /// There's a gotcha in that this type can't be used to access the information
 /// directly as it's merely bytes.
 pub trait ExternalPlugin {
-    fn run_external(&self, ExternalIO) -> Fallible<ExternalIO>;
+    fn run_external(&self, input: ExternalIO) -> Fallible<ExternalIO>;
 }
 
 /// Dummy converter to satisfy the Trait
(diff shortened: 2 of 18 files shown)


quay/src/v1/mod.rs
```diff
@@ -1,5 +1,5 @@
 use failure::Fallible;
-use reqwest::{self, async};
+use reqwest::{self, r#async};
```
lucab: You can turn this into
```suggestion
use reqwest::{self, r#async as asynchro};
```
and then use `asynchro::` everywhere else in the rest of this file.

.gitignore
```diff
@@ -1,3 +1,5 @@
 target/
 **/*.rs.bk
 **/*.swp
+
```
lucab: Spurious newline.

quay/src/v1/mod.rs
steveej: Maybe better to just remove the async from the `use` here and change the code to `reqwest::async` where needed

cincinnati/src/plugins/mod.rs
```diff
@@ -61,15 +61,15 @@ where
 
 /// Trait to be implemented by internal plugins with their native IO type
 pub trait InternalPlugin {
-    fn run_internal(&self, InternalIO) -> Fallible<InternalIO>;
+    fn run_internal(&self, _: InternalIO) -> Fallible<InternalIO>;
```
steveej: To me the `_` is a bit confusing as in other places it means you can't use the value. Let's give it a name to avoid any confusion.

```suggestion
    fn run_internal(&self, input: InternalIO) -> Fallible<InternalIO>;
```

cincinnati/src/plugins/mod.rs
```diff
@@ -61,15 +61,15 @@ where
...
 }
 
 /// Trait to be implemented by external plugins with its native IO type
 ///
 /// There's a gotcha in that this type can't be used to access the information
 /// directly as it's merely bytes.
 pub trait ExternalPlugin {
-    fn run_external(&self, ExternalIO) -> Fallible<ExternalIO>;
+    fn run_external(&self, _: ExternalIO) -> Fallible<ExternalIO>;
```
steveej: see my previous comment.

```suggestion
    fn run_external(&self, input: ExternalIO) -> Fallible<ExternalIO>;
```

quay/src/v1/mod.rs
steveej:
```suggestion
use reqwest;
```

(note that this will break the other code)

steveej (changes requested):
Thanks for your first contribution here :raised_hands: I've never gone through an edition transition so we're learning together :wink:

quay/src/v1/mod.rs
lucab: `async` is a reserved keyword in edition-2018, which means that will end up with ` r#async` everywhere, thus why I suggested to rename-use as `asynchro` instead.

steveej: Are keywords also reserved even if they're not used as a whole word? I thought that `reqwest::async` shouldn't be an issue.

shiywang: it is as Luca said so I will change to `use reqwest::{self, r#async as asynchro};`

steveej: I can't confirm that. I tried my suggestion locally and it works.

Please see this diff: https://github.com/shiywang/cincinnati/compare/2018edition...steveeJ-forks:pr/upstream-104?expand=1

shiywang: okay, I think `cargo fix --edition` just replace `async` to `r#async` unconditionally, so rustc should be fine with this, I merged your pr, thanks!!

shiywang:
@steveeJ updated, PTAL, do I need to squash?

steveej:
> updated, PTAL, do I need to squash?

Yes please

steveej (approved):
/lgtm

Thanks!

PR #105, opened 2019-05-17
plugins/edge_add_remove: add special previous/next value for removing all edges in that direction

steveej:
This implements the spec changes in #103.

TODO:
- [x] Consent on concept #103
- [x] Blocked by #106
- [x] Docstrings
- [x] Plugin unit tests
- [x] Cincinnati library unit tests

/cc @lucab

### DIFF
--- a/cincinnati/src/lib.rs
+++ b/cincinnati/src/lib.rs
@@ -111,6 +111,30 @@ impl<'a> Iterator for NextReleases<'a> {
     }
 }
 
+/// Can be used to iterate over all direct parents of the given release.
+///
+/// See the `previous_releases` method for more information.
+pub struct PreviousReleases<'a> {
+    parents: daggy::Parents<Release, Empty, daggy::petgraph::graph::DefaultIx>,
+    dag: &'a Dag<Release, Empty>,
+}
+
+impl<'a> Iterator for PreviousReleases<'a> {
+    type Item = (daggy::EdgeIndex, daggy::NodeIndex, &'a Release);
+
+    fn next(&mut self) -> Option<Self::Item> {
+        self.parents
+            .walk_next(self.dag)
+            .map(|(edge_index, node_index)| {
+                (
+                    edge_index,
+                    node_index,
+                    self.dag.node_weight(node_index).expect(EXPECT_NODE_WEIGHT),
+                )
+            })
+    }
+}
+
 /// Dummy type used as edge-weights inside `Graph`.
 #[derive(Debug, Clone)]
 pub struct Empty;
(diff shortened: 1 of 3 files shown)


lucab:
This overall looks sane to me, but I didn't dig into the code yet.

cincinnati/src/plugins/internal/edge_add_remove.rs
```diff
@@ -184,13 +203,65 @@ mod tests {
...
+    fn ensure_previous_remove_all() {
+        let key_prefix = "test_prefix".to_string();
+        let key_suffix = "previous.remove".to_string();
+
+        let metadata: HashMap<usize, HashMap<String, String>> = [
+            (0, [].iter().cloned().collect()),
+            (1, [].iter().cloned().collect()),
+            (
+                2,
+                [(format!("{}.{}", key_prefix, key_suffix), String::from("*"))]
```
wking: should the explicit `"*"` be `remove_all_edges_value` here?

wking (commented):
I don't really understand what I'm looking at, but as far as I can tell this looks good to me ;).

steveej:
@lucab the initial work on this is complete and it's good to be reviewed. PTAL when you have time.

cincinnati/src/plugins/internal/edge_add_remove.rs
```diff
@@ -8,12 +8,16 @@ use crate::ReleaseId;
...
+pub static DEFAULT_REMOVE_ALL_EDGES_VALUE: &str = "*";
 
 #[derive(Clone, Debug, Deserialize, SmartDefault)]
 #[serde(default)]
 pub struct EdgeAddRemovePlugin {
     #[default(DEFAULT_KEY_FILTER.to_string())]
     pub key_prefix: String,
+
+    #[default(DEFAULT_REMOVE_ALL_EDGES_VALUE.to_string())]
+    pub remove_all_edges_value: String,
```
lucab, line 20: Does it make sense to have this configurable? It needs to be a special reserved identifier anyway, otherwise we cannot disambiguate it from a normal release version label.

cincinnati/src/lib.rs
```diff
@@ -662,6 +662,8 @@ mod tests {
 
     use super::*;
 
+    type Fallible<T> = Result<T, Box<std::error::Error>>;
```
lucab: I'd rather not squat `failure::Fallible` name, unless required. Can we either use that one directly or name it in our own way (e.g. `TestResult`)?

cincinnati/src/lib.rs
```diff
@@ -184,6 +208,25 @@ impl Graph {
...
+        match self.dag.remove_edge(index) {
+            Some(_) => Ok(()),
+            None => bail!("could not remove edge with index {:?}", index),
+        }
+    }
+
+    /// Remove the edges with the given indices.
+    ///
+    /// Stops and fails at the first edge which couldn't be removed.
+    pub fn remove_edges_by_index(&mut self, indices: Vec<daggy::EdgeIndex>) -> Result<(), Error> {
```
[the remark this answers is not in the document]
steveej: Did you forget an ampersand somewhere in your comment?

cincinnati/src/plugins/internal/edge_add_remove.rs
steveej: It makes sense to me to have this configurable as a way to expose it, since we currently don't have a proper way of communicating these decisions. Since we almost have the configuration infrastructure in place this poses no overhead.

cincinnati/src/lib.rs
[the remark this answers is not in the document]
steveej: Assuming you meant `Vec<&daggy::EdgeIndex>` I tried it out and eventually hit:

```
error[E0597]: `edge_index` does not live long enough
  --> cincinnati/src/plugins/internal/edge_add_remove.rs:74:56
   |
74 |                             .map(|(edge_index, _, _)| &edge_index)
   |                                                        ^^^^^^^^^-
   |                                                        |        |
   |                                                        |        `edge_index` dropped here while still borrowed
   |                                                        borrowed value does not live long enough
...
79 |                     }
   |                     - borrowed value needs to live until here
```

I'm not sure if it's worth resolving that.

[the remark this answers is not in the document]
lucab: Sorry, I typoed it, I meant `&[<daggy::EdgeIndex>]`.

cincinnati/src/plugins/internal/edge_add_remove.rs
lucab: Ack, let's keep it then. I'm mostly concerned with deployments configuring this in a silly (i.e. ambiguous) way.

PR #106, opened 2019-05-18
*: refactors, docstrings, formatting, clippy fixes

steveej:
These changes shouldn't alter any functionality and are split off of #105 to keep that focused on the topic.

### DIFF
--- a/cincinnati/src/plugins/internal/edge_add_remove.rs
+++ b/cincinnati/src/plugins/internal/edge_add_remove.rs
@@ -97,7 +97,7 @@ impl EdgeAddRemovePlugin {
         Ok(())
     }
 
-    /// Add next and previous releases specified by metadata
+    /// Add next and previous releases specified by metadata.
     ///
     /// The labels are assumed to have the syntax `<prefix>.(previous|next).add=(<Version>,)*<Version>`
     fn add_edges(&self, graph: &mut cincinnati::Graph) -> Fallible<()> {
--- a/cincinnati/src/plugins/internal/edge_add_remove.rs
+++ b/cincinnati/src/plugins/internal/edge_add_remove.rs
@@ -1,11 +1,11 @@
 //! This plugin adds and removes Edges from Nodes accordingly
 
 use crate as cincinnati;
-use failure::Fallible;
 use crate::plugins::{
     InternalIO, InternalPlugin, InternalPluginWrapper, Plugin, PluginIO, PluginSettings,
 };
 use crate::ReleaseId;
+use failure::Fallible;
 
 static DEFAULT_KEY_FILTER: &str = "io.openshift.upgrades.graph";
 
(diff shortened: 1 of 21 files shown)


graph-builder/src/registry.rs
```diff
@@ -280,13 +280,13 @@ pub fn fetch_releases(
 /// Update Images with release metadata should be immutable, but
 /// tags on registry can be mutated at any time. Thus, the cache
 /// is keyed on the hash of tag layers.
-fn cache_release(
+fn cache_release<S: ::std::hash::BuildHasher>(
```
lucab, line 283: This lint mostly makes sense when writing libraries with public API, where the consumer may need to pick the hasher. It isn't super-relevant in our case, I'd just allow it.

cincinnati/src/plugins/internal/edge_add_remove.rs
```diff
@@ -34,7 +34,7 @@ impl PluginSettings for EdgeAddRemovePlugin {
     }
 }
 
-/// Adds and removes next and previous releases to/from quay labels
+/// Adds and removes next and previous releases specified by metadata
```
lucab: Missing final dot.

cincinnati/src/plugins/internal/edge_add_remove.rs
```diff
@@ -50,7 +50,7 @@ impl EdgeAddRemovePlugin {
         Ok(Box::new(plugin))
     }
 
-    /// Remove next and previous releases from quay labels
+    /// Remove next and previous releases specified by metadata
```
lucab: Missing final dot.

cincinnati/src/plugins/internal/edge_add_remove.rs
```diff
@@ -97,7 +97,7 @@ impl EdgeAddRemovePlugin {
         Ok(())
     }
 
-    /// Add next and previous releases from quay labels
+    /// Add next and previous releases specified by metadata
```
lucab: Missing final dot.

lucab (commented):
Thanks a lot for the cleanup, looks very good to me. Just a bunch of very minor nits.

graph-builder/src/registry.rs
steveej: I understand. However, [the caller of this function](https://github.com/openshift/cincinnati/pull/106/files#diff-991c40183d69ebfda48cb830e94be824R194), which is part of a public API, passes down a variable which has this signature.

lucab: I'm fine keeping the new form if you like it (it's good practice anyway).

steveej: My point was that we have to decide for the caller and the callee as they have to match. So if we want this form in the public API we have to use it internally too.

steveej:
@lucab thanks for the review. I also noticed that I had an unintended left-over from the other PR in here which I removed now. Should be good for the final stamp now.

PR #108, opened 2019-05-24
hack/graph.sh: Add a script for rendering Cincinnati to DOT

wking:
Also has a drive-by fix for "Figure 2" -> "Figure 1" typo from 1221d80710 (#103).

### DIFF
--- a/hack/graph.sh
+++ b/hack/graph.sh
@@ -1,0 +1,34 @@
+#!/bin/sh
+#
+# Usage:
+#
+#   cat cincinnati.json | graph.sh [URI] >graph.dot
+#
+# For example:
+#
+#   curl -sH 'Accept:application/json' 'https://api.openshift.com/api/upgrades_info/v1/graph?channel=prerelease-4.1' | graph.sh https://cincinnati.example.org/your/graph | dot -Tsvg >graph.svg
+set -e
+
+JQ_SCRIPT="$(cat <<EOF
+  "digraph Upgrades {\n  labelloc=t;\n  rankdir=BT;" as \$header |
+  (
+    [
+      .nodes |
+      to_entries[] |
+      "  " + (.key | tostring) +
+             " [ label=\"" + .value.version + "\"" + (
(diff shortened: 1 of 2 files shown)


wking:
```console
$ hack/graph.sh | dot -Tpng >graph.png
```

graph

hack/graph.sh
```diff
@@ -0,0 +1,35 @@
...
+  (
+    [
+      .edges[] |
+      "  " + (.[0] | tostring) + "->" + (.[1] | tostring) + ";"
+    ] | join("\n")
+  ) as \$edges |
+  [\$header, \$nodes, \$edges, "}"] | join("\n")
+EOF
+)"
+curl --silent --header 'Accept:application/json' "${URI}" | jq -r "${JQ_SCRIPT}"
```
steveej: I would decrease the scope of this program and make it expect the content on stdin, leaving the fetching of the graph to the caller.

hack/graph.sh
```diff
@@ -0,0 +1,35 @@
...
+# Usage:
+#
+#   graph.sh [URI] >graph.dot
+#
+# For example:
+#
+#   graph.sh https://cincinnati.example.org/your/graph | dot -Tsvg >graph.svg
+set -e
+
+URI="${1:-https://api.openshift.com/api/upgrades_info/v1/graph?channel=prerelease-4.1}"
```
steveej: I was thinking of a way to make this URI a template with a nested variable for the channel as this is very guaranteed to change in the future. However that thought was superseded by the comment below, which solves the issue for this script altogether.

hack/graph.sh
wking: > I would decrease the scope of this program and make it expect the content on stdin...

I think it's a bit tedious to have to remind myself of "the usual" URI and set the `Accept` header.  But I don't feel *that* strongly about it ;).  Done with a3bb428 -> 7ec4253.

hack/graph.sh
wking: Obsolete after [this][1].

[1]: https://github.com/openshift/cincinnati/pull/108#discussion_r288158349

PR #110, opened 2019-05-28
rename file pr_check.sh to cargo_test.sh

shiywang:


shiywang:
@steveeJ @lucab this is another catch-22 PR, need to force merge I guess : D

steveej:
> @steveeJ @lucab this is another catch-22 PR, need to force merge I guess : D

You could add a symlink `pr_check.sh -> cargo_test.sh`  as an intermediate state so we don't have to force-merge.

steveej:
Merging this should unblock https://github.com/openshift/release/pull/3895. After merging the latter we can remove the symlink again.

steveej (approved):
I think these commits should be squashed and then it's good to go.

shiywang:
@steveeJ done

steveej (approved):
/lgtm

PR #112, opened 2019-05-28
hack/graph: Drop URI argument from the example

wking:
I'd forgotten to remove these while rerolling 7ec4253c3f (#108) to read the graph from stdin.


steveej:
Thanks, I missed that too in the review :see_no_evil: 

/lgtm

PR #113, opened 2019-05-30
remove pr_check.sh entierly and update README.md

shiywang:


shiywang:
Ops, do you know what is this `ci.ext.devshift.net` PR build? @steveeJ

steveej:
The script path for that build job is defined here:
* https://gitlab.cee.redhat.com/service/jjb/blob/af25a5e430435ed5dc6a65d58fc97e60fae38a74/ci-ext/cincinnati/40-jobs.yaml#L13I

How about the following suggestion.
We leave a `./dist/pr_check.sh` in place which runs both the other scripts? I would make the call to the yaml-lint script depend on the availability of the required binaries, and print a warning if it's not found.

steveej:
The PR job for `ci.ext.devshift.net` has been disabled. I'm not sure how to remove it from this PR. Maybe we need to delete a webhook.

steveej:
[test]

steveej:
I'm afraid we're stuck with this status check and I don't have permissions to check the settings. On newly created PRs we only have prow CI checks which is in a working state. Are you okay if I pick your commits into #114?

PR #116, opened 2019-06-06
add ghp-updater.sh script for generate doc pages to gh-pages

shiywang:
**1. about our index.html**
I take (project rust-random) https://github.com/rust-random/rand/blob/master/.travis.yml#L158-L161 and
https://github.com/huonw/travis-cargo/blob/master/travis_cargo.py#L145 
as a reference for generate our `index.html` 

`maybe we could use docstrings in the top-level crate for this?` since I do not quite understand the alternative way you suggested @steveeJ 

**2. about using travis or prow to push to gh-pages branch** 
if I added `ghp-updater.sh` to origin/release prow job, Does that ci-boot has permission to push to gh-pages by default ?

### DIFF
--- a/dist/ghp_updater.sh
+++ b/dist/ghp_updater.sh
@@ -1,0 +1,22 @@
+#!/usr/bin/env bash
+
+set -e
+
+ABSOLUTE_PATH="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+source "${ABSOLUTE_PATH}/commons.sh"
+DOC_OUTPUT_DIR="${PROJECT_PARENT_DIR}/docs"
+GH_REPO="shiywang/cincinnati"
+FULL_REPO="git@github.com:${GH_REPO}.git"
+
+
+if [[ -z "${CINCINNATI_GITHUB_PUSH_TOKEN_PATH}" ]]; then
+	echo 'No CINCINNATI_GITHUB_PUSH_TOKEN_PATH found, do not push to the remote repo'
+else
+	ssh-add ${CINCINNATI_GITHUB_PUSH_TOKEN_PATH}
+	cargo doc --no-deps --all --all-features --target-dir=${DOC_OUTPUT_DIR} --release
+	echo '<meta http-equiv=refresh content=0;url=doc/cincinnati/index.html>' > ${DOC_OUTPUT_DIR}/index.html
+	# commit and push changes
+	git add -A
+	git commit -m "GH-Pages update by openshift-merge-robot"
+	git push origin master
+fi
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,5 @@ target/
 **/*.rs.bk
 **/*.swp
 .idea
+
+docs/release
--- a/dist/commons.sh
+++ b/dist/commons.sh
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 
 IMAGE_BUILD="${IMAGE_BUILD:-local/cincinnati-build:latest}"
-PROJECT_PARENT_DIR="${ABSOLUTE_PATH:?need ABSOLUTE_PATH set}/../"
+PROJECT_PARENT_DIR="${ABSOLUTE_PATH:?need ABSOLUTE_PATH set}/.."
 GIT_REV="$(git rev-parse --short=7 HEAD)"
 BUILD_VOLUME_CARGO_GIT="build_cargo_git_${GIT_REV}"
 BUILD_VOLUME_CARGO_REGISTRY="build_cargo_registry_${GIT_REV}"


steveej (commented):
Are the other docs reachable by the cincinnati documentation which is the landing page now?

Also some bikesheddy suggestions.

shiywang:
@steveeJ comments addressed, the left question will be where do we get `GITHUB_TOKEN`?

if we using Travis it will be ideally we have an independent bot account which has write permission of project. 


if I added this script into prow origin/release, does it has the permission to do push to gh-pages by default ? cc @stevekuznetsov

stevekuznetsov:
What permissions are needed? Do you need to authenticate with GitHub to push?

steveej:
> if we using Travis it will be ideally we have an independent bot account which has write permission of project.

I think we can use that bot account on the existing CI too.
@stevekuznetsov are the GH credentials to pushing to this repo available in the regular test template or do we have to use a custom job again?

stevekuznetsov:
I think you should turn on the auto-update feature in GitHub for the pages instead of doing this. I messaged @shiywang about this.

steveej:
@stevekuznetsov 

> I think you should turn on the auto-update feature in GitHub for the pages instead of doing this. I messaged @shiywang about this.

Unless I'm missing something that's not going to work for us. We aren't going to use jekyll nor are we going to upload the docs in html format. We'd like the CI to render the code docstrings.

stevekuznetsov:
>  nor are we going to upload the docs in html format.

What are you serving from the website if not HTML?

steveej:
> What are you serving from the website if not HTML?

Sorry that was a bit misleading. We want to serve HTML but we don't want the developers to render and upload it. We rather want the CI to do that step on every merge to master.

stevekuznetsov:
I think checking it in and using GitHub's automation to update your pages would be an order of magnitude simpler than what we'd have to string together to get this to happen.

steveej:
> I think checking it in and using GitHub's automation to update your pages would be an order of magnitude simpler than what we'd have to string together to get this to happen.

I've done this before in Travis and it wasn't a huge deal at all. I acknowledge your warning, however I will try to integrate this with CI anyway. After all we already have a CI job which supports external secrets, so we might as well make use of it :-)

dist/ghp-updater.sh
```diff
@@ -0,0 +1,32 @@
+#!/usr/bin/env bash
+
+set -ex
+
+ABSOLUTE_PATH="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+source "${ABSOLUTE_PATH}/commons.sh"
+TARGET_DIR="${PROJECT_PARENT_DIR}target"
+GH_REPO="shiywang/cincinnati"
```
shiywang: will change to openshift, once postsubmit job is done

dist/ghp-updater.sh
```diff
@@ -0,0 +1,32 @@
+#!/usr/bin/env bash
+
+set -ex
+
+ABSOLUTE_PATH="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+source "${ABSOLUTE_PATH}/commons.sh"
+TARGET_DIR="${PROJECT_PARENT_DIR}target"
```
steveej: Is there a `/` missing or is that included in `PROJECT_PARENT_DIR`? If the latter is true we should probably remove it from there.

dist/ghp-updater.sh
```diff
@@ -0,0 +1,32 @@
+#!/usr/bin/env bash
+
+set -ex
```
steveej: This will expose the token in the logs due to the `set -x` earlier. We should probably remove the `-x` from it:

```suggestion
set -e
```

dist/ghp-updater.sh
shiywang: `+ RELEASE_OUTPUT_DIR=/home/shiywang/ART/cincinnati/dist/..//release-20190621.104710`

Yeah, I will remove it

dist/ghp_updater.sh
```diff
@@ -0,0 +1,32 @@
+#!/usr/bin/env bash
+
+set -e
+
+ABSOLUTE_PATH="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+source "${ABSOLUTE_PATH}/commons.sh"
+DOC_OUTPUT_DIR="${PROJECT_PARENT_DIR}/target/doc"
+GH_REPO="shiywang/cincinnati"
+FULL_REPO="https://${CINCINNATI_DOC_PUSH_TOKEN}@github.com/${GH_REPO}.git"
```
steveej:
```suggestion
FULL_REPO="https://${CINCINNATI_GITHUB_PUSH_TOKEN}@github.com/${GH_REPO}.git"
```

dist/ghp_updater.sh
```diff
@@ -0,0 +1,32 @@
...
+
+
+# clean up target/doc before generate docs
+rm -rf ${DOC_OUTPUT_DIR}
+mkdir ${DOC_OUTPUT_DIR}
+cd ${DOC_OUTPUT_DIR}
+
+if [[ -z "${CINCINNATI_DOC_PUSH_TOKEN}" ]]; then
```
steveej:
```suggestion
if [[ -z "${CINCINNATI_GITHUB_PUSH_TOKEN}" ]]; then
```

dist/ghp_updater.sh
```diff
@@ -0,0 +1,32 @@
...
+	echo 'No CINCINNATI_DOC_PUSH_TOKEN ENV found, do not push to the remote repo'
```
steveej:
```suggestion
	echo 'No CINCINNATI_GITHUB_PUSH_TOKEN ENV found, do not push to the remote repo'
```

dist/ghp_updater.sh
```diff
@@ -0,0 +1,32 @@
+#!/usr/bin/env bash
+
+set -e
+
+ABSOLUTE_PATH="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+source "${ABSOLUTE_PATH}/commons.sh"
+DOC_OUTPUT_DIR="${PROJECT_PARENT_DIR}/target/doc"
+GH_REPO="shiywang/cincinnati"
```
steveej, line 8: We should respect `CINCINNATI_GITHUB_PUSH_TOKEN_PATH` which could exist in CI and point to a file rather than containing the token directly.

```suggestion
GH_REPO="shiywang/cincinnati"
CINCINNATI_GITHUB_PUSH_TOKEN="${CINCINNATI_GITHUB_PUSH_TOKEN:-$(cat ${CINCINNATI_GITHUB_PUSH_TOKEN_PATH})"
```

dist/ghp_updater.sh
```diff
@@ -0,0 +1,32 @@
...
+
+if [[ -z "${CINCINNATI_DOC_PUSH_TOKEN}" ]]; then
+	echo 'No CINCINNATI_DOC_PUSH_TOKEN ENV found, do not push to the remote repo'
+else
+	git init
+	git remote add doc "${FULL_REPO}"
+	git config user.name "openshift-merge-robot"
+	git config user.email "openshift-merge-robot@redhat.com"
+	git fetch doc gh-pages
+    git checkout gh-pages
```
steveej: inconsistent line indentation here

dist/ghp_updater.sh
```diff
@@ -0,0 +1,32 @@
...
+    cargo doc --no-deps --all --all-features
```
steveej: inconsistent line indentation here

dist/ghp_updater.sh
```diff
@@ -0,0 +1,32 @@
...
+    echo '<meta http-equiv=refresh content=0;url=cincinnati/index.html>' > index.html
```
steveej: inconsistent line indentation here

dist/ghp_updater.sh
```diff
@@ -0,0 +1,32 @@
...
+	# commit and push changes
+    git add -A
```
steveej: inconsistent line indentation here

steveej (commented):
I adjusted the file to my review on https://github.com/openshift/release/pull/4138.
I also pointed out the inconsistent line indentations. I don't have a strong opinion on what indentation we use, just that it doesn't change within the same script ;-)

dist/ghp_updater.sh
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env bash
+
+set -e
+
+ABSOLUTE_PATH="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+source "${ABSOLUTE_PATH}/commons.sh"
+DOC_OUTPUT_DIR="${PROJECT_PARENT_DIR}/docs"
+GH_REPO="shiywang/cincinnati"
```
steveej, line 8:
```suggestion
GH_REPO="${GH_REPO:-openshift/cincinnati}"
```

dist/ghp_updater.sh
```diff
@@ -0,0 +1,22 @@
...
+FULL_REPO="git@github.com:${GH_REPO}.git"
+
+
+if [[ -z "${CINCINNATI_GITHUB_PUSH_TOKEN_PATH}" ]]; then
+	echo 'No CINCINNATI_GITHUB_PUSH_TOKEN_PATH found, do not push to the remote repo'
+else
+	ssh-add ${CINCINNATI_GITHUB_PUSH_TOKEN_PATH}
+	cargo doc --no-deps --all --all-features --target-dir=${DOC_OUTPUT_DIR} --release
+	echo '<meta http-equiv=refresh content=0;url=doc/cincinnati/index.html>' > ${DOC_OUTPUT_DIR}/index.html
```
steveej, line 17: I don't think we need to write this file every time. Can you please check this file in manually and remove this line?

dist/ghp_updater.sh
```diff
@@ -0,0 +1,22 @@
...
+	# commit and push changes
+	git add -A
+	git commit -m "GH-Pages update by openshift-merge-robot"
+	git push origin master
```
steveej, line 21: I would prefer if we used the `gh-pages` branch for the documentation like in the previous iterations of this script. Why did you change the approach?

shiywang: In postsubmit job you mentioned we are using master branch instead of gh-pages?

steveej: Ah, I think that's a misunderstanding. The postsubmit job has to run on merge to master, but the job itself should (in my opinion) push the doc updates to the gh-pages branch.

shiywang: ah.... I see, thank you for the clarification, I have very little time working on Cincinnati this week, I will update the PR this weekends if I got chance.

PR #117, opened 2019-06-07
policy-engine/graph: log served errors

steveej:
This might help debugging problematic requests until we have a better
way of tracing them.

### DIFF
--- a/policy-engine/src/graph.rs
+++ b/policy-engine/src/graph.rs
@@ -126,6 +126,15 @@ pub(crate) fn index(
         })
         .then(move |r| {
             timer.observe_duration();
+
+            if let Err(e) = &r {
+                error!(
+                    "Error serving request with parameters '{:?}': {}",
+                    req.query(),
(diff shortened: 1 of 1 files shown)


policy-engine/src/graph.rs
```diff
@@ -126,6 +126,15 @@ pub(crate) fn index(
         })
         .then(move |r| {
             timer.observe_duration();
+
+            if let Err(e) = &r {
+                error!(
+                    "Error serving request with parameters '{:?}': {}",
+                    req.query().to_owned(),
```
lucab: If possible, it would be nice not to perform this `to_owned` here.

lucab:
This is not our preferred solution, as it abuses/spams application logs as an error/access log.
However we don't have any better tool for this in the short term.
Longer term we are considering lifting up errors to a proper inspection facility, likely an on-premises https://sentry.io/.

/lgtm

PR #118, opened 2019-06-07
[WIP/RFC] dist: add script to test merge to master and build

steveej:
The use-case for this script is to hook it up into the app-sre CI to predict build failures before they happen on master.

TODO
- [ ] discuss if this is worth it despite the race between PRs merging to master

/cc @shiywang

### DIFF
--- a/dist/pr_merge_and_build.sh
+++ b/dist/pr_merge_and_build.sh
@@ -1,0 +1,27 @@
+#!/usr/bin/env bash
+#
+# This script tries to merge the current branch with master and then runs dist/build_deploy.sh
+#
+
+set -e
+
+ABSOLUTE_PATH="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+
+rev_to_merge="$(git rev-parse --abbrev-ref HEAD)"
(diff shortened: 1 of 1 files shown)


steveej:
I learned that on the app-sre CI [the merge happens for every PR before starting the pr_check](https://gitlab.cee.redhat.com/service/jjb/blob/511841f340ce2dd081b7520388ee9370e731f132/ci-ext/00-common.yaml#L153-155). Therefore this script is not required on CI. 

@shiywang @lucab would any of you find this useful for local usage? Otherwise I can close the PR unmerged.

lucab:
I'm used to manual pull+rebase, so I think you can close this. On a similar note, we can turn on enforcing PRs to be based on later master directly on GH, if you really want (I'm not a big fan of that for larger public projects, though).

steveej:
> I'm used to manual pull+rebase, so I think you can close this. On a similar note, we can turn on enforcing PRs to be based on later master directly on GH, if you really want (I'm not a big fan of that for larger public projects, though).

Thanks. Since the *ci.ext.devshift.net PR build* does merge to master before running the build I'm fine dropping this.

/close

PR #119, opened 2019-06-07
hack,dist: move pr_check.sh and update README.md

steveej:
/cc @shiywang

PR #120, opened 2019-06-11
update README.md remove file doesnt exist

shiywang:
`dist/build/Dockerfile` is no longer exist, so we might intent to use `dist/openshift-release/Dockerfile.builder` ?


steveej (commented):
AFAICT this is change is equivalent to https://github.com/openshift/cincinnati/pull/119/files#diff-04c6e90faac2675aa89e2176d2eec7d8R72. Any reason for having a separate PR for it?

shiywang:
cool, I think I can close mine

PR #123, opened 2019-06-17
*: bump actix/actix-web to 0.8.3/1.0.2

steveej:
TODO
- [x] migrate to new API
- [x] cleanup unwraps
- [x] fix tests
- [x] investigate why the response body from the policy-engine is empty in case of an error 
- [x] add a test case verifying the response body content


cincinnati/src/plugins/interface.rs
```diff
@@ -3,7 +3,7 @@
 
 // https://github.com/Manishearth/rust-clippy/issues/702
 #![allow(unknown_lints)]
-#![allow(clippy::all)]
+#![allow(clippy)]
```
steveej, line 6: FIXME

policy-engine/src/graph.rs
```diff
@@ -73,10 +71,18 @@ pub(crate) fn index(
...
     };
 
-    let plugin_params = req.query().to_owned();
+    let plugin_params = req
+        .query_string()
+        .to_owned()
+        .split("&")
+        .map(|pair| {
+            let kv_split: Vec<&str> = pair.split("=").collect();
+            (kv_split[0].to_string(), kv_split[1].to_string())
```
lucab: This will result in a client-controlled panic if `len < 2`, right?

steveej: Indeed! where `len < 2` means the query argument doesn't contain a `=`. In that case we should probably give a meaningful error to the client.

steveej: Since the processing of the query string isn't in the future chain it's not straight forward to generate an error for the user here. The current behavior in master is that such values are filled in with the empty string, so I'm opting for retaining that.

steveej: @lucab let me know if you agree, then I'll squash the fixup commit.

lucab: Ack. I agree it makes sense to either use an empty string, or filter-out valueless keys (assuming we'll never need a valueless key).

graph-builder/src/graph.rs
```diff
@@ -65,19 +65,26 @@ pub fn register_metrics(registry: &prometheus::Registry) -> Fallible<()> {
...
     V1_GRAPH_INCOMING_REQS.inc();
 
     // Check that the client can accept JSON media type.
     commons::ensure_content_type(req.headers(), CONTENT_TYPE)?;
 
     // Check for required client parameters.
-    let mandatory_params = &req.state().mandatory_params;
+    let mandatory_params = &req
+        .app_data::<State>()
+        .expect("the request has no app_data attached. this is a bug.")
```
lucab: Here and similar code elsewhere, I am bit on the fence on this. On one hand, this shouldn't really happen, so a panic is justified. On the other hand, it could be easier to just handle this as another error variant, so that we can easily watch via metrics and alert on it.

policy-engine/src/graph.rs
```diff
@@ -73,12 +74,38 @@ pub(crate) fn index(
...
+    // TODO(steveeJ): take another look at the actix-web docs for a method that
+    // provides this parameters split.
+    let plugin_params = req
+        .query_string()
+        .to_owned()
+        .split('&')
+        .map(|pair| {
+            let kv_split: Vec<&str> = pair.split('=').collect();
+
+            let value = if kv_split.len() < 2 {
```
lucab: Not sure if easier to read, but I think this one can be also written as `kv_split.get(1).unwrap_or_else(|| trace(); "")`. Assuming both are feasible, keep the one you prefer.

graph-builder/src/graph.rs
steveej: IMO this error shouldn't ever pass CI and I don't think it makes sense to track it.

policy-engine/src/graph.rs
steveej: Your suggestions is semantically equivalent in our use-case and avoids the if/else syntax :+1:

graph-builder/src/main.rs
```diff
@@ -61,27 +61,41 @@ fn main() -> Result<(), Error> {
...
+            )
+            .service(
+                actix_web::web::resource("/metrics")
+                    .route(actix_web::web::get().to(status::serve_metrics)),
+            )
+            .service(
+                actix_web::web::resource("/readiness")
+                    .route(actix_web::web::get().to(status::serve_readiness)),
+            )
+        // .middleware(Logger::default())
```
lucab: Is this a TODO or a leftover?

steveej: leftover, thanks!

steveej:
On local tests I noticed that the errors aren't reported as JSON strings to the client from the policy-engine. Instead the body is empty. I added some more tests to `the graph::index` method which seem fine. I suspect there has also been change in how the errors are converted into responses. I'll investigate that tomorrow.

graph-builder/src/graph.rs
steveej: I think the new test cases cover this one sufficiently.

steveej:
@lucab this one's ready for the next round. PTAL

graph-builder/src/graph.rs
lucab: Why isn't this using the constant from `commons`?

graph-builder/src/graph.rs
```diff
@@ -65,19 +65,26 @@ pub fn register_metrics(registry: &prometheus::Registry) -> Fallible<()> {
...
+        .mandatory_params;
     commons::ensure_query_params(mandatory_params, req.query_string())?;
 
-    let resp = HttpResponse::Ok()
-        .content_type(CONTENT_TYPE)
-        .body(req.state().json.read().clone());
+    let resp = HttpResponse::Ok().content_type(CONTENT_TYPE).body(
+        req.app_data::<State>()
+            .expect("the request has no app_data attached. this is a bug.")
```
lucab: Why isn't this using the constant from `commons`?

graph-builder/src/status.rs
```diff
@@ -39,9 +39,13 @@ pub fn serve_metrics(
 ///  * Live (200 code): The upstream scrape loop thread is running
 ///  * Not Live (500 code): everything else.
 pub fn serve_liveness(
-    req: HttpRequest<State>,
+    req: HttpRequest,
 ) -> Box<Future<Item = HttpResponse, Error = failure::Error>> {
-    let resp = if req.state().is_live() {
+    let resp = if req
+        .app_data::<State>()
+        .expect("the request has no app_data attached. this is a bug.")
```
lucab: Why isn't this using the constant from `commons`?

graph-builder/src/status.rs
```diff
@@ -55,9 +59,13 @@ pub fn serve_liveness(
 ///  * Ready (200 code): a JSON graph as the result of a successful scrape is available.
 ///  * Not Ready (500 code): no JSON graph available yet.
 pub fn serve_readiness(
-    req: HttpRequest<State>,
+    req: HttpRequest,
 ) -> Box<Future<Item = HttpResponse, Error = failure::Error>> {
-    let resp = if req.state().is_ready() {
+    let resp = if req
+        .app_data::<State>()
+        .expect("the request has no app_data attached. this is a bug.")
```
lucab: Why isn't this using the constant from `commons`?

policy-engine/src/graph.rs
```diff
@@ -189,4 +240,185 @@ mod tests {
...
+        let mandatory_params = vec!["channel".to_string()].into_iter().collect();
+        let state = AppState {
+            mandatory_params,
+            upstream: hyper::Uri::from_str(&mockito::server_url())?,
+            ..Default::default()
+        };
+
+        let http_req = actix_web::test::TestRequest::get()
+            .data(state)
+            .uri(&format!("{}?channel=':'", "http://doesn.really/matter"))
```
lucab: `doesnt`

But let's use something under `.test` TLD, it's reserved for this exact purpose.

policy-engine/src/graph.rs
```diff
@@ -189,4 +240,185 @@ mod tests {
...
+                    String::new()
+                } else {
+                    passed_params
+                        .iter()
+                        .fold(std::string::String::from("?"), |existing, current| {
+                            format!("{}{}={}", existing, current.0, current.1)
+                        })
+                }
+            );
+            println!("calculated service_uri: {}", &service_uri);
```
lucab: Is this ever shown when running tests?

policy-engine/src/graph.rs
```diff
@@ -189,4 +240,185 @@ mod tests {
...
+                if msg.contains("does not match regex") =>
+            {
+                Ok(())
+            }
+            res => Err(format!("expected FailedPluginExecution error, got: {:?}", res).into()),
+        }
+    }
+
+    #[test]
+    fn httpserver_must_respond_with_appropriate_json() -> Result<(), Box<Error>> {
```
lucab: I think this is only testing JSON error responses. Something like `webservice_json_error_reponses` would be a bit more descriptive here.

policy-engine/src/main.rs
```diff
@@ -50,10 +50,13 @@ fn main() -> Result<(), Error> {
 
     // Metrics service.
     graph::register_metrics(&metrics::PROM_REGISTRY)?;
-    server::new(|| {
+    HttpServer::new(|| {
         App::new()
-            .middleware(Logger::default())
-            .route("/metrics", Method::GET, metrics::serve)
+            // .middleware(Logger::default())
```
lucab: Leftover?

policy-engine/src/openapi.rs
```diff
@@ -23,7 +26,12 @@ pub(crate) fn index(req: HttpRequest<AppState>) -> HttpResponse {
 
     // Add mandatory parameters to the `graph` endpoint.
     if let Some(path) = spec_object.paths.get_mut("/v1/graph") {
-        add_mandatory_params(path, &req.state().mandatory_params);
+        add_mandatory_params(
+            path,
+            &req.app_data::<AppState>()
+                .expect("the request has no app_data attached. this is a bug.")
```
lucab: Why isn't this using the constant from `commons`?

lucab:
@steveeJ thanks for bumping it! I did another pass and only spot a few minor things, so I think this is mostly ready to land. There are some manifest+lockfile conflicts to adjust after the `prometheus` bump.

policy-engine/src/graph.rs
steveej: If a test fails you'll see it in the output. I used this for debugging the tests and it's not useful beyond that so it can be removed.

steveej:
@lucab Thanks for the review! All comments are addressed and the rebase is complete. The latter required me to adapt the tests you had introduced in #125, I hope I got them right. PTAL

PR #124, opened 2019-06-18
dist: add a Fedora-based container for fedora-infra

lucab:
This uses a Fedora environment for the container image. It is meant
to be consumed in fedora-infra.

/cc @steveeJ


dist/fedora-infra/Dockerfile
```diff
@@ -0,0 +1,26 @@
...
+
+# build: system utilities and libraries
+RUN dnf -y install g++ openssl-devel
+
+# build: pinned Rust stable toolchain (only use in case of temporary distro breakage)
+#ADD https://static.rust-lang.org/dist/rust-1.35.0-x86_64-unknown-linux-gnu.tar.gz rust.tar.gz
+#RUN tar -xf rust.tar.gz --strip 1 && ./install.sh
+
+# build: system Rust toolchain
+RUN dnf -y install rust cargo
```
steveej: you might want to pin this by version to make this Dockerfile reproducible

dist/fedora-infra/Dockerfile
```diff
@@ -0,0 +1,26 @@
...
+
+# build: graph-builder binary
+RUN cd /src && cargo build --release --bin graph-builder && mv /src/target/release/graph-builder /usr/local/bin/
```
steveej: instead of using `cd` you can use [`WORKDIR`](https://docs.docker.com/engine/reference/builder/#workdir).

dist/fedora-infra/Dockerfile
```diff
@@ -0,0 +1,26 @@
...
+
+# build: policy-engine binary
+RUN cd /src && cargo build --release --bin policy-engine && mv /src/target/release/policy-engine /usr/local/bin/
+
+# build: cleanup
+RUN cd /src && cargo clean
+
+# run: default config
+WORKDIR /
+CMD [ "/usr/local/bin/policy-engine" ]
```
steveej, line 21: I would suggest using a multi-stage build to separate the target binaries from the build sources and toolchain.

dist/fedora-infra/Dockerfile
lucab: This has (on purpose) some room to float. Version gating is done by whatever Fedora decides to promote to stable/F30. This allows fedora-infra to rebuild all the images with fixed packages whenever they need.

dist/fedora-infra/Dockerfile
lucab: Unfortunately the environment where this runs does not support multi-stage. I plan to investigate [S2I chained-builds](https://docs.openshift.com/container-platform/3.11/dev_guide/builds/advanced_build_operations.html#dev-guide-chaining-builds) later on, and switch to them if positive.

lucab:
Thanks for the `WORKDIR` suggestion, I wasn't aware of that usage. I've also left comments to explain the slightly unusual choices. PTAL.

dist/fedora-infra/Dockerfile
steveej: I see. That approach leaves the risk of re-using an old version until the cache is pruned. I'm leaning towards the idealistic side here as reproducibility issues are easy to miss when debugging. I don't want to block this PR on this discussion though as the scope is much larger.

dist/fedora-infra/Dockerfile
steveej: :man_shrugging:

dist/fedora-infra/Dockerfile
```diff
@@ -0,0 +1,28 @@
...
+# build: project sources
+ADD . /src
+WORKDIR /src
+
+# build: system utilities and libraries
+RUN dnf -y install g++ openssl-devel
+
+# build: pinned Rust stable toolchain (only use in case of temporary distro breakage)
+#ADD https://static.rust-lang.org/dist/rust-1.35.0-x86_64-unknown-linux-gnu.tar.gz rust.tar.gz
+#RUN tar -xf rust.tar.gz --strip 1 && ./install.sh
```
steveej: This leaves a mess on the filesystem which ends right up in the production image without separate bulid and deployment containers.

I would suggest to just remove the comments, maybe replaced by a link to https://github.com/openshift/cincinnati/blob/48fb7d57eda9948984af3a1c258217137ea98dc0/dist/Dockerfile.build/Dockerfile#L13-L18.

dist/fedora-infra/Dockerfile
lucab: Good point. I'm taking a note to double-check and bypass any caching at the build site.

dist/fedora-infra/Dockerfile
```diff
@@ -0,0 +1,28 @@
...
+RUN dnf -y install rust cargo
+
+# build: graph-builder binary
+RUN cargo build --release --bin graph-builder && mv /src/target/release/graph-builder /usr/local/bin/
+
+# build: policy-engine binary
+RUN cargo build --release --bin policy-engine && mv /src/target/release/policy-engine /usr/local/bin/
+
+# build: cleanup
+RUN cargo clean
```
steveej: If you want to clean up you could remove the source directory as a whole, as well as `$HOME/.cargo` and removing the toolchain.

steveej (commented):
I don't want to block you here but I'm going to leave a final warning that the Dockerfile has several impurity issues with regards to reproducibility and that the container size could be a lot smaller.

I'll leave further optimization at your discretion though :wink: 

/lgtm
/hold

cgwalters:
Honestly I'd push instead for at least deploying a 4.0 (or 3.11 if really needed) "builder environment" and pushing containers from there into prod rather than inflicting non-multi-stage builds on everyone.

dist/fedora-infra/Dockerfile
lucab: Ack, I dropped this altogether.

dist/fedora-infra/Dockerfile
lucab: Ack, I made this more aggressive as you suggested. I'm just leaving the RPMs around at this point.

lucab:
@steveeJ some more cleaning, but the green-label dropped. I acknowledge the fact that this is a less-than-stellar dockerfile.

PR #125, opened 2019-06-20
cargo: update prometheus to 0.7

lucab:
This updates prometheus library to latest version.

/cc @steveeJ


quay/Cargo.toml
```diff
@@ -18,4 +18,4 @@ tokio = "^0.1.14"
 
 [features]
 test-net = []
-test-net-private = []
```
steveej: this change seems unrelated

steveej (commented):
I'd feel more comfortable stamping this if we had metrics tests. Do you have time to work on that within the scope of this PR?

lucab:
@steveeJ I think so, but a complete metrics test would require an integration suite. Would a "the metrics handler returns 200 and a non-empty body" test be enough here (maybe with some keyword grepping too)?

steveej:
@lucab 

> Would a "the metrics handler returns 200 and a non-empty body" test be enough here

I'm fine with that for this PR :+1:

lucab:
@steveeJ this now comes with tests too, PTAL.

graph-builder/src/status.rs
```diff
@@ -64,3 +64,56 @@ pub fn serve_readiness(
     };
     Box::new(future::ok(resp))
 }
+
+#[cfg(test)]
+mod tests {
+    extern crate tokio;
```
steveej: AFAIK this is not needed anymore

graph-builder/src/status.rs
```diff
@@ -64,3 +64,56 @@ pub fn serve_readiness(
...
+
+    use self::tokio::runtime::current_thread::Runtime;
+    use crate::graph::State;
+    use crate::status;
+    use actix_web::test::TestRequest;
+    use parking_lot::RwLock;
+    use std::collections::HashSet;
+    use std::sync::Arc;
+
+    fn common_init() -> Runtime {
```
steveej: What do you think of using `Result` everywhere in the tests? This could then be 
```suggestion
    fn common_init() -> Result<Runtime, Box<Error>> {
```

graph-builder/src/status.rs
```diff
@@ -64,3 +64,56 @@ pub fn serve_readiness(
...
+        let _ = env_logger::try_init_from_env(env_logger::Env::default());
+        let test_gauge = prometheus::Gauge::new("test_gauge", "test help").unwrap();
+        test_gauge.set(42.0);
+        status::PROM_REGISTRY
+            .register(Box::new(test_gauge))
+            .unwrap();
```
steveej:
```suggestion
            ?;
```

graph-builder/src/status.rs
```diff
@@ -64,3 +64,56 @@ pub fn serve_readiness(
...
+        Runtime::new().unwrap()
```
steveej:
```suggestion
        Runtime::new()
```

graph-builder/src/status.rs
```diff
@@ -64,3 +64,56 @@ pub fn serve_readiness(
...
+        State::new(
+            json_graph.clone(),
+            HashSet::new(),
+            live.clone(),
+            ready.clone(),
+        )
+    }
+
+    #[test]
+    fn serve_metrics_basic() {
```
steveej:
```suggestion
    fn serve_metrics_basic() -> Result<(), Box<Error>> {
```

graph-builder/src/status.rs
```diff
@@ -64,3 +64,56 @@ pub fn serve_readiness(
...
+        let mut rt = common_init();
```
steveej:
```suggestion
        let mut rt = common_init()?;
```

graph-builder/src/status.rs
```diff
@@ -64,3 +64,56 @@ pub fn serve_readiness(
...
+
+        let http_req = TestRequest::with_state(mock_state()).finish();
+        let metrics_call = status::serve_metrics(http_req);
+        let resp = rt.block_on(metrics_call).unwrap();
```
steveej:
```suggestion
        let resp = rt.block_on(metrics_call)?;
```

graph-builder/src/status.rs
```diff
@@ -64,3 +64,56 @@ pub fn serve_readiness(
...
+
+        assert_eq!(resp.status().as_u16(), 200);
+        assert!(resp.body().is_binary());
+
+        if let actix_web::Body::Binary(body) = resp.body() {
+            assert!(!body.is_empty());
+            assert!(twoway::find_bytes(body.as_ref(), b"cincinnati_gb_test_gauge 42\n").is_some());
+        }
+    }
```
steveej, line 109:
```suggestion
        Ok(())
    }
```

steveej (commented):
I haven't fully gone through here but my general suggestion for all tests is that we use `Result` like in regular code.

lucab:
@steveeJ I've updated to use `Fallible` everywhere in those tests, does it look better now?

policy-engine/src/metrics.rs
```diff
@@ -31,3 +31,37 @@ pub(crate) fn serve(
...
 }
+
+#[cfg(test)]
+mod tests {
+    use crate::metrics;
+    use actix_web::test::TestRequest;
+    use failure::Fallible;
+    use tokio::runtime::current_thread::Runtime;
+
+    fn common_init() -> Fallible<Runtime> {
```
steveej: could `common_init` be shared between these? maybe in a `tests/lib.rs` file?

steveej (commented):
except for the duplicate `common_init`* method this looks pretty good, thanks for the tests!

\* *I went back and forth two twice and they looked the same*

policy-engine/src/metrics.rs
lucab: They are bit-by-bit identical. The problem is that they live in two separate top-level **binary crates**. The only way that comes to my mind is via a third crate library, but that sounds a bit gross.

steveej: I think it's fine having them in the *common* crate.

graph-builder/src/status.rs
```diff
@@ -64,3 +64,54 @@ pub fn serve_readiness(
...
+
+        let http_req = TestRequest::with_state(mock_state()).finish();
+        let metrics_call = status::serve_metrics(http_req);
+        let resp = rt.block_on(metrics_call)?;
+
+        assert_eq!(resp.status().as_u16(), 200);
+        assert!(resp.body().is_binary());
+
+        if let actix_web::Body::Binary(body) = resp.body() {
+            assert!(!body.is_empty());
```
steveej, line 105: Since you're using `failure` in here you can use the `ensure!()` macro here and elsewhere instead of `assert!()`

steveej (commented):
Since you're using `failure` in here you can use the `ensure!()` macro here and elsewhere instead of `assert!()`.

graph-builder/src/status.rs
lucab: `assert` and `ensure` are not directly interchangeable, as the latter doesn't have fileline information and additional introspection details provided by the former.

I'm sticking to `assert` for enforcing invariants, but feel free to have a spike ticket to investigate/build a blended `ensure`+`assert_*` macro.

lucab:
I've moved shared testing helpers to `commons::testing`, but kept the `assert` checks. PTAL.

steveej:
> I've moved shared testing helpers to `commons::testing`, but kept the `assert` checks. PTAL.

Thanks for walking the extra mile! :running_man: 

/lgtm

PR #127, opened 2019-06-26
cincinnati/plugins/edge_add_remove: change, document, and test label precedence

steveej:
/cc @shiywang @lucab 

FYI @wking

### DIFF
--- a/cincinnati/src/plugins/internal/edge_add_remove.rs
+++ b/cincinnati/src/plugins/internal/edge_add_remove.rs
@@ -23,8 +23,8 @@ pub struct EdgeAddRemovePlugin {
 impl InternalPlugin for EdgeAddRemovePlugin {
     fn run_internal(&self, io: InternalIO) -> Fallible<InternalIO> {
         let mut graph = io.graph;
-        self.remove_edges(&mut graph)?;
         self.add_edges(&mut graph)?;
+        self.remove_edges(&mut graph)?;
         Ok(InternalIO {
             graph,
             parameters: io.parameters,
@@ -41,6 +41,14 @@ impl PluginSettings for EdgeAddRemovePlugin {
 /// Adds and removes next and previous releases specified by metadata.
 ///
 /// The labels are assumed to have the syntax `<prefix>.(previous|next).(remove|add)=(<Version>,)*<Version>`
+///
+/// # Label processing order
+/// The labels are grouped and processed in two separate passes in the following order:
+///
+/// 1. *.add
+/// 2. *.remove
+///
+/// This ordering has implications on the result of semantical contradictions, so that the `*.remove` labels take precedence over `*.add`.
 impl EdgeAddRemovePlugin {
     /// Plugin name, for configuration.
     pub(crate) const PLUGIN_NAME: &'static str = "edge-add-remove";
@@ -165,6 +173,8 @@ mod tests {
     use super::*;
     use crate as cincinnati;
     use std::collections::HashMap;
+
+    static KEY_PREFIX: &str = "test_key";
 
     #[test]
     fn ensure_previous_remove() {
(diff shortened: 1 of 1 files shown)


cincinnati/src/plugins/internal/edge_add_remove.rs
```diff
@@ -423,4 +434,81 @@ mod tests {
...
+            (
+                1,
+                [
+                    (
+                        format!("{}.{}", key_prefix, "previous.remove"),
+                        String::from("0.0.0"),
+                    ),
+                    (
+                        format!("{}.{}", key_prefix, "next.add"),
+                        String::from("2.0.0"),
```
lucab: I may be misunderstanding something. Shouldn't this be supposed to cover (i.e. re-add) the same edge that the directive above removes?

lucab (commented):
Thanks, very valuable to have in an explicit form! I just have a doubt on what's the effective coverage of the new test.

cincinnati/src/plugins/internal/edge_add_remove.rs
steveej: This label is contradicted by the subsequent label at https://github.com/openshift/cincinnati/pull/127/files#diff-0b0028d0d2d6c8d8c41ae291b2dce0ebR474.

I could add more contradictions to an extent where every label has one. Would you prefer that?

lucab: I see.
My understanding was that this was only covering contradicting metadata processing **for each node**, while the case seems to relies on contradicting metadata across **two different nodes**. Shallow reading from my side, sorry.

This test is fine then, maybe an hint in the docstring would make the above obvious.

cincinnati/src/plugins/internal/edge_add_remove.rs
```diff
@@ -41,6 +41,17 @@ impl PluginSettings for EdgeAddRemovePlugin {
 /// Adds and removes next and previous releases specified by metadata.
 ///
 /// The labels are assumed to have the syntax `<prefix>.(previous|next).(remove|add)=(<Version>,)*<Version>`
+///
+/// # Label processing order
+/// The labels are processed in the following order:
```
lucab: Maybe `Labels are processed in multiple graph-wide passes, in the following order:`?

cincinnati/src/plugins/internal/edge_add_remove.rs
steveej: That is an interesting twist which I didn't intend to cover in this test. I will go back and think about incorporating that one too. I'll add a note to the docstring either way.

steveej:
@lucab hopefully all comments are addressed to both of our understand now. PTAL

cincinnati/src/plugins/internal/edge_add_remove.rs
```diff
@@ -41,6 +41,14 @@ impl PluginSettings for EdgeAddRemovePlugin {
 /// Adds and removes next and previous releases specified by metadata.
 ///
 /// The labels are assumed to have the syntax `<prefix>.(previous|next).(remove|add)=(<Version>,)*<Version>`
+///
+/// # Label processing order
+/// The labels are grouped and processed in two separate passes in the following:
```
lucab: "in the following order"

lucab:
Small typo, all fine otherwise. I'll leave this pending so that @wking has time to catch up and raise concerns (if any) on the semantics break.

/hold
/lgtm

lucab:
/lgtm

(Hold still in place on purpose)

steveej:
Thanks for the review @lucab. 

I'm fine merging this without further discussion, because the original ordering was a random choice at the time of implementing it, and the code change actually comes from a place of thought. I observed no significant\* difference in the result compared to the production instance so consider this change safe.

/hold cancel

\* the output is not bit-equal due to random re-ordering of JSON dictionary entries.

PR #129, opened 2019-06-28
policy-engine: add TOML configuration for policy plugins 

lucab:
This adds logic and basic tests to source policy plugins settings
from TOML configuration file.

Ref: https://jira.coreos.com/browse/CIN-3

### DIFF
--- a/cincinnati/src/plugins/mod.rs
+++ b/cincinnati/src/plugins/mod.rs
@@ -17,6 +17,18 @@ use std::collections::HashMap;
 use std::fmt::Debug;
 use try_from::{TryFrom, TryInto};
 
+/// Type for storing policy plugins in applications.
+pub type BoxedPlugin = Box<Plugin<PluginIO> + Sync + Send>;
+
+// NOTE(lucab): this abuses `Debug`, because `PartialEq` is not object-safe and
+// thus cannot be required on the underlying trait. It is a crude hack, but
+// only meant to be used by test assertions.
+impl PartialEq<BoxedPlugin> for BoxedPlugin {
+    fn eq(&self, other: &Self) -> bool {
+        format!("{:?}", self) == format!("{:?}", other)
+    }
+}
+
 /// Enum for the two IO variants used by InternalPlugin and ExternalPlugin respectively
 #[derive(Debug)]
 pub enum PluginIO {
--- a/commons/src/config.rs
+++ b/commons/src/config.rs
@@ -13,12 +13,12 @@ macro_rules! assign_if_some {
     }};
 }
 
-/// Merge configuration options into runtime settings.
+/// Try to merge configuration options into runtime settings.
 ///
-/// This consumes a generic configuration object, merging its options
+/// This consumes a generic configuration object, trying to merge its options
 /// into runtime settings. It only overlays populated values from config,
 /// leaving unset ones preserved as-is from existing settings.
 pub trait MergeOptions<T> {
     /// MergeOptions values from `options` into current settings.
-    fn merge(&mut self, options: T);
+    fn try_merge(&mut self, options: T) -> failure::Fallible<()>;
 }
--- a/policy-engine/src/config/file.rs
+++ b/policy-engine/src/config/file.rs
@@ -49,13 +49,14 @@ impl FileOptions {
 }
 
 impl MergeOptions<Option<FileOptions>> for AppSettings {
-    fn merge(&mut self, opts: Option<FileOptions>) {
+    fn try_merge(&mut self, opts: Option<FileOptions>) -> failure::Fallible<()> {
         if let Some(file) = opts {
             assign_if_some!(self.verbosity, file.verbosity);
-            self.merge(file.service);
-            self.merge(file.status);
-            self.merge(file.upstream);
+            self.try_merge(file.service)?;
+            self.try_merge(file.status)?;
+            self.try_merge(file.upstream)?;
         }
+        Ok(())
     }
 }
 
@@ -70,10 +71,11 @@ pub struct UpstreamOptions {
 }
 
 impl MergeOptions<Option<UpstreamOptions>> for AppSettings {
-    fn merge(&mut self, opts: Option<UpstreamOptions>) {
+    fn try_merge(&mut self, opts: Option<UpstreamOptions>) -> Fallible<()> {
         if let Some(upstream) = opts {
-            self.merge(upstream.cincinnati);
+            self.try_merge(upstream.cincinnati)?;
         }
+        Ok(())
     }
 }
 
@@ -99,7 +101,7 @@ mod tests {
         let toml_input = "status.port = 2222";
         let file_opts: FileOptions = toml::from_str(toml_input).unwrap();
 
-        settings.merge(Some(file_opts));
+        settings.try_merge(Some(file_opts)).unwrap();
         assert_eq!(settings.status_port, 2222);
     }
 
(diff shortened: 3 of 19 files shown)


policy-engine/src/config/file.rs
```diff
@@ -18,6 +18,9 @@ pub struct FileOptions {
     /// Upstream options.
     pub upstream: Option<UpstreamOptions>,
 
+    /// Policy plugins options.
+    pub policy: Option<Vec<toml::Value>>,
```
lucab, line 22: @steveeJ I was unsure whether to carve out `Vec<toml::Value>` to its own type (assuming serde allows me to make it transparent again when deserializing).

policy-engine/src/config/settings.rs
```diff
@@ -61,12 +67,28 @@ impl AppSettings {
...
+    /// Validate and return policy plugins.
+    pub fn policy_plugins(&self) -> Fallible<Vec<BoxedPlugin>> {
+        let mut configured_plugins = Vec::with_capacity(self.policies.len());
+        for conf in &self.policies {
+            let plugin = conf.build_plugin()?;
+            configured_plugins.push(plugin);
+        }
+
+        // TODO(lucab): drop this as soon as all config-maps are in place (prod & staging).
+        if configured_plugins.is_empty() {
```
lucab, line 84: This PR seemed big enough already, so I'm trying not to entangle it with prod+staging config bumping.
Thus, policy-engine still fallbacks to the hardcoded config for now, until we specificy (and populate) a TOML file via ConfigMap in both environments.
@steveeJ  let me know if you want to proceed otherwise.

policy-engine/tests/fixtures/policy-channel-filter.toml
```diff
@@ -0,0 +1,4 @@
+[[policy]]
+name = "channel-filter"
+key_prefix = "io.openshift.upgrades.graph"
+key_suffix = "release.channels"
```
lucab: Self note: missing newline at the end of file.

policy-engine/src/config/file.rs
steveej: I think it doesn't warrant for a new type.

policy-engine/src/config/settings.rs
steveej: I'm fine with that. I would be even more fine if we had an integration test which provides a TOML file and verifies the system as close production as possible.

steveej (commented):
Preliminary review pass with a few high-level comments.

cincinnati/src/plugins/mod.rs
```diff
@@ -17,6 +17,16 @@ use std::collections::HashMap;
 use std::fmt::Debug;
 use try_from::{TryFrom, TryInto};
 
+/// Type for storing policy plugins in applications.
+pub type BoxedPlugin = Box<Plugin<PluginIO> + Sync + Send>;
+
+impl PartialEq<BoxedPlugin> for BoxedPlugin {
+    fn eq(&self, other: &Self) -> bool {
+        // NOTE(lucab): this abuses `Debug`, as the trait object cannot require `PartialEq`.
+        format!("{:?}", self) == format!("{:?}", other)
```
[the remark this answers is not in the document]
lucab: This is used (so far exactly once) in tests: https://github.com/openshift/cincinnati/pull/129/files#diff-abed32e590a01cf6f55fd42829e13404R159.

This is indeed pretty ugly. However [trait object safety rules](https://doc.rust-lang.org/book/ch17-02-trait-objects.html#object-safety-is-required-for-trait-objects) disallow any use of generics (which `PartialEq` has), so we can't just require/use the implementation from underlying real type. Similar consideration for `Serialize`. Instead, `Debug` contains no generics, so it works here.

The only improvement I can think of here is to make this `#[cfg(test)]`, as I think we are not going to use it outside of tests.

policy-engine/src/config/settings.rs
lucab: Right now I'm hardcoding the default plugins in `run_test` here: https://github.com/openshift/cincinnati/pull/129/files#diff-ffbd2fce11fad2ac24f3974e0e763228R334.

Do you want me to use a TOML config in there instead?

cincinnati/src/plugins/mod.rs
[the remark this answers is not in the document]
steveej: > The only improvement I can think of here is to make this `#[cfg(test)]`, as I think we are not going to use it outside of tests.

That seems like a reasonable workaround. You could even pull the `impl` into the tests module.

policy-engine/src/config/settings.rs
steveej: > Do you want me to use a TOML config in there instead?

I'd prefer that so it would cover the read-from-file code too, right?

lucab: Sure, a TOML config read from a file. I was more asking about re-using that test. It seems like it'll be fine.

steveej: You lost me there, would you mind to elaborate?

cincinnati/src/plugins/mod.rs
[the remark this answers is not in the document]
lucab: Bad news: the consumer of this is in a separate crate, so neither of the two approaches above work (as the `cfg(test)` affects the top-level crate, not the dependencies).

policy-engine/src/config/settings.rs
lucab: Clarified out-of-band, we were stating the same thing in different ways.

lucab:
Amended, PTAL.

cincinnati/src/plugins/mod.rs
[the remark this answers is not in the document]
steveej: That's unfortunate. So we can only place a sentence in the docstring that it's only used for testing but we can't statically enforce that right now.

commons/src/config.rs
```diff
@@ -22,3 +22,14 @@ pub trait MergeOptions<T> {
...
     fn merge(&mut self, options: T);
 }
+/// Try to merge configuration options into runtime settings.
+///
+/// This consumes a generic configuration object, trying to merge its options
+/// into runtime settings. It only overlays populated values from config,
+/// leaving unset ones preserved as-is from existing settings.
+///
+/// It is a fallible version of `MergeOptions`.
+pub trait TryMergeOptions<T> {
```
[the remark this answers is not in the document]
lucab: Discussed out of band: let's unify and just keep a single trait (the fallible one).

policy-engine/src/main.rs
```diff
@@ -59,10 +61,12 @@ fn main() -> Result<(), Error> {
     .start();
 
     // Main service.
+    let plugins = Arc::new(settings.policy_plugins()?);
     let state = AppState {
         mandatory_params: settings.mandatory_client_parameters.clone(),
         upstream: settings.upstream.clone(),
         path_prefix: settings.path_prefix.clone(),
+        plugins: plugins,
```
steveej: We should verify what happens to the plugins if the AppState is cloned with regard to mutable, persistent fields in the plugins.

lucab: I've updated this logic to use the syntax recommend in https://doc.rust-lang.org/std/sync/struct.Arc.html#cloning-references.

That one makes it clearer that the code is creating a new reference, rather than duplicating the whole content.

lucab:
@steveeJ rebased to latest master and addressed all comments, PTAL.

steveej (approved):
This
/lgtm
But putting it on
/hold
to give you another opportunity to dig for a solution to the `PartiqlEq` workaround. Feel free to cancel the hold on your own accord.

lucab:
I looked a bit more into `cargo test` environment variables but didn't find anything I can directly use for conditional behavior. I'm merging as is.

/hold cancel

PR #130, opened 2019-07-01
cincinnati/plugins: add "assign_wariness" plugin

lucab:
This implements a new "assign_wariness" plugin, which is mainly meant
to be used on policy-engine side.

This plugin ensures that each request has an attached "wariness" score
for throttling purposes, which can be either provided by the client
or computed server-side based on some other client input paramenters.

Wariness value can be then used by other plugins to enact (static or dynamic)
throttling decisions. It has the property that a set of request parameters
will always result in the same wariness value, for stable/deterministic
bucketing.

/cc @steveeJ

### DIFF
--- a/cincinnati/src/plugins/catalog.rs
+++ b/cincinnati/src/plugins/catalog.rs
@@ -3,6 +3,7 @@
 //! This catalog relies on a static list of all available plugins,
 //! referenced by name. It is used for configuration purposes.
 
+use super::internal::assign_wariness::AssignWariness;
 use super::internal::channel_filter::ChannelFilterPlugin;
 use super::internal::cincinnati_graph_fetch::CincinnatiGraphFetchPlugin;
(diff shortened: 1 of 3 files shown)


lucab:
/hold

This is temporarily based on https://github.com/openshift/cincinnati/pull/129, as it uses `BoxedPlugin` from there.

lucab:
Out of band discussion: openshift systems uses floats from 0.0 to 1.0, it would be good to make the two compatible.

lucab:
@steveeJ I've updated this PR to use floats within the range `(0.0, 1.0]` as throttling scores, as discussed.

The overall algorithm that will consume this value is sketched at https://github.com/coreos/zincati/issues/82 and aligns with the current draft of OpenShift "risk appetite" logic.

vrutkovs:
/hold cancel

Testing permissions

squeed:
/hold
because @lucab asked me to

PR #131, opened 2019-07-01
cincinnati/plugins/edge_add_remove: graceful and consistent handling of all inconsistent label constellations

steveej:
Introduces consistency in gracefully handle all expected inconsistent label constellations for edge addition and removal.

Please see the commit messages for more details.

### DIFF
--- a/cincinnati/src/lib.rs
+++ b/cincinnati/src/lib.rs
@@ -35,7 +35,7 @@ pub mod plugins;
 
 use daggy::petgraph::visit::{IntoNodeReferences, NodeRef};
 use daggy::{Dag, Walker};
-use failure::Error;
+use failure::{Error, Fallible};
 use serde::de::{self, Deserialize, Deserializer, MapAccess, Visitor};
 use serde::ser::{Serialize, SerializeStruct, Serializer};
 use std::collections::HashMap;
@@ -193,6 +193,13 @@ impl Graph {
             .node_references()
             .find(|nr| nr.weight().version() == version)
             .map(|nr| ReleaseId(nr.id()))
+    }
+
+    /// Returns a Release for the given &ReleaseId
+    pub fn find_by_releaseid(&self, id: &ReleaseId) -> Fallible<&Release> {
+        self.dag
+            .node_weight(id.0)
+            .ok_or_else(move || format_err!("could not find Release with id: {:?}", id))
     }
 
     /// Removes the directed edge between the given releases.
--- a/cincinnati/src/lib.rs
+++ b/cincinnati/src/lib.rs
@@ -145,6 +145,25 @@ impl<'a> Iterator for PreviousReleases<'a> {
 #[derive(Debug, Clone)]
 pub struct Empty;
 
+/// Errors that can be returned by the methods in this library
+pub mod errors {
+    /// Edge already exists
+    #[derive(Debug, Fail, Eq, PartialEq)]
+    #[fail(display = "edge from {:?} to {:?} already exists", from, to)]
+    pub struct EdgeAlreadyExists {
+        pub(crate) from: String,
+        pub(crate) to: String,
+    }
+
+    /// Edge doesn't exist
+    #[derive(Debug, Fail, Eq, PartialEq)]
+    #[fail(display = "edge from '{:?}' to '{:?}' doesn't exist", from, to)]
+    pub struct EdgeDoesntExist {
+        pub(crate) from: String,
+        pub(crate) to: String,
+    }
+}
+
 impl Graph {
     /// Add a release to the graph.
     ///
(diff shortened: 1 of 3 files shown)


cincinnati/src/lib.rs
```diff
@@ -145,6 +146,17 @@ impl<'a> Iterator for PreviousReleases<'a> {
 #[derive(Debug, Clone)]
 pub struct Empty;
 
+/// Errors that can be returned by the methods in this library
+pub mod errors {
+    /// Edge already exists
+    #[derive(Debug, Fail, Eq, PartialEq)]
+    #[fail(display = "edge from {:?} to {:?} already exists", from, to)]
```
lucab, line 152: I'm not sure this is very helpful. Node indexes depend on the specific graph instance/ordering. Should we maybe track source and destination versions instead?

cincinnati/src/plugins/internal/edge_add_remove.rs
```diff
@@ -128,18 +132,30 @@ impl EdgeAddRemovePlugin {
...
+
         graph
             .find_by_metadata_key(&format!("{}.{}", self.key_prefix, "previous.add"))
             .into_iter()
-            .try_for_each(|(to, to_version, from_csv)| -> Fallible<()> {
-                for from_version in from_csv.split(',').map(str::trim) {
-                    if let Some(from) = graph.find_by_version(&from_version) {
-                        info!("[{}]: adding previous {}", &from_version, &to_version);
-                        graph.add_edge(&from, &to)?
+            .try_for_each(|(to, to_string, from_csv)| -> Fallible<()> {
```
lucab, line 166: It looks like this `try_for_each` was drifting rightward and you had to introduce a macro. Did you maybe consider splitting this whole closure to its own function?

commons/src/testing.rs
```diff
@@ -3,9 +3,15 @@
 use failure::Fallible;
 use tokio::runtime::current_thread::Runtime;
 
+/// Initialize logging.
+pub fn init_logger() -> Fallible<()> {
+    env_logger::try_init_from_env(env_logger::Env::default())?;
```
lucab, line 8: The original location where you moved this from didn't have the error unwrapping. I think the `?` in here will make this failing as soon as there two tests using it, as the logger for subsequent tests is already initialized by the first running test.

steveej: That's right, I didn't want to swallow the error here. The caller can if it wants to. Do you prefer an opinionated abstraction here?

cincinnati/src/plugins/internal/edge_add_remove.rs
steveej: I tried to use the macro to reduce the duplication between the closures for the previous and next. A function wouldn't help with that use-case, and I wasn't sure how much effort I want to put into making one macro for both cases.

cincinnati/src/lib.rs
steveej: I changed it to be a String here so it can be fed a version string at instantiation.

steveej:
@lucab I increased the scope of this PR a bit as I kept hitting more error cases in the tests and wanted them handled properly. Please take a look at the new tests and the commit messages for a complete understanding.

commons/src/testing.rs
lucab: If that was on purpose and we plan to stick with swallowing error at caller site, I'm fine with keeping this.

cincinnati/src/plugins/internal/edge_add_remove.rs
lucab: I guess we can leave this as-is for now then (I think that my minor itches here are the two macro-in-function).

If/when we need to have the same permissive add/remove logic in some other place, we can reconsider how to better share this logic.

cincinnati/src/plugins/internal/edge_add_remove.rs
```diff
@@ -46,9 +46,17 @@ impl PluginSettings for EdgeAddRemovePlugin {
...
+///     1. previous
+///     2. next
 /// 2. *.remove
+///     1. previous
+///     2. next
 ///
 /// This ordering has implications on the result of semantical contradictions, so that the `*.remove` labels take precedence over `*.add`.
+///
+/// # Strictness
+/// The plugin aims to gracefully handle any inconstistencies to make the operation as robust as possible.
```
lucab: Typo: `inconsistencies`

cincinnati/src/lib.rs
```diff
@@ -145,6 +146,25 @@ impl<'a> Iterator for PreviousReleases<'a> {
...
+    pub struct EdgeAlreadyExists {
+        pub(crate) from: String,
+        pub(crate) to: String,
+    }
+
+    /// Edge already exists
```
lucab: I think this is a copy-paste mismatch.

lucab:
@steveeJ thanks, I think it makes sense to have them all here as they fall under the same logical umbrella. I made another pass, and I only spotted minor typos.

steveej:
@lucab Thanks for the review. I addressed your comments and also removed two commits which added unused code and the lack of a good test use-case. Please take another look and stamp if you think it's good.

lucab (approved):

PR #136, opened 2019-07-26
graph-builder/metrics: track timestamp of last graph refresh

lucab:
This instruments and tracks via metrics the timestamp of last
successful graph refresh.

Ref: https://jira.coreos.com/browse/CIN-60
/cc @steveeJ

### DIFF
--- a/graph-builder/src/graph.rs
+++ b/graph-builder/src/graph.rs
@@ -31,6 +31,11 @@ lazy_static! {
         "Number of releases in the final graph, after processing"
     )
     .unwrap();
+    static ref GRAPH_LAST_SUCCESSFUL_REFRESH: IntGauge = IntGauge::new(
+        "graph_last_successful_refresh_timestamp",
+        "UTC timestamp of last successful graph refresh"
+    )
+    .unwrap();
     static ref GRAPH_UPSTREAM_RAW_RELEASES: IntGauge = IntGauge::new(
         "graph_upstream_raw_releases",
         "Number of releases fetched from upstream, before processing"
@@ -57,6 +62,7 @@ lazy_static! {
 pub fn register_metrics(registry: &prometheus::Registry) -> Fallible<()> {
     commons::register_metrics(&registry)?;
     registry.register(Box::new(GRAPH_FINAL_RELEASES.clone()))?;
+    registry.register(Box::new(GRAPH_LAST_SUCCESSFUL_REFRESH.clone()))?;
     registry.register(Box::new(GRAPH_UPSTREAM_RAW_RELEASES.clone()))?;
     registry.register(Box::new(UPSTREAM_ERRORS.clone()))?;
     registry.register(Box::new(UPSTREAM_SCRAPES.clone()))?;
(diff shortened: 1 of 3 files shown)


graph-builder/src/graph.rs
```diff
@@ -249,21 +255,27 @@ pub fn run<'a>(settings: &'a config::AppSettings, state: &State) -> ! {
...
-                GRAPH_FINAL_RELEASES.set(nodes_count as i64);
-                debug!("graph update completed, {} valid releases", nodes_count);
-            }
-            Err(err) => error!("Failed to serialize graph: {}", err),
+        if first_success {
+            *state.ready.write() = true;
+            first_success = false;
         };
+
+        let last_refresh = chrono::Utc::now();
```
steveej: I think you could use this directly in the next statement

graph-builder/src/graph.rs
```diff
@@ -249,21 +255,27 @@ pub fn run<'a>(settings: &'a config::AppSettings, state: &State) -> ! {
...
+        GRAPH_LAST_SUCCESSFUL_REFRESH.set(last_refresh.timestamp() as i64);
+
+        let nodes_count = graph.releases_count();
```
steveej, line 275: I think you could use this directly in the next statement

graph-builder/src/graph.rs
```diff
@@ -31,6 +31,11 @@ lazy_static! {
         "Number of releases in the final graph, after processing"
     )
     .unwrap();
+    static ref GRAPH_LAST_SUCCESSFUL_REFRESH: IntGauge = IntGauge::new(
+        "graph_last_successful_refresh_timestamp",
```
steveej, line 35: maybe `graph_upstream_last_successful_refresh_timestamp` for consistency?

graph-builder/src/graph.rs
```diff
@@ -31,6 +31,11 @@ lazy_static! {
         "Number of releases in the final graph, after processing"
     )
     .unwrap();
+    static ref GRAPH_LAST_SUCCESSFUL_REFRESH: IntGauge = IntGauge::new(
```
steveej, line 34: maybe `GRAPH_UPSTREAM_LAST_SUCCESSFUL_REFRESH_TIMESTAMP` for consistency?

steveej (commented):
LGTM in general. Just some naming nitpicking and possible code refactor comments.

graph-builder/src/graph.rs
lucab: I'd say no, at this point this metrics is not anymore related to reaching upstream (e.g. a plugin could be failing, without any quay-fetching issues) and it is already consistent with the other `graph_final_releases` one.

graph-builder/src/graph.rs
lucab: Similar no, see below.

graph-builder/src/graph.rs
lucab: No, this is also used by the `debug!` one line further below.

lucab:
@steveeJ I've amended with one relevant suggestion of yours, and re-pushed.

graph-builder/src/graph.rs
steveej: You could read `GRAPH_FINAL_RELEASES` back, but I'm also fine keeping the variable. Just out of curiosity, can you think of a situation where that global value would be different from the global variable?

lucab: Fair point. I think I'm generally approaching metrics as write-only for the sake of logic separation, as the values flow from our logic to the metrics and not viceversa.
The only real difference is that those have different types (unsigned vs signed), but I think I'm more concerned over keeping metrics as write-only over possible sign-conversion issues.

PR #138, opened 2019-08-15
extract graph-fetcher code from the policy-engine into an internal plugin

steveej:
Depends on
- [x] #133 
- [x] #141.

TODO:
- [x] Discuss the metrics approach for internal plugins
- [x] Push `CustomDebug` as far down as possible
- [x] Rename the `http_upstream_unreachable` to `http_upstream_errors`
- [x] Ensure there are no leftover panics (unwraps/expects)
- [x] Elevate the `GraphError`s from plugins a level up to make them appear as native errors to the requester.

### DIFF
--- a/cincinnati/Cargo.toml
+++ b/cincinnati/Cargo.toml
@@ -22,9 +22,16 @@ env_logger = "^0.6.0"
 lazy_static = "^1.2.0"
 regex = "^1.1.0"
 smart-default = "^0.5.2"
+reqwest = "^0.9.19"
+actix-web = "^1.0.2"
+serde_json = "^1.0.22"
+prometheus = "^0.7.0"
+custom_debug_derive = "^0.1.7"
 
 [dev-dependencies]
 serde_json = "1.0.22"
+mockito = "^0.20.0"
+twoway = "^0.2"
 
 [build-dependencies]
 protoc-rust = "2.8"
--- a/cincinnati/src/plugins/catalog.rs
+++ b/cincinnati/src/plugins/catalog.rs
@@ -4,6 +4,7 @@
 //! referenced by name. It is used for configuration purposes.
 
 use super::internal::channel_filter::ChannelFilterPlugin;
+use super::internal::cincinnati_graph_fetch::CincinnatiGraphFetchPlugin;
 use super::internal::edge_add_remove::EdgeAddRemovePlugin;
 use super::internal::metadata_fetch_quay::QuayMetadataFetchPlugin;
 use super::internal::node_remove::NodeRemovePlugin;
@@ -17,7 +18,7 @@ static CONFIG_PLUGIN_NAME_KEY: &str = "name";
 /// Settings for a plugin.
 pub trait PluginSettings: Debug + Send {
     /// Build the corresponding plugin for this configuration.
-    fn build_plugin(&self) -> Fallible<BoxedPlugin>;
+    fn build_plugin(&self, registry: Option<&prometheus::Registry>) -> Fallible<BoxedPlugin>;
 }
 
 /// Validate configuration for a plugin and fill in defaults.
@@ -34,8 +35,25 @@ pub fn deserialize_config(cfg: toml::Value) -> Fallible<Box<dyn PluginSettings>>
         EdgeAddRemovePlugin::PLUGIN_NAME => EdgeAddRemovePlugin::deserialize_config(cfg),
         NodeRemovePlugin::PLUGIN_NAME => NodeRemovePlugin::deserialize_config(cfg),
         QuayMetadataFetchPlugin::PLUGIN_NAME => QuayMetadataFetchPlugin::deserialize_config(cfg),
+        CincinnatiGraphFetchPlugin::PLUGIN_NAME => {
+            CincinnatiGraphFetchPlugin::deserialize_config(cfg)
+        }
         x => bail!("unknown plugin '{}'", x),
     }
+}
+
+/// Bulid a vector of plugins from PluginSettings
+pub fn build_plugins(
+    policies: &[Box<dyn PluginSettings>],
+    registry: Option<&prometheus::Registry>,
+) -> Fallible<Vec<BoxedPlugin>> {
+    let mut plugins = Vec::with_capacity(policies.len());
+    for conf in policies {
+        let plugin = conf.build_plugin(registry)?;
+        plugins.push(plugin);
+    }
+
+    Ok(plugins)
 }
 
 #[cfg(test)]
@@ -52,7 +70,7 @@ mod tests {
 
         let node_remove_default: toml::Value = toml::from_str("name = 'node-remove'").unwrap();
         let nr_settings = deserialize_config(node_remove_default).unwrap();
-        nr_settings.build_plugin().unwrap();
+        nr_settings.build_plugin(None).unwrap();
 
         let cfg = r#"
             name = "quay-metadata"
@@ -60,6 +78,6 @@ mod tests {
         "#;
         let quay_metadata_repo: toml::Value = toml::from_str(cfg).unwrap();
         let qm_settings = deserialize_config(quay_metadata_repo).unwrap();
-        qm_settings.build_plugin().unwrap();
+        qm_settings.build_plugin(None).unwrap();
     }
 }
--- a/cincinnati/src/plugins/internal/cincinnati_graph_fetch.rs
+++ b/cincinnati/src/plugins/internal/cincinnati_graph_fetch.rs
@@ -1,0 +1,335 @@
+//! Plugin which implements fetching a Cincinnati graph via HTTP from a `/v1/graph`-compliant endpoint.
+//!
+//! Instead of processing the input graph, ghis plugin fetches a graph from a
+//! remote endpoint, which makes it effectively discard any given input graph.
+
+use crate::plugins::{
+    AsyncIO, BoxedPlugin, InternalIO, InternalPlugin, InternalPluginWrapper, PluginSettings,
+};
+use crate::CONTENT_TYPE;
+use commons::GraphError;
+use failure::Fallible;
+use futures::{future, Future, Stream};
+use prometheus::{Counter, Registry};
+use reqwest;
+use reqwest::header::{HeaderValue, ACCEPT};
+
+/// Default URL to upstream graph provider.
+pub static DEFAULT_UPSTREAM_URL: &str = "http://localhost:8080/v1/graph";
+
+/// Plugin settings.
+#[derive(Clone, CustomDebug, Deserialize, SmartDefault)]
+#[serde(default)]
+struct CincinnatiGraphFetchSettings {
+    #[default(DEFAULT_UPSTREAM_URL.to_string())]
+    upstream: String,
+}
+
+/// Graph fetcher for Cincinnati `/v1/graph` endpoints.
+#[derive(CustomDebug)]
+pub struct CincinnatiGraphFetchPlugin {
+    /// The upstream from which to fetch the graph
+    pub upstream: String,
+
+    /// The optional metric for counting upstream requests
+    #[debug(skip)]
+    pub http_upstream_reqs: Counter,
+
+    /// The optional metric for counting failed upstream requests
+    #[debug(skip)]
+    pub http_upstream_errors_total: Counter,
+}
+
+impl PluginSettings for CincinnatiGraphFetchSettings {
+    fn build_plugin(&self, registry: Option<&Registry>) -> Fallible<BoxedPlugin> {
+        let cfg = self.clone();
+        let plugin = CincinnatiGraphFetchPlugin::try_new(cfg.upstream, registry)?;
+        Ok(new_plugin!(InternalPluginWrapper(plugin)))
+    }
+}
+
+impl CincinnatiGraphFetchPlugin {
+    /// Plugin name, for configuration.
+    pub const PLUGIN_NAME: &'static str = "cincinnati-graph-fetch";
+
+    /// Validate plugin configuration and fill in defaults.
+    pub fn deserialize_config(cfg: toml::Value) -> Fallible<Box<dyn PluginSettings>> {
+        let settings: CincinnatiGraphFetchSettings = cfg.try_into()?;
+
+        ensure!(!settings.upstream.is_empty(), "empty upstream");
+
+        Ok(Box::new(settings))
+    }
+
+    fn try_new(
+        upstream: String,
+        prometheus_registry: Option<&prometheus::Registry>,
+    ) -> Fallible<Self> {
+        let http_upstream_reqs = Counter::new(
+            "http_upstream_requests_total",
+            "Total number of HTTP upstream requests",
+        )?;
+
+        let http_upstream_errors_total = Counter::new(
+            "http_upstream_errors_total",
+            "Total number of HTTP upstream unreachable errors",
+        )?;
+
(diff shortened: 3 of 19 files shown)


steveej:
Putting this on 

/hold

just to prevent an accidental merge.

cincinnati/src/plugins/internal/cincinnati_graph_fetch.rs
```diff
@@ -0,0 +1,281 @@
...
+    /// Validate plugin configuration and fill in defaults.
+    pub fn deserialize_config(cfg: toml::Value) -> Fallible<Box<PluginSettings>> {
+        let settings: CincinnatiGraphFetchSettings = cfg.try_into()?;
+
+        ensure!(!settings.upstream.is_empty(), "empty upstream");
+
+        Ok(Box::new(settings))
+    }
+
+    pub fn try_new(
```
lucab: I think this should not be `pub`.

cincinnati/src/plugins/internal/cincinnati_graph_fetch.rs
```diff
@@ -0,0 +1,281 @@
...
+#[serde(default)]
+struct CincinnatiGraphFetchSettings {
+    #[default(DEFAULT_UPSTREAM_URL.to_string())]
+    upstream: String,
+}
+
+/// Graph fetcher for Cincinnati `/v1/graph` endpoints.
+pub struct CincinnatiGraphFetchPlugin {
+    pub upstream: String,
+    pub prometheus_registry: Option<Registry>,
```
lucab: Do we need to store this one internally? I think it's only used when building the plugin, and not anymore after that.

cincinnati/src/plugins/internal/cincinnati_graph_fetch.rs
```diff
@@ -0,0 +1,281 @@
...
+        let settings: CincinnatiGraphFetchSettings = cfg.try_into()?;
+
+        ensure!(!settings.upstream.is_empty(), "empty upstream");
+
+        Ok(Box::new(settings))
+    }
+
+    pub fn try_new(
+        upstream: String,
+        prometheus_registry: Option<prometheus::Registry>,
```
lucab: I think this does not need to be wrapped in an `Option` (or, at least, I didn't spot why you'd need it).

cincinnati/src/plugins/internal/cincinnati_graph_fetch.rs
```diff
@@ -0,0 +1,281 @@
...
+    #[default(DEFAULT_UPSTREAM_URL.to_string())]
+    upstream: String,
+}
+
+/// Graph fetcher for Cincinnati `/v1/graph` endpoints.
+pub struct CincinnatiGraphFetchPlugin {
+    pub upstream: String,
+    pub prometheus_registry: Option<Registry>,
+    pub http_upstream_reqs: Option<Counter>,
+    pub http_upstream_unreachable: Option<Counter>,
```
lucab: Why are these two `Option`al?

cincinnati/src/plugins/internal/cincinnati_graph_fetch.rs
```diff
@@ -0,0 +1,281 @@
...
+/// Plugin settings.
+#[derive(Clone, Debug, Deserialize, SmartDefault)]
+#[serde(default)]
+struct CincinnatiGraphFetchSettings {
+    #[default(DEFAULT_UPSTREAM_URL.to_string())]
+    upstream: String,
+}
+
+/// Graph fetcher for Cincinnati `/v1/graph` endpoints.
+pub struct CincinnatiGraphFetchPlugin {
```
lucab, line 30: Sorry for the name bikeshedding, but I think this is effectively an `UpstreamCincinnatiPlugin` (or similar).

cincinnati/src/plugins/internal/cincinnati_graph_fetch.rs
steveej: The same thought occurred to me while being offline. Looks like we have consensus then!

cincinnati/src/plugins/internal/cincinnati_graph_fetch.rs
steveej: I wanted to make the metrics optional, and so the registry isn't always needed.

cincinnati/src/plugins/internal/cincinnati_graph_fetch.rs
steveej: I'm fine renaming it. I tried to form a scheme together with the _QuayMetadataFetchPlugin_.

cincinnati/src/plugins/internal/cincinnati_graph_fetch.rs
steveej: Same as for the registry, I wanted to make the metrics optional.

cincinnati/src/plugins/internal/cincinnati_graph_fetch.rs
steveej: I didn't put much thought into it. I wanted to clarify with you what the way of instantiating the plugins will be once we have the config file.

cincinnati/src/plugins/internal/cincinnati_graph_fetch.rs
lucab: From an out of band discussion: let's keep the current name, for consistency.

cincinnati/src/plugins/internal/cincinnati_graph_fetch.rs
lucab: I think `deserialize_config` and ` build_plugin` should be enough as a public interface towards the applications. We may want to tweak the latter to pass the prometheus registry, though.

cincinnati/src/plugins/internal/cincinnati_graph_fetch.rs
lucab: From out-of-band discussion: this is to make testing boilerplate less heavy.

steveeJ:
@lucab I rebased this after we merged #146 and #147 (thanks for these by the way!). Also, this is ready for review from my side.

steveej:
@lucab I rebased this after we merged #146 and #147 (thanks for these by the way!). Also, this is ready for review from my side.

cincinnati/src/plugins/internal/cincinnati_graph_fetch.rs
```diff
@@ -0,0 +1,308 @@
...
+    upstream: String,
+}
+
+/// Graph fetcher for Cincinnati `/v1/graph` endpoints.
+#[derive(CustomDebug)]
+pub struct CincinnatiGraphFetchPlugin {
+    /// The upstream from which to fetch the graph
+    pub upstream: String,
+
+    /// The optinal metric for counting upstrema requests
```
lucab: Typos:
 - **optional**
 - **upstream**

cincinnati/src/plugins/internal/cincinnati_graph_fetch.rs
```diff
@@ -0,0 +1,308 @@
+//! Plugin which implements fetching a Cincinnati graph via HTTP from a `/v1/graph`-compliant endpoint.
+//!
+//! This plugin is special so that it is expected to be the first plugin in the
```
lucab: This is not strictly true, right?
Anything doing client-parameters validation/manipulation should be happening before this, I guess.

cincinnati/src/plugins/internal/cincinnati_graph_fetch.rs
```diff
@@ -0,0 +1,308 @@
...
+                error!("failed upstream request. {}", e);
+                http_upstream_errors_total_failed_request.inc();
+                GraphError::FailedUpstreamFetch(e.to_string())
+            })
+            .and_then(move |res| {
+                if res.status().is_success() {
+                    future::ok(res)
+                } else {
+                    error!("unsucessful status on upstream request: {}", res.status());
+                    http_upstream_errors_total_wrong_status.inc();
```
lucab: I think it may be possible to refactor this and the `http_upstream_errors_total_failed_request.inc();` above together, by pushing them down into a single `.map_err()` placed at the bottom.

cincinnati/src/plugins/internal/cincinnati_graph_fetch.rs
```diff
@@ -0,0 +1,308 @@
...
+                    future::err(GraphError::FailedUpstreamFetch(res.status().to_string()))
+                }
+            })
+            .and_then(|res| {
+                res.into_body()
+                    .concat2()
+                    .map_err(|e| GraphError::FailedUpstreamFetch(e.to_string()))
```
lucab: It looks like this error case is not accounted into the metrics counter.

cincinnati/src/plugins/internal/cincinnati_graph_fetch.rs
```diff
@@ -0,0 +1,308 @@
...
+            })
+            .and_then(|body| {
+                serde_json::from_slice(&body).map_err(|e| GraphError::FailedJsonIn(e.to_string()))
```
lucab: It looks like this error case is not accounted into the metrics counter.

cincinnati/src/plugins/catalog.rs
```diff
@@ -17,7 +18,7 @@ static CONFIG_PLUGIN_NAME_KEY: &str = "name";
 /// Settings for a plugin.
 pub trait PluginSettings: Debug + Send {
     /// Build the corresponding plugin for this configuration.
-    fn build_plugin(&self) -> Fallible<BoxedPlugin>;
+    fn build_plugin(&self, registry: Option<&prometheus::Registry>) -> Fallible<BoxedPlugin>;
```
lucab, line 21: Side-note for the next time: this would have been great as its own commit/PR.

cincinnati/src/plugins/macros.rs
```diff
@@ -48,6 +48,17 @@ macro_rules! new_plugins {
     ($($x:expr),*) => { vec![$(new_plugin!($x)),*] };
 }
 
+#[macro_export]
+macro_rules! plugin_config {
```
lucab, line 52: Side-note for the next time: this would have been great as its own commit/PR.

policy-engine/src/config/settings.rs
```diff
@@ -73,19 +73,20 @@ impl AppSettings {
...
-
-        // TODO(lucab): drop this as soon as all config-maps are in place (prod & staging).
-        if plugins.is_empty() {
-            plugins = default_openshift_plugins();
-        }
+    pub fn policy_plugins(
+        &self,
+        registry: Option<&prometheus::Registry>,
+    ) -> Fallible<Vec<BoxedPlugin>> {
+        // TODO(steveeJ):  prevent this call in case it's not required later
```
lucab: In the spirit of this TODO, perhaps in the short term it is better to move the `?` in the non-empty conditional branch below?

policy-engine/src/config/settings.rs
```diff
@@ -94,18 +95,33 @@ impl AppSettings {
...
-    // TODO(lucab): drop this as soon as all config-maps are in place (prod & staging).
-    use cincinnati::plugins::internal::channel_filter::ChannelFilterPlugin;
-    use cincinnati::plugins::internal::metadata_fetch_quay::DEFAULT_QUAY_LABEL_FILTER;
-    use cincinnati::plugins::prelude::*;
-
-    new_plugins!(InternalPluginWrapper(ChannelFilterPlugin {
-        key_prefix: String::from(DEFAULT_QUAY_LABEL_FILTER),
-        key_suffix: String::from("release.channels"),
-    }))
+    fn default_openshift_policies(&self) -> Fallible<Vec<Box<dyn PluginSettings>>> {
```
lucab, line 105: While reading this function, at first I didn't realize why the return type here needed to be Fallible.
To make it less surprising, it would be nice to move out the `?` which is currently hidden inside `plugin_config!()`.

cincinnati/Cargo.toml
```diff
@@ -22,9 +22,16 @@ env_logger = "^0.6.0"
...
 smart-default = "^0.5.2"
+hyper = "^0.12.6"
+actix-web = "^1.0.2"
+serde_json = "^1.0.22"
+prometheus = "^0.7.0"
+custom_debug_derive = "^0.1.7"
 
 [dev-dependencies]
 serde_json = "1.0.22"
+mockito = "^0.17.1"
```
lucab: This one can be bumped to 0.20.0.

policy-engine/src/config/settings.rs
steveej: Unpacking the result in the if scope creates the same issue as moving the call there altogether:

```
error[E0716]: temporary value dropped while borrowed
  --> policy-engine/src/config/settings.rs:83:14
   |
82 |         let policies: &Vec<Box<dyn PluginSettings>> = if self.policies.is_empty() {
   |             -------- borrow later stored here
83 |             &default_policies?
   |              ^^^^^^^^^^^^^^^^^ creates a temporary which is freed while still in use
84 |         } else {
   |         - temporary value is freed at the end of this statement
```

cincinnati/src/plugins/internal/cincinnati_graph_fetch.rs
steveej: Indeed. I removed that part of the description.

cincinnati/src/plugins/internal/cincinnati_graph_fetch.rs
steveej: Coincidentally we already had that bottom `.map_err()`. I'm currently trying to add tests to trigger an error in every combinator, and so far I can confirm your idea!

cincinnati/src/plugins/internal/cincinnati_graph_fetch.rs
steveej: The outer scope's `.map_err()` will also catch this one

cincinnati/src/plugins/internal/cincinnati_graph_fetch.rs
steveej: The bottom `.map_err()` will also catch this one

policy-engine/src/config/settings.rs
lucab: Ack, let's keep it as is for the moment.

steveej:
@lucab thanks for the review! I addressed all comments. I tried to cover all error cases from the fetch request processing chain in a test-case which worked for all but the `concat2()` call, which isn't straight forward to trigger. I'm open for suggestions but I'm fine leaving that in as a TODO, as the more general question -- if the last `.map_err()` covers all the previous combinators works -- is resolved.

cincinnati/src/plugins/internal/cincinnati_graph_fetch.rs
```diff
@@ -0,0 +1,335 @@
...
+
+/// Default URL to upstream graph provider.
+pub static DEFAULT_UPSTREAM_URL: &str = "http://localhost:8080/v1/graph";
+
+/// Plugin settings.
+#[derive(Clone, CustomDebug, Deserialize, SmartDefault)]
+#[serde(default)]
+struct CincinnatiGraphFetchSettings {
+    #[default(DEFAULT_UPSTREAM_URL.to_string())]
+    upstream: String,
```
lucab, line 25: Side note: we talked about splitting this into a `base_url` + `api_endpoint` later on.

steveej:
Thanks @lucab!

/hold cancel

PR #140, opened 2019-08-19
policy-engine/graph: split query params into at most 2 parts

steveej:
The overarching assumption is that we're dealing with K=V query
parameters. Before this change any characters after a '=' in the value
are not passed on as a plugin parameter value. This leads to incorrect
behavior, e.g. ignoring invalid values or losing information in case the
'=' is actually valid.

### DIFF
--- a/policy-engine/src/graph.rs
+++ b/policy-engine/src/graph.rs
@@ -2,6 +2,7 @@
 
 use crate::AppState;
 use actix_web::http::header::{self, HeaderValue};
+use actix_web::web::Query;
 use actix_web::{HttpRequest, HttpResponse};
 use cincinnati::CONTENT_TYPE;
 use commons::{self, GraphError};
@@ -10,6 +11,7 @@ use futures::{future, Future, Stream};
 use hyper::{Body, Client, Request};
 use prometheus::{Counter, Histogram, HistogramOpts, Registry};
 use serde_json;
+use std::collections::HashMap;
 
 lazy_static! {
     static ref HTTP_UPSTREAM_REQS: Counter = Counter::new(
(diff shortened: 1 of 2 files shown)


policy-engine/src/graph.rs
```diff
@@ -61,14 +61,14 @@ pub(crate) fn index(req: HttpRequest) -> Box<Future<Item = HttpResponse, Error =
     if let Err(e) = commons::ensure_query_params(mandatory_params, req.query_string()) {
         return Box::new(future::err(e));
     }
-    // TODO(steveeJ): take another look at the actix-web docs for a method that
-    // provides this parameters split.
```
lucab, line 66: Have you seen/considered https://docs.rs/actix-web/1.0.5/actix_web/web/struct.Query.html? Can this be done by deserializing to a serde-transparent `HashMap<String, String>`?

steveej: I saw it but I couldn't make much sense of it as the docstring seems to be outdated. It mentions a `username` field which doesn't exist in the example code. Further it's not obvious to me how to access the request headers if the `Query` type is used.

steveej: Looking at the docs again I saw `Query::from_query` which works for our needs :raised_hands: I added a commit on top which I'll squash if we agree on it.

policy-engine/src/graph.rs
```diff
@@ -61,29 +63,15 @@ pub(crate) fn index(req: HttpRequest) -> Box<Future<Item = HttpResponse, Error =
...
-                    &""
-                })
-                .to_string();
-
-            (kv_split[0].to_string(), value)
-        })
-        .collect();
+
+    let plugin_params = match Query::from_query(req.query_string()) {
+        Ok(query) => (query as Query<HashMap<String, String>>).into_inner(),
```
lucab: Just a minor question: is there a way to move this type hint to the call site? Something like `Query<HashMap<String, String>>::from_query()` or a similar syntax, if that exists.

steveej: This is indeed possible, thanks for the suggestion!

PR #141, opened 2019-08-20
policy-engine/metrics: don't use a global registry

steveej:
Depends on 
- [x] #133.


lucab:
:heart:

policy-engine/src/config/settings.rs
```diff
@@ -46,6 +49,14 @@ pub struct AppSettings {
...
     /// Required client parameters for the main service.
     pub mandatory_client_parameters: HashSet<String>,
+
+    /// Whether to spawn the metrics service.
+    #[default(true)]
+    pub enable_metrics: bool,
+
+    /// Common prefix for policy-engine metrics.
+    #[default(DEFAULT_METRICS_PREFIX.to_string())]
+    pub metrics_prefix: String,
```
lucab: I'd rather not have these two configuration knobs.
We can discuss them separately if you want, but I wouldn't put them in this PR anyway.

steveej: >  We can discuss them separately if you want, but I wouldn't put them in this PR anyway.

I think a discussion will be helpful here. Why would you prefer not to have them?

lucab: For several reasons, most of them related to putting this into a larger context:
 * metrics names are fixed by the application, not by runtime configuration, and we want them to be stable/reliable across deployments
 * metrics gathering is a first-class part of the application, and we want to encourage more people to consume them (without breakages)
 * an high number of unused configuration knobs just increase the cognitive load on operators and the maintenance cost on developers
 * the status service is an important (i.e. non-conditional) piece of the graph-builder. I'd like the policy-engine to not diverge from GB, unless where strictly required

steveej: >     * metrics names are fixed by the application, not by runtime configuration, and we want them to be stable/reliable across deployments

I guess they are hard-coded until they're not? In a similar fashion we also have the `path_prefix` for the main service in configurable, which is also quite an important service. For reliability we could remove the default and force the operator to provide the value.

>     * the status service is an important (i.e. non-conditional) piece of the graph-builder. I'd like the policy-engine to not diverge from GB, unless where strictly required

I was going to suggest to make the equivalent changes to the GB.

>     * an high number of unused configuration knobs just increase the cognitive load on operators and the maintenance cost on developers

By removing the default we would force it to be used (and known) by the operator.

>     * metrics gathering is a first-class part of the application, and we want to encourage more people to consume them (without breakages)

Not sure what to say about this though I of course agree that we try hard to not break consumers.

---

Now that I gave my view and envisioned directions on the above items, I want to state that my opinion isn't strong enough to force this through. I just added the setting for completeness after splitting the prefix out of the metrics for not repeating the magic string in the tests. I'll agree to remove the user-facing setting from this PR and possibly revisit this at a later time. ACK?

steveej:
@lucab PTAL at 76192ced4839c58de9404bff354bc8aba7be134c which shares the metrics module among the two services. Some trickery is involved with how the registry is stored and retrieved into the web service; it looks bearable to me.

policy-engine/src/config/settings.rs
lucab: We started without a `path_prefix` and we added it once we had a usecase/consumer for that. My point here is that a fully-configurable registry has no such usecase/consumer (surely now, and possible in the future), and additionally it isn't an encouraged pattern for metrics.

For context, people are expected to write and share monitoring mixins, like https://github.com/kubernetes-monitoring/kubernetes-mixin/blob/master/dashboards/apiserver.libsonnet.

> I just added the setting for completeness after splitting the prefix out of the metrics for not repeating the magic string in the tests

Thanks, now I see where this is coming from. Based on that, my preference would be:
 * unifying PE towards GB, i.e. a non-conditional status service, with shared logic
 * not introducing neither a `registry_enabled` nor a `registry_prefix` configuration knob
 * carrying the registry_prefix somewhere in the application (for code sharing), but making it hardcoded

steveej: I see, thanks for elaborating! As far as the metrics is concerned the code is shared in the latest commit of this PR. I would postpone the full status service for the PE as this PR is only about refactoring the metrics code.

lucab: Ack, SGTM. Ping me back when you drop the WIP label.

steveej: @lucab :bell:

PR #142, opened 2019-08-22
fix compiler warnings with version 1.37.0 -- includes a rust-protoc bump and code regeneration

steveej:


lucab:
Note to self: the protoc-rust bump is in order to bring in https://github.com/stepancheg/rust-protobuf/commit/df9780f4bc4328dc23e29ece8dab20fc88b975b7.

PR #147, opened 2019-08-26
cincinnati: fix and switch to Rust edition 2018

lucab:
This runs a `cargo fix --edition` on `cincinnati` crate and then switches it to Rust edition 2018.

/cc @steveej


lucab:
Somehow we missed to convert this one. I checked and it is the last one not yet at edition-2018.

PR #148, opened 2019-08-30
readme: update example curl call with channel name and expected output

vrutkovs:
This PR adds required `channel` key and sample output to the curl command in README

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,30 @@ export CINCINNATI_REPO="redhat/openshift-cincinnati-test-public-manual"
 cargo run --package graph-builder -- --service.address 0.0.0.0 --upstream.registry.url "${CINCINNATI_REGISTRY}" --upstream.registry.repository "${CINCINNATI_REPO}" &
 cargo run --package policy-engine -- --service.address 0.0.0.0 &
 curl --verbose --header 'Accept:application/json' http://localhost:8081/v1/graph
+{
+    "nodes":
+    [{
+        "version":"0.0.0",
+        "payload":"quay.io/redhat/openshift-cincinnati-test-public-manual@sha256:a264db3ac5288c9903dc3db269fca03a0b122fe4af80b57fc5087b329995013d",
+        "metadata":{
+            "io.openshift.upgrades.graph.release.channels":"a",
+            "io.openshift.upgrades.graph.release.manifestref":"sha256:a264db3ac5288c9903dc3db269fca03a0b122fe4af80b57fc5087b329995013d"
+        }
+    },
+    {
+        "version":"0.0.1",
+        "payload":"quay.io/redhat/openshift-cincinnati-test-public-manual@sha256:73df5efa869eaf57d4125f7655e05e1a72b59d05e55fea06d3701ea5b59234ff",
+        "metadata":{
+            "io.openshift.upgrades.graph.release.manifestref":"sha256:73df5efa869eaf57d4125f7655e05e1a72b59d05e55fea06d3701ea5b59234ff",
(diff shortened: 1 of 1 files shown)


lucab:
But the channel name is not a parameter required by the Cincinnati protocol, though.
It is just a (configuration) detail of the current OpenShift deployment.
That one is documented, with a short example, here: https://github.com/openshift/cincinnati/blob/master/docs/design/openshift.md#request.

vrutkovs:
The sample command doesn't seem to work without it anyway:

```
$ curl --verbose --header 'Accept:application/json' http://localhost:8081/v1/graph
...
{"kind":"failed_plugin_execution","value":"failed to execute plugins: could not find key 'channel'"}
```
Is this due to the channel plugin being autoenabled?

lucab:
That's a fact, I'm not disputing it. However here you are observing the symptom of this TODO item: https://github.com/openshift/cincinnati/blob/ffdc1ff7d3e63b0ddf698f549b1748c4da2c2a0e/policy-engine/src/config/settings.rs#L102.

We should indeed be fixing it (and we are) but in the opposite direction of this PR (i.e. by un-hardcoding OpenShift details).

vrutkovs:
Ah, okay, I see. So the same JSON is expected (but without required `channel` once config maps are done), right?

lucab:
The structure yes, as shown here: https://github.com/openshift/cincinnati/blob/master/docs/design/cincinnati.md#example.

The content semantics (value and metadata) no, as that is specific to the current OpenShift pipeline.

To see the difference in practice, check the output of `curl --header 'Accept:application/json' 'https://updates.coreos.stg.fedoraproject.org/v1/graph?basearch=x86_64&stream=testing'`

vrutkovs:
Added a short note about that. It doesn't really sound good so I'm not sure if its worth keeping

lucab:
I'd leave this one for final review to @steveej.
I am not for over-stuffing the readme, but I acknowledge that is just my personal taste.

README.md
```diff
@@ -18,8 +18,31 @@ export CINCINNATI_REPO="redhat/openshift-cincinnati-test-public-manual"
 cargo run --package graph-builder -- --service.address 0.0.0.0 --upstream.registry.url "${CINCINNATI_REGISTRY}" --upstream.registry.repository "${CINCINNATI_REPO}" &
 cargo run --package policy-engine -- --service.address 0.0.0.0 &
 curl --verbose --header 'Accept:application/json' http://localhost:8081/v1/graph
```
steveej:
```suggestion
 curl --verbose --header 'Accept:application/json' http://localhost:8081/v1/graph\?channel=a
```

README.md
```diff
@@ -18,8 +18,31 @@ export CINCINNATI_REPO="redhat/openshift-cincinnati-test-public-manual"
...
+            "io.openshift.upgrades.graph.release.manifestref":"sha256:73df5efa869eaf57d4125f7655e05e1a72b59d05e55fea06d3701ea5b59234ff",
+            "io.openshift.upgrades.graph.release.channels":"a",
+            "kind":"test"
+        }
+    }],
+    "edges":[[0,1]]
+}
 ```
 
+Note, that although Cincinnati protocol doesn't require additional params, current build from master requires `?channel` to be set (use `a` or `b` for test images)
```
steveej:
```suggestion
***Note:*** the default configuration of the policy-engine requires the `channel` parameter to be present in each request.
```

steveej (changes requested):
Thanks for updating the example output. I've added suggestions for the example curl command and adjusted the note.

steveej (approved):
Thanks!

/lgtm

PR #151, opened 2019-09-03
channel_filter: return GraphErrors instead of FailedPluginExecution

vrutkovs:
MissingParams error is returned if channel is not set, InvalidParams error - if channel does not match the regexp.

Ref: https://jira.coreos.com/browse/CIN-69

```
$ curl -s --header 'Accept:application/json' http://localhost:8081/v1/graph
{"kind":"missing_params","value":"mandatory client parameters missing: could not find key 'channel'"}
$ curl -s --header 'Accept:application/json' http://localhost:8081/v1/graph\?channel\=invalid:channel
{"kind":"invalid_params","value":"invalid client parameters: channel 'invalid:channel' does not match regex '^[0-9a-z\\-\\.]+$'"}
```

### DIFF
--- a/cincinnati/src/plugins/internal/channel_filter.rs
+++ b/cincinnati/src/plugins/internal/channel_filter.rs
@@ -5,6 +5,7 @@
 use crate::plugins::{
     AsyncIO, BoxedPlugin, InternalIO, InternalPlugin, InternalPluginWrapper, PluginSettings,
 };
+use commons::GraphError;
 use failure::Fallible;
 use futures::Future;
 use prometheus::Registry;
@@ -58,14 +59,15 @@ impl InternalPlugin for ChannelFilterPlugin {
             self.key_suffix.to_owned(),
         ))
         .and_then(|(internal_io, key_prefix, key_suffix)| {
-            let channel = get_multiple_values!(internal_io.parameters, "channel")?.clone();
+            let channel = get_multiple_values!(internal_io.parameters, "channel")
+                .map_err(|e| GraphError::MissingParams(vec![e.to_string()]))?
+                .clone();
 
             if !CHANNEL_VALIDATION_REGEX_RE.is_match(&channel) {
-                bail!(
+                return Err(GraphError::InvalidParams(format!(
                     "channel '{}' does not match regex '{}'",
-                    channel,
-                    CHANNEL_VALIDATION_REGEX_STR
-                );
+                    channel, CHANNEL_VALIDATION_REGEX_STR
+                )));
             };
 
             let mut graph = internal_io.graph;
@@ -106,7 +108,8 @@ impl InternalPlugin for ChannelFilterPlugin {
                 graph,
                 parameters: internal_io.parameters,
             })
-        });
+        })
+        .map_err(move |e| failure::Error::from(e));
 
         Box::new(future_result)
     }
--- a/cincinnati/src/plugins/macros.rs
+++ b/cincinnati/src/plugins/macros.rs
@@ -9,7 +9,7 @@ macro_rules! get_multiple_values {
                     if let Some(value) = $map.get($key) {
                         value
                     } else {
-                        bail!("could not find key '{}'", $key)
+                        bail!("{}", $key)
                     },
                 )
             };
@@ -25,7 +25,7 @@ macro_rules! get_multiple_values {
                             if let Some(value) = $map.get($key) {
                                 value
                             } else {
-                                bail!("could not find key '{}'", $key)
+                                bail!("{}", $key)
                             },
                         )*
                     )
--- a/commons/src/errors.rs
+++ b/commons/src/errors.rs
@@ -57,7 +57,7 @@ pub enum GraphError {
     MissingParams(Vec<String>),
 
     /// Invalid client parameters.
-    #[fail(display = "invalid client parameters")]
+    #[fail(display = "invalid client parameters: {}", _0)]
     InvalidParams(String),
 }
 
--- a/policy-engine/src/graph.rs
+++ b/policy-engine/src/graph.rs
@@ -219,12 +219,12 @@ mod tests {
             .create();
 
         match rt.block_on(graph_call) {
-            Err(graph::GraphError::FailedPluginExecution(ref msg))
+            Err(graph::GraphError::InvalidParams(ref msg))
                 if msg.contains("does not match regex") =>
             {
                 Ok(())
             }
-            res => Err(format!("expected FailedPluginExecution error, got: {:?}", res).into()),
+            res => Err(format!("expected InvalidParams error, got: {:?}", res).into()),
         }
     }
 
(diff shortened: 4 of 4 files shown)


lucab:
ok to test

cincinnati/src/plugins/internal/channel_filter.rs
```diff
@@ -56,14 +57,17 @@ impl InternalPlugin for ChannelFilterPlugin {
             self.key_suffix.to_owned(),
         ))
         .and_then(|(internal_io, key_prefix, key_suffix)| {
-            let channel = get_multiple_values!(internal_io.parameters, "channel")?.clone();
+            let channel = match get_multiple_values!(internal_io.parameters, "channel"){
+                Ok(c) => c.clone(),
+                Err(e) => return Err(GraphError::MissingParams(vec!(e.to_string()))),
```
lucab: This should be `vec!["channel".to_string()]`.

cincinnati/src/plugins/internal/channel_filter.rs
```diff
@@ -56,14 +57,17 @@ impl InternalPlugin for ChannelFilterPlugin {
             self.key_suffix.to_owned(),
         ))
         .and_then(|(internal_io, key_prefix, key_suffix)| {
-            let channel = get_multiple_values!(internal_io.parameters, "channel")?.clone();
+            let channel = match get_multiple_values!(internal_io.parameters, "channel"){
```
lucab: I suspect we could directly fix the macro to return the proper error instead of `bail!()`, and adjust its name to make it explicitly scoped to client parameters. @steveej?
Sidenote: for the single value case here, this should be effectively equal to a `Option::ok_or()`.

cincinnati/src/plugins/internal/channel_filter.rs
```diff
@@ -104,6 +108,10 @@ impl InternalPlugin for ChannelFilterPlugin {
                 graph,
                 parameters: internal_io.parameters,
             })
+        })
+        .map_err(move |e| {
+            error!("error filtering graph: {}", e);
```
lucab: Policy-engine should be already logging this error, can you confirm?

lucab (commented):
Good patch! I have just a few minor comments.

cincinnati/src/plugins/internal/channel_filter.rs
vrutkovs: I was trying to avoid hardcoding "channel" in yet another place, but sure, I'll change that

cincinnati/src/plugins/internal/channel_filter.rs
vrutkovs: I'd rather wrap the error in the plugin so that this macros could be reused elsewhere

cincinnati/src/plugins/internal/channel_filter.rs
lucab: Sorry, let me rephrase for clarity: the vector should contain only the keynames of missing parameters, not full human-friendly error strings.

cincinnati/src/plugins/internal/channel_filter.rs
vrutkovs: Right, this is being ignored - removed

cincinnati/src/plugins/internal/channel_filter.rs
steveeJ: How about this:


```diff
diff --git a/cincinnati/src/plugins/internal/channel_filter.rs b/cincinnati/src/plugins/internal/channel_filter.rs
index 87231eb..c26dac5 100644
--- a/cincinnati/src/plugins/internal/channel_filter.rs
+++ b/cincinnati/src/plugins/internal/channel_filter.rs
@@ -57,10 +57,9 @@ impl InternalPlugin for ChannelFilterPlugin {
             self.key_suffix.to_owned(),
         ))
         .and_then(|(internal_io, key_prefix, key_suffix)| {
-            let channel = match get_multiple_values!(internal_io.parameters, "channel"){
-                Ok(c) => c.clone(),
-                Err(_) => return Err(GraphError::MissingParams(vec!["channel".to_string()])),
-            };
+            let channel = get_multiple_values!(internal_io.parameters, "channel")
+                .map_err(|e| GraphError::MissingParams(vec![e.to_string()]))?
+                .clone();

             if !CHANNEL_VALIDATION_REGEX_RE.is_match(&channel) {
                 return Err(GraphError::InvalidParams(format!(
diff --git a/cincinnati/src/plugins/macros.rs b/cincinnati/src/plugins/macros.rs
index 2cc216c..1911a64 100644
--- a/cincinnati/src/plugins/macros.rs
+++ b/cincinnati/src/plugins/macros.rs
@@ -9,7 +9,7 @@ macro_rules! get_multiple_values {
                     if let Some(value) = $map.get($key) {
                         value
                     } else {
-                        bail!("could not find key '{}'", $key)
+                        bail!("{}", $key)
                     },
                 )
             };
@@ -25,7 +25,7 @@ macro_rules! get_multiple_values {
                             if let Some(value) = $map.get($key) {
                                 value
                             } else {
-                                bail!("could not find key '{}'", $key)
+                                bail!("{}", $key)
                             },
                         )*
                     )
```

steveej: How about this:


```diff
diff --git a/cincinnati/src/plugins/internal/channel_filter.rs b/cincinnati/src/plugins/internal/channel_filter.rs
index 87231eb..c26dac5 100644
--- a/cincinnati/src/plugins/internal/channel_filter.rs
+++ b/cincinnati/src/plugins/internal/channel_filter.rs
@@ -57,10 +57,9 @@ impl InternalPlugin for ChannelFilterPlugin {
             self.key_suffix.to_owned(),
         ))
         .and_then(|(internal_io, key_prefix, key_suffix)| {
-            let channel = match get_multiple_values!(internal_io.parameters, "channel"){
-                Ok(c) => c.clone(),
-                Err(_) => return Err(GraphError::MissingParams(vec!["channel".to_string()])),
-            };
+            let channel = get_multiple_values!(internal_io.parameters, "channel")
+                .map_err(|e| GraphError::MissingParams(vec![e.to_string()]))?
+                .clone();

             if !CHANNEL_VALIDATION_REGEX_RE.is_match(&channel) {
                 return Err(GraphError::InvalidParams(format!(
diff --git a/cincinnati/src/plugins/macros.rs b/cincinnati/src/plugins/macros.rs
index 2cc216c..1911a64 100644
--- a/cincinnati/src/plugins/macros.rs
+++ b/cincinnati/src/plugins/macros.rs
@@ -9,7 +9,7 @@ macro_rules! get_multiple_values {
                     if let Some(value) = $map.get($key) {
                         value
                     } else {
-                        bail!("could not find key '{}'", $key)
+                        bail!("{}", $key)
                     },
                 )
             };
@@ -25,7 +25,7 @@ macro_rules! get_multiple_values {
                             if let Some(value) = $map.get($key) {
                                 value
                             } else {
-                                bail!("could not find key '{}'", $key)
+                                bail!("{}", $key)
                             },
                         )*
                     )
```

steveej (changes requested):
I added one comment as requested by @lucab. While trying it out locally I also noted that the code hasn't been run through rustfmt which we usually do.

steveej:
Thanks!

/lgtm

/hold
In case @lucab has any final remarks.

steveej (approved):

PR #152, opened 2019-09-04
Create a metric to track duration of upstream scrapes

vrutkovs:
Add a new histogram metric which measures duration of upstream scrapes. Initial scrape is not being included, saved into `initial_upstream_scrape_duration` instead.

Scrape duration is being recorded in `releases` validation (ignored if releases fetch failed). On first success `initial_upstream_scrape_duration` gauge is being later, in the following cycles duration is recorded in `graph_upstream_scrapes_duration` histogram.

`prometheus` crate is being built from git for now in order to be able to discard `HistogramTimer`, as its not included in tagged release yet

Ref: https://jira.coreos.com/browse/CIN-65

```
$ curl --verbose --header 'Accept:application/json' http://localhost:9080/metrics
...
# HELP cincinnati_gb_graph_initial_upstream_scrape_duration Duration of initial upstream scrape
# TYPE cincinnati_gb_graph_initial_upstream_scrape_duration gauge
cincinnati_gb_graph_initial_upstream_scrape_duration 7.338227124
...
# HELP cincinnati_gb_graph_upstream_scrapes_duration Upstream scrape duration in seconds
# TYPE cincinnati_gb_graph_upstream_scrapes_duration histogram
cincinnati_gb_graph_upstream_scrapes_duration_bucket{le="1"} 0
cincinnati_gb_graph_upstream_scrapes_duration_bucket{le="2"} 0
cincinnati_gb_graph_upstream_scrapes_duration_bucket{le="3"} 0
cincinnati_gb_graph_upstream_scrapes_duration_bucket{le="4"} 0
cincinnati_gb_graph_upstream_scrapes_duration_bucket{le="5"} 0
cincinnati_gb_graph_upstream_scrapes_duration_bucket{le="6"} 4
cincinnati_gb_graph_upstream_scrapes_duration_bucket{le="7"} 4
cincinnati_gb_graph_upstream_scrapes_duration_bucket{le="8"} 4
cincinnati_gb_graph_upstream_scrapes_duration_bucket{le="9"} 4
cincinnati_gb_graph_upstream_scrapes_duration_bucket{le="10"} 4
cincinnati_gb_graph_upstream_scrapes_duration_bucket{le="11"} 4
cincinnati_gb_graph_upstream_scrapes_duration_bucket{le="+Inf"} 4
cincinnati_gb_graph_upstream_scrapes_duration_sum 22.200288569
cincinnati_gb_graph_upstream_scrapes_duration_count 4
...
```

TODO:
* [x] a setting to adjust buckets?
  Hardcoded 1.0-11.0 buckets

### DIFF
--- a/graph-builder/src/graph.rs
+++ b/graph-builder/src/graph.rs
@@ -29,10 +29,6 @@ use std::collections::{HashMap, HashSet};
 use std::sync::Arc;
 use std::thread;
 
-/// Custom bucket values for upstream scraping duration in seconds
-pub const UPSTREAM_SCRAPE_BUCKETS: &[f64] =
-    &[1.0, 2.0, 3.0, 4.0, 5.0, 6.0, 7.0, 8.0, 9.0, 10.0, 11.0];
-
 lazy_static! {
     static ref GRAPH_FINAL_RELEASES: IntGauge = IntGauge::new(
         "graph_final_releases",
@@ -64,10 +60,11 @@ lazy_static! {
         "Duration of initial upstream scrape"
     )
     .unwrap();
+    /// Histogram with custom bucket values for upstream scraping duration in seconds
     static ref UPSTREAM_SCRAPES_DURATION: Histogram = Histogram::with_opts(histogram_opts!(
         "graph_upstream_scrapes_duration",
         "Upstream scrape duration in seconds",
-        UPSTREAM_SCRAPE_BUCKETS.to_vec()
+        vec![1.0, 2.0, 3.0, 4.0, 5.0, 6.0, 7.0, 8.0, 9.0, 10.0, 11.0]
     ))
     .unwrap();
     static ref V1_GRAPH_INCOMING_REQS: Counter = Counter::new(
@@ -186,11 +183,12 @@ pub fn run<'a>(settings: &'a config::AppSettings, state: &State) -> ! {
 
     // Don't wait on the first iteration
     let mut first_iteration = true;
-    // Don't collect upstream scrape duration on cold start
-    let mut first_scrape = true;
     let mut first_success = true;
 
     loop {
+        // Store scrape duration value. It would be used for initial scrape gauge or scrape histogram
+        let scrape_value: f64;
+
         if first_iteration {
             *state.live.write() = true;
             first_iteration = false;
@@ -220,16 +218,8 @@ pub fn run<'a>(settings: &'a config::AppSettings, state: &State) -> ! {
                     );
                 };
 
-                // Record scrape duration if it succeeded
-                // First iteration is slower due to a cold cache,
-                // so it is being stored in graph_initial_upstream_scrape_duration gauge instead
-
-                if first_scrape {
-                    first_scrape = false;
-                    GRAPH_UPSTREAM_INITIAL_SCRAPE.set(scrape_timer.stop_and_discard());
-                } else {
-                    scrape_timer.observe_duration();
-                }
+                // Record scrape duration
+                scrape_value = scrape_timer.stop_and_discard();
                 releases
             }
             Err(err) => {
@@ -283,7 +273,10 @@ pub fn run<'a>(settings: &'a config::AppSettings, state: &State) -> ! {
         if first_success {
             *state.ready.write() = true;
             first_success = false;
-        };
+            GRAPH_UPSTREAM_INITIAL_SCRAPE.set(scrape_value);
+        } else {
+            UPSTREAM_SCRAPES_DURATION.observe(scrape_value);
+        }
 
         GRAPH_LAST_SUCCESSFUL_REFRESH.set(chrono::Utc::now().timestamp() as i64);
 
--- a/graph-builder/src/graph.rs
+++ b/graph-builder/src/graph.rs
@@ -23,11 +23,15 @@ use failure::{Error, Fallible};
 use futures::Future;
 use lazy_static;
 pub use parking_lot::RwLock;
-use prometheus::{self, Counter, IntGauge};
+use prometheus::{self, histogram_opts, Counter, Gauge, Histogram, IntGauge};
 use serde_json;
 use std::collections::{HashMap, HashSet};
 use std::sync::Arc;
 use std::thread;
+
+/// Custom bucket values for upstream scraping duration in seconds
+pub const UPSTREAM_SCRAPE_BUCKETS: &[f64] =
+    &[1.0, 2.0, 3.0, 4.0, 5.0, 6.0, 7.0, 8.0, 9.0, 10.0, 11.0];
 
 lazy_static! {
     static ref GRAPH_FINAL_RELEASES: IntGauge = IntGauge::new(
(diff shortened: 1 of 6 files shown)


lucab:
Two high-level things that may be relevant here:
 * the first scrape loop is always an outlier (it takes longer as it needs to prime a cache from a cold-start). Perhaps we should exclude it or track separately (e.g. a one-shot start-to-ready duration gauge)
 * not sure if we need to split positive and negative (i.e. erroring) scrapes. I'm inclined to say no, but I am not sure if I am missing a usecase for that.

> TODO: a setting to adjust buckets?

I wouldn't explode the amount of configuration knobs we expose, especially not to tweak metrics. However, we can totally adjust buckets after observing a baseline for a few weeks.

vrutkovs:
>Perhaps we should exclude it or track separately (e.g. a one-shot start-to-ready duration gauge)

Excluded from `upstream_scrapes_duration_bucket`, its being set in `initial_upstream_scrape_duration` now

>However, we can totally adjust buckets after observing a baseline for a few weeks.

Agree, started with hardcoded 1.0 - 11.0 buckets

graph-builder/src/graph.rs
```diff
@@ -23,12 +23,15 @@ use failure::{Error, Fallible};
...
 use lazy_static;
 pub use parking_lot::RwLock;
-use prometheus::{self, Counter, IntGauge};
+use prometheus::{self, histogram_opts, Counter, Gauge, Histogram, IntGauge};
 use serde_json;
 use std::collections::{HashMap, HashSet};
 use std::sync::Arc;
 use std::thread;
 
+pub const UPSTREAM_SCRAPE_BUCKETS: &[f64; 11] =
```
steveej: I remember to have asked for a docstring on this public constant but it must've been swallowed. Please add a docstring here with a description of the values and their use.

steveej (changes requested):

graph-builder/src/graph.rs
vrutkovs: Oh, right, I added one in #154 only. Fixed

graph-builder/src/graph.rs
```diff
@@ -190,6 +209,13 @@ pub fn run<'a>(settings: &'a config::AppSettings, state: &State) -> ! {
             &settings.manifestref_key,
         );
         UPSTREAM_SCRAPES.inc();
+        // First iteration takes unusually slower, so it needs to be dropped
```
steveej:
```suggestion
        // First iteration is slower due to a cold cache, so it needs to be dropped
```

graph-builder/src/graph.rs
```diff
@@ -23,12 +23,16 @@ use failure::{Error, Fallible};
...
-use prometheus::{self, Counter, IntGauge};
+use prometheus::{self, histogram_opts, Counter, Gauge, Histogram, IntGauge};
 use serde_json;
 use std::collections::{HashMap, HashSet};
 use std::sync::Arc;
 use std::thread;
 
+/// Custom bucket values for upstream scraping duration in seconds
+pub const UPSTREAM_SCRAPE_BUCKETS: &[f64; 11] =
+    &[1.0, 2.0, 3.0, 4.0, 5.0, 6.0, 7.0, 8.0, 9.0, 10.0, 11.0];
```
steveej: Do you mind explaining how you chose these values and why it is eleven of them rather than some other arbitrary value?

graph-builder/src/graph.rs
```diff
@@ -23,12 +23,16 @@ use failure::{Error, Fallible};
...
 pub use parking_lot::RwLock;
-use prometheus::{self, Counter, IntGauge};
+use prometheus::{self, histogram_opts, Counter, Gauge, Histogram, IntGauge};
 use serde_json;
 use std::collections::{HashMap, HashSet};
 use std::sync::Arc;
 use std::thread;
 
+/// Custom bucket values for upstream scraping duration in seconds
+pub const UPSTREAM_SCRAPE_BUCKETS: &[f64; 11] =
```
steveej: I think you can drop the `; 11` here.

graph-builder/src/graph.rs
vrutkovs: Since this is entirely new metric all values (and the number of them) is arbitrary - I ran `curl --verbose --header 'Accept:application/json' http://localhost:9080/metrics` using test quay repo and values landed in 6-7 buckets usually, thus 11 values from 1.0 to 11.0

It may be entirely different on stage/prod deployments, so soon these would be updated

graph-builder/src/graph.rs
vrutkovs: Done

graph-builder/src/graph.rs
steveej: I'm starting to grasp it. Do we want the first scrape to land within one of these buckets too?

steveej: Self answer: no, because we don't even measure it.

graph-builder/src/graph.rs
```diff
@@ -190,6 +209,13 @@ pub fn run<'a>(settings: &'a config::AppSettings, state: &State) -> ! {
             &settings.manifestref_key,
         );
         UPSTREAM_SCRAPES.inc();
+        // First iteration is slower due to a cold cache, so it needs to be dropped
```
steveej: Nit: a newline here would be nice

graph-builder/src/graph.rs
```diff
@@ -169,6 +186,8 @@ pub fn run<'a>(settings: &'a config::AppSettings, state: &State) -> ! {
 
     // Don't wait on the first iteration
     let mut first_iteration = true;
+    // Don't collect upstream scrape duration on cold start
```
steveej: I'm wondering if we would prefer to wait until first_success with measuring the scrapes. An unsuccessful first scrape is unlikely to warm up the cache properly.

lucab: Yes, me too I'd prefer this logic to be linked to `first_success` instead (if possible).

vrutkovs: `first_success` is being updated way too late if I understood correctly - it happens after graph has been successfully built. 

Updated the code to store after scrape has been successful (`registries` are set, even if empty) - not entirely sure if that is correct though

lucab: I believe the suggestion was to stop the timer, keep the value locally, and then later set/observe in the correct metrics based on success status.

vrutkovs: Implemented decision making in `first_success`

lucab:
/hold
/lgtm

If @steveej has no further inputs, I think we can land this now.

graph-builder/src/graph.rs
```diff
@@ -55,6 +59,17 @@ lazy_static! {
...
     .unwrap();
+    static ref GRAPH_UPSTREAM_INITIAL_SCRAPE: Gauge = Gauge::new(
+        "graph_initial_upstream_scrape_duration",
+        "Duration of initial upstream scrape"
+    )
+    .unwrap();
+    static ref UPSTREAM_SCRAPES_DURATION: Histogram = Histogram::with_opts(histogram_opts!(
+        "graph_upstream_scrapes_duration",
+        "Upstream scrape duration in seconds",
+        UPSTREAM_SCRAPE_BUCKETS.to_vec()
```
steveej: Same comment as in #154: I'd inline the value here. If you decide to keep it separate, then there's a small naming nit where we're using the plural "scrapes" on other occurrences.

steveej (approved):
Just a style/naming nitpick, otherwise this

/lgtm
/hold

Please feel free to cancel the hold if you prefer to keep it this way.

graph-builder/src/graph.rs
vrutkovs: Fixed

PR #153, opened 2019-09-04
code-of-conduct.md: Add CoC

### LINKED ISSUE #144: Missing code of conduct file.

ashcrow:
Based off https://github.com/openshift/cluster-monitoring-operator/blob/master/code-of-conduct.md

Resolved #144 

/cc @lucab @smarterclayton

### DIFF
--- a/code-of-conduct.md
+++ b/code-of-conduct.md
@@ -1,0 +1,30 @@
+## cincinnati Community Code of Conduct
+
+### Contributor Code of Conduct
+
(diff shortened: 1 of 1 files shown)


lucab:
I think there is a need for a proper CoC covering OpenShift (+COS) projects.

Major problems with the one in this PR:
 * it is missing any contact points
 * it is missing a designed (and trained) responsible to mediate issues
 * it reads like a quick sed-replace of the original CoreOS one
 * there is no such a thing like "cincinnati events" and "cincinnati entity"

Instead, we'd like to have an umbrella CoC and just point to it, like https://github.com/prometheus/prometheus/blob/master/code-of-conduct.md.

PR #154, opened 2019-09-04
policy-engine: set custom buckets for v1_graph_serve_duration_seconds

vrutkovs:
Set custom buckets based on metrics from August

### DIFF
--- a/policy-engine/src/graph.rs
+++ b/policy-engine/src/graph.rs
@@ -7,9 +7,10 @@ use cincinnati::CONTENT_TYPE;
 use commons::{self, GraphError};
 use failure::Fallible;
 use futures::{future, Future};
-use prometheus::{Counter, Histogram, HistogramOpts, Registry};
+use prometheus::{histogram_opts, Counter, Histogram, Registry};
 use serde_json;
 use std::collections::HashMap;
+
 
 lazy_static! {
     static ref V1_GRAPH_INCOMING_REQS: Counter = Counter::new(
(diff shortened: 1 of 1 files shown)


policy-engine/src/graph.rs
```diff
@@ -7,20 +7,27 @@ use cincinnati::CONTENT_TYPE;
 use commons::{self, GraphError};
 use failure::Fallible;
 use futures::{future, Future};
-use prometheus::{Counter, Histogram, HistogramOpts, Registry};
+use prometheus::{Counter, Histogram, histogram_opts, Registry};
 use serde_json;
 use std::collections::HashMap;
 
+pub const GRAPH_SERVE_BUCKETS: &[f64; 11] = &[
```
steveej: A docstring with a rationale and a short explanation of the values would be nice.

steveej (changes requested):

policy-engine/src/graph.rs
vrutkovs: Done, PTAL

steveej: Thanks! I was hoping for a description on the values, but I realize that I simply have to read the prometheus docs for their meaning.

policy-engine/src/graph.rs
```diff
@@ -7,19 +7,25 @@ use cincinnati::CONTENT_TYPE;
...
         "v1_graph_incoming_requests_total",
         "Total number of incoming HTTP client request to /v1/graph"
     )
     .unwrap();
-    static ref V1_GRAPH_SERVE_HIST: Histogram = Histogram::with_opts(HistogramOpts::new(
+    static ref V1_GRAPH_SERVE_HIST: Histogram = Histogram::with_opts(histogram_opts!(
         "v1_graph_serve_duration_seconds",
-        "HTTP graph serving latency in seconds"
+        "HTTP graph serving latency in seconds",
+        GRAPH_SERVE_BUCKETS.to_vec()
```
steveej: If this is the only place where `GRAPH_SERVE_BUCKETS` is consumed I'd inline the value.

steveej (approved):
In addition to the nitpick in the comments, there's also the commit message which should say `policy-engine: ...`. The term _graph_engine_ isn't used anywhere AFAIK.

steveej (changes requested):
Accidentally approved. Please see the comments from the previous review

policy-engine/src/graph.rs
vrutkovs: Done, it makes comment a bit ugly though

vrutkovs:
```
+ rmdir /srv/volume/jenkins/workspace/openshift-cincinnati-gh-pr-check@2/dist/..//release-20190913.113637
rmdir: failed to remove '/srv/volume/jenkins/workspace/openshift-cincinnati-gh-pr-check@2/dist/..//release-20190913.113637': No such file or directory
Build step 'Execute shell' marked build as failure
```
/retest

PR #157, opened 2019-09-05
Dockerfile: use rustup and install 3 Rust versions

vrutkovs:
Fetch rust installer from `rustup.rs` and install 3 Rust versions (stable + 2 previous versions) so that tests would pass on several previous versions (would be added later)

TODO:
* [x] Sort out why CI is using previous container as a base 
  https://github.com/openshift/release/pull/4941 fixes that

### DIFF
--- a/dist/Dockerfile.build/Dockerfile
+++ b/dist/Dockerfile.build/Dockerfile
@@ -8,17 +8,13 @@ RUN yum -y groupinstall 'Development Tools'
 RUN yum -y install openssl-devel
 
 ENV HOME="/root"
-
-# build: Rust stable toolchain
-RUN \
-  mkdir $HOME/rust && \
-  curl https://static.rust-lang.org/dist/rust-1.37.0-x86_64-unknown-linux-gnu.tar.gz | \
-  tar -xzvf - -C $HOME/rust --strip 1 && \
-  $HOME/rust/install.sh; \
-  rm -rf $HOME/rust
-
-# add cargo binaries to PATH
 ENV PATH="${HOME}/.cargo/bin:${PATH}"
+
+# build: Rust stable toolchain + 2 previous versions
+RUN curl https://sh.rustup.rs -sSf | sh -s -- --default-toolchain 1.37.0 -y && \
+    rustup install 1.36.0 && \
+    rustup install 1.35.0 && \
+    rustup default 1.37.0
 
 # test: linters
 RUN yum -y install yamllint


dist/Dockerfile.build/Dockerfile
```diff
@@ -8,18 +8,15 @@ RUN yum -y groupinstall 'Development Tools'
...
-
-# add cargo binaries to PATH
 ENV PATH="${HOME}/.cargo/bin:${PATH}"
 
+# build: Rust stable toolchain + 2 previous versions
+RUN mkdir $HOME/rust && \
+    curl https://sh.rustup.rs -sSf | sh -s -- --default-toolchain 1.37.0 -y && \
+    rustup install 1.36.0 && \
+    rustup install 1.35.0 && \
+    rustup default 1.37.0
```
steveej, line 17: do we need this even though `--default-toolchain` is defined above with this version? If this is required, I would favor to remove the version from the installation command.

steveej (commented):
Two minor comments (actually the same but duplicate). I'd like to discuss if we can get rid of one of the Dockerfiles. I remember to have kept the symlink for the time of transitioning to a common path.

dist/Dockerfile.build/Dockerfile
vrutkovs: I left `--default-toolchain 1.37.0` there so that whenever stable changes we'd still get 1.37 installed - and we'd have to update to new stable with a PR.

It can be removed if we commit to supporting "rolling stable" though

steveej: Ideally we would install the most recent stable, parse the version value and also install and test against the previous 2 or 3 versions. A solution which doesn't require any hardcoding, neither in the install section nor in the tests, would be awesome.

vrutkovs: Ah, got it, sure lets see if that can be nicely done

dist/Dockerfile.build/Dockerfile
```diff
@@ -8,18 +8,16 @@ RUN yum -y groupinstall 'Development Tools'
...
-  $HOME/rust/install.sh; \
-  rm -rf $HOME/rust
-
-# add cargo binaries to PATH
 ENV PATH="${HOME}/.cargo/bin:${PATH}"
 
+# build: Rust stable toolchain + 2 previous versions
+RUN mkdir $HOME/rust && \
+    curl https://sh.rustup.rs -sSf | sh -s -- -y && \
+    MINOR="$(rustc --version | grep -Po 'rustc 1.\K([\d]+?).')" && \
```
steveej: Fingers crossed Rust never releases version 2 or uses patch releases :smile:

steveej (commented):
I think it is good enough for now. It would be possible to do something like

```
$ curl https://raw.githubusercontent.com/rust-lang/rust/master/RELEASES.md | grep '^Version' | head -n3 | awk '{ print $2 }' | xargs -L1 echo about to install
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  399k  100  399k    0     0   675k      0 --:--:-- --:--:-- --:--:--  675k
about to install 1.37.0
about to install 1.36.0
about to install 1.35.0
```

but I think it's not worth the hassle of adding the necessary special cases for patch releases which might be in between. This is definitely better than what we have now, and we can detect the available toolchains in the tests using `rustup toolchain list`.

/lgtm
/hold

I'll leave the final call to other reviewers or to yourself.

steveeJ:
/lgtm case

As discussed in the weekly sync we'd like to install a fixed baseline version and also install the previous two versions.

steveej:
/lgtm cancel

As discussed in the weekly sync we'd like to install a fixed baseline version and also install the previous two versions.

vrutkovs:
Discussed on the meeting today: it would be confusing when the next Rust version goes stable, it gets pulled in during the build and if the tests break it would be unclear which change caused this. 

Instead of parsing the version we should set versions explicitly - also in order to have images reproducible on every commit

vrutkovs:
/hold cancel

@steveeJ ready for review

dist/Dockerfile.build/Dockerfile
```diff
@@ -8,18 +8,15 @@ RUN yum -y groupinstall 'Development Tools'
...
-  curl https://static.rust-lang.org/dist/rust-1.37.0-x86_64-unknown-linux-gnu.tar.gz | \
-  tar -xzvf - -C $HOME/rust --strip 1 && \
-  $HOME/rust/install.sh; \
-  rm -rf $HOME/rust
-
-# add cargo binaries to PATH
 ENV PATH="${HOME}/.cargo/bin:${PATH}"
 
+# build: Rust stable toolchain + 2 previous versions
+RUN mkdir $HOME/rust && \
```
steveej: This `mkdir` is leftover and not required for rustup. This made me wonder if `rustup` cleans up after itself, or if the installation files end up in the image.

steveej (changes requested):
I hadn't caught this leftover before. sorry!

dist/Dockerfile.build/Dockerfile
vrutkovs: Good catch, fixed

steveej: Thanks. I checked the rustup script and it does clean up after itself so we're good.

steveej (approved):
Thanks!

/lgtm

PR #158, opened 2019-09-05
graph-builder/registry: require the manifestref

steveej:
This change enforces the presence of the manifest reference in the
response from the upstream registry. The motivation is to always deliver
a predictable payload URL in the graph.

Ref: https://bugzilla.redhat.com/show_bug.cgi?id=1686589


lucab:
/hold
/lgtm

I think this is in order to close https://bugzilla.redhat.com/show_bug.cgi?id=1686589, right? If so, can you add a reference to the commit/PR message?

steveej:
> If so, can you add a reference to the commit/PR message?

Done, thanks!

/hold cancel

PR #161, opened 2019-09-24
openshift: enforce CPU limits on each container

lucab:
This introduces CPU limit settings on both graph-builder and
policy-engine, to avoid disrupting all cluster resources in case
of issues.
The default value (750 millicores) has been picked as 2x the maximum
value observed so far in production. It can be tweaked later on based
on further observations.
It also adds baseline resources request for pod scheduling.

Closes: https://jira.coreos.com/browse/APPSRE-308

### DIFF
--- a/dist/openshift/cincinnati.yaml
+++ b/dist/openshift/cincinnati.yaml
@@ -90,7 +90,11 @@ objects:
                 timeoutSeconds: 3
               resources:
                 limits:
-                  memory: ${MEMORY_LIMIT}
+                  cpu: ${GB_CPU_LIMIT}
+                  memory: ${GB_MEMORY_LIMIT}
+                requests:
+                  cpu: ${GB_CPU_REQUEST}
+                  memory: ${GB_MEMORY_REQUEST}
               volumeMounts:
                 - name: secrets
                   mountPath: /etc/secrets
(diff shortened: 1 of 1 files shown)


dist/openshift/cincinnati.yaml
```diff
@@ -90,7 +90,11 @@ objects:
                 timeoutSeconds: 3
               resources:
                 limits:
+                  cpu: ${CPU_LIMIT}
                   memory: ${MEMORY_LIMIT}
+                requests:
+                  cpu: "350m"
```
aditya-konarde: Why do we have parameterized limits but hardcoded requests? Would it make sense to have parameters for both instead? 

I would also suggest we don't lock ourselves to always having the same CPU and Memory requirements on Policy Engine and Graph builder, so you can perhaps split this as: 
`PE_CPU_LIMIT` `GB_CPU_LIMIT` and so on...

aditya-konarde (changes requested):

dist/openshift/cincinnati.yaml
lucab: So the suggestion would introduce the following parameters:
 * `GB_CPU_LIMIT`
 * `GB_MEMORY_LIMIT`
 * `PE_CPU_REQUEST`
 * `PE_MEMORY_REQUEST`

correct?

Alternatively, what about going the other direction and hardcoding everything? We are anyway doing devops-style and keeping all parameters in this file, so we may as well remove one level of indirection, I guess? Or do we use/rely on these being parameters, somewhere else?

aditya-konarde: Yes, and we can set the default parameter values in the file (in case someone wants to not use it) 

And then we override them here: https://gitlab.cee.redhat.com/service/saas-cincinnati/blob/master/cincinnati-services/cincinnati.yaml

Like so: https://gitlab.cee.redhat.com/service/saas-telemeter/blob/master/telemeter-services/observatorium.yaml

lucab:
@aditya-konarde amended with your suggestions, PTAL.

aditya-konarde (approved):
/lgtm

PR #163, opened 2019-09-27
Add cincinnati_*_build_info metric with git commit label

vrutkovs:
This adds a new metric with git commit label which designates that graph-builder/policy-engine has started. Git commit label tracking would be useful during pod restarts / deploys.

xref https://jira.coreos.com/browse/CIN-73

### DIFF
--- a/graph-builder/src/graph.rs
+++ b/graph-builder/src/graph.rs
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use crate::built_info;
 use crate::config;
 use crate::registry::{self, Registry};
 use actix_web::{HttpRequest, HttpResponse};
@@ -23,7 +24,7 @@ use failure::{Error, Fallible};
 use futures::Future;
 use lazy_static;
 pub use parking_lot::RwLock;
-use prometheus::{self, histogram_opts, Counter, Gauge, Histogram, IntGauge};
+use prometheus::{self, histogram_opts, labels, opts, Counter, Gauge, Histogram, IntGauge};
 use serde_json;
 use std::collections::{HashMap, HashSet};
 use std::sync::Arc;
@@ -72,6 +73,12 @@ lazy_static! {
         "Total number of incoming HTTP client request to /v1/graph"
     )
     .unwrap();
+    static ref BUILD_INFO: Counter = Counter::with_opts(opts!(
+        "build_info",
+        "Build information",
+        labels!{"git_commit" => built_info::GIT_VERSION.unwrap(),}
+    ))
+    .unwrap();
 }
 
 /// Register relevant metrics to a prometheus registry.
@@ -85,6 +92,7 @@ pub fn register_metrics(registry: &prometheus::Registry) -> Fallible<()> {
     registry.register(Box::new(GRAPH_UPSTREAM_INITIAL_SCRAPE.clone()))?;
     registry.register(Box::new(UPSTREAM_SCRAPES_DURATION.clone()))?;
     registry.register(Box::new(V1_GRAPH_INCOMING_REQS.clone()))?;
+    registry.register(Box::new(BUILD_INFO.clone()))?;
     Ok(())
 }
 
@@ -184,6 +192,8 @@ pub fn run<'a>(settings: &'a config::AppSettings, state: &State) -> ! {
     // Don't wait on the first iteration
     let mut first_iteration = true;
     let mut first_success = true;
+
+    BUILD_INFO.inc();
 
     loop {
         // Store scrape duration value. It would be used for initial scrape gauge or scrape histogram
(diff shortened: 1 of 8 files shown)


graph-builder/src/graph.rs
```diff
@@ -72,6 +77,12 @@ lazy_static! {
         "Total number of incoming HTTP client request to /v1/graph"
     )
     .unwrap();
+    static ref UP: Counter = Counter::with_opts(opts!(
+        "up",
```
lucab: I have some concern that this name is misleading. `up` is a special metrics which is not controlled by the producer and has its well defined semantics: https://prometheus.io/docs/concepts/jobs_instances/

This is an info-metrics recording built-time details. The convention I've seen in other projects (e.g. Prometheus itself) is to name these `<APP>_build_info{...}`.

lucab (dismissed):

graph-builder/src/graph.rs
vrutkovs: Ah, good idea. Any other info except the commit hash we'd like to expose?

wking: Is there a way to get a staging/prod distinction into this metric?  I can't think of one, but having some kind of deployment-group-name would be nice.  Maybe Prometheus will layer that on when slurping.

lucab: The staging-vs-prod information is already injected, even in a redundant way, by the infrastructure (they belong to different namespaces and datasources, and we are using the former for querying pruposes).

vrutkovs: The only info we can get there is being set during the build (see https://docs.rs/built/0.3.2/built/), I don't think we can distinguish stage and prod here as these are built from the single source

lucab: @vrutkovs I had a quick look of what is exposed by `built`, and I didn't see anything else that would be immediately useful for our monitoring purposes.

lucab:
@vrutkovs cool, just a minor placing&naming comment from my side.
FYI we usually split (TOML + lockfile + code) changes in different commits. That makes it easier to rebase PRs that are progressing in parallel. Not a strict requirement here, but if you are rebasing once more you could as well split them (see master for prior examples).

graph-builder/src/graph.rs
```diff
@@ -23,12 +23,17 @@ use failure::{Error, Fallible};
...
 pub use parking_lot::RwLock;
-use prometheus::{self, histogram_opts, Counter, Gauge, Histogram, IntGauge};
+use prometheus::{self, histogram_opts, labels, opts, Counter, Gauge, Histogram, IntGauge};
 use serde_json;
 use std::collections::{HashMap, HashSet};
 use std::sync::Arc;
 use std::thread;
 
+/// Build info
+pub mod built_info {
```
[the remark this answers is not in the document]
vrutkovs: >I think this does not belong in this module, as it isn't a property of the graph endpoint logic.

Right, moved it to `lib.rs`

>Additionally, I think you can align the module name with the info-metrics name, and uniformly call both build_info.

Both of these are prefixed, so metrics endpoints would have `cincinnati_gb_build_info` and `cincinnati_pe_build_info`

graph-builder/src/lib.rs
```diff
@@ -36,3 +36,8 @@ pub mod graph;
 pub mod registry;
 pub mod release;
 pub mod status;
+
+/// Build info
+pub mod built_info {
```
steveej: are you planning to use this from somewhere else? Otherwise it doesn't need to be public.

policy-engine/src/main.rs
```diff
@@ -96,6 +111,7 @@ fn main() -> Result<(), Error> {
     })
     .bind((settings.address, settings.port))?
     .start();
+    BUILD_INFO.inc();
```
steveej, line 115: I'm curious, do you have a plan to use the counter with the build info for something?

graph-builder/src/lib.rs
steveej: It might even be possible to put this into the `commons` crate and use it from the GB and PE. Have you tried that?

vrutkovs: Removed `pub` (also fixed warnings during build as some const are not being used).

>It might even be possible to put this into the commons crate and use it from the GB and PE. Have you tried that?

I'm not sure if that would work. GB and PE could be built and run independently, not sure what would build_info info from a common crate would show. I'd rather keep these separate for now - and attempt to merge it later

policy-engine/src/main.rs
vrutkovs: No, all we need is labels from this metric. `inc` here would be executed just once

steveej: I see, so this is very similar to the `up` metric just with custom labels attached.

graph-builder/src/lib.rs
steveej: I'm fine postponing this

steveej:
/lgtm

Thanks!

PR #165, opened 2019-10-02
graph-builder/src/graph: update bucket values for graph_upstream_scrapes_duration

vrutkovs:
Based on stage data buckets should be adjusted:

* Set minimum to 5 seconds (the actual observed minimum is 7, perhaps this can be improved later on)
* Add 15.0, 20.0 and 30.0

/cc @lucab @steveeJ


steveej:
/lgtm

Thanks!

steveej (approved):

PR #166, opened 2019-10-02
Add prow_rustfmt helper to run rustfmt

vrutkovs:
Adds a handy script to install rustfmt for selected rust version (latest if unset). See https://github.com/openshift/release/pull/4936#issuecomment-537394744

/cc @lucab @steveeJ

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,8 +86,9 @@ The *dist/* directory contains various CI/CD related files.
 #### Openshift Dev
 * Uses *dist/Dockerfile.build/Dockerfile* as the build container image
 * Run the following scripts on PR
-    * `dist/cargo_test.sh`
     * `dist/prow_yaml_lint.sh`
+    * `dist/prow_rustfmt.sh`
+    * `dist/cargo_test.sh`
 
 For details please see [github.com/openshift/release/(...)/openshift-cincinnati-master.yaml][1].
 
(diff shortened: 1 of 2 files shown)


dist/prow_rustfmt.sh
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
+
+set -e
+
+# Install and run cargofmt. Input: rust version
+test -z ${1} || rustup default ${1}
```
steveej: I suggest to use the `--toolchain` argument to `rustup` instead of setting a new default. That way it can also be used on developer machines.

README.md
```diff
@@ -86,8 +86,9 @@ The *dist/* directory contains various CI/CD related files.
 #### Openshift Dev
 * Uses *dist/Dockerfile.build/Dockerfile* as the build container image
 * Run the following scripts on PR
-    * `dist/cargo_test.sh`
     * `dist/prow_yaml_lint.sh`
+    * `dist/prow_rustfmt.sh`
+    * `dist/cargo_test.sh`
```
steveej, line 91: Out of curiosity, why the reorder?

steveej (changes requested):

README.md
vrutkovs: Probably its not needed, but I'd prefer to run static checks before unit tests, so that obvious errors would be caught first

dist/prow_rustfmt.sh
vrutkovs: Good idea, implemented

steveej (approved):
/lgtm

thanks!

PR #167, opened 2019-10-02
cincinnati/src/plugins/internal/channel_filter: Expose all channels

wking:
This filter landed in cfa652e251 (#79), with filtering the metadata as an explicit goal:

> If a release belongs to several channels the other channel names are removed from the result.

But the channel-switching procedure is:

1. Upgrade to a release that belongs to both your current channel (e.g. `prerelease-4.1`) and the target channel (e.g. `stable-4.1`).
2. Change your channel.
3. Upgrade in the new channel.

This commit leaves the channel metadata alone, so now the channel filter is only removing releases that don't belong to the channel and is no longer adjusting metadata for releases that do belong to the channel.  The open list won't work with secret channels, but we don't have any short-/mid-term plans for secret channels, and we have very short-term plans for 4.1->4.2 edges.  Tooling like the web console's channel-picker can use the new approach to retrieve available channels for the current release, instead of hard-coding a list of channel choices that really has nothing to do with the current release (e.g. openshift/console#1498).

### DIFF
--- a/cincinnati/src/plugins/internal/channel_filter.rs
+++ b/cincinnati/src/plugins/internal/channel_filter.rs
@@ -80,12 +80,7 @@ impl InternalPlugin for ChannelFilterPlugin {
                                 .metadata
                                 .get_mut(&format!("{}.{}", key_prefix, key_suffix))
                                 .map_or(true, |values| {
-                                    if values.split(',').any(|value| value.trim() == channel) {
-                                        *values = channel.clone();
-                                        false
-                                    } else {
-                                        true
-                                    }
+                                    !values.split(',').any(|value| value.trim() == channel)
                                 }),
                             // remove if it's not a ConcreteRelease
                             _ => true,
@@ -257,7 +252,7 @@ mod tests {
                             0,
                             [(
                                 format!("{}.{}", &key_prefix, &key_suffix),
-                                String::from("a"),
+                                String::from("a, c"),
                             )]
                             .iter()
                             .cloned()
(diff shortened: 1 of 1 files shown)


steveej:
> This commit leaves the channel metadata alone, so now the channel filter is only removing releases that don't belong to the channel and is no longer adjusting metadata for releases that do belong to the channel.

This makes sense in terms of the release procedure.

>The open list won't work with secret channels, but we don't have any short-/mid-term plans for secret channels, and we have very short-term plans for 4.1->4.2 edges.

I'm also not concerned about exposing this in Cincinnati, as the scraped metadata is public anyway.

> Tooling like the web console's channel-picker can use the new approach to retrieve available channels for the current release, instead of hard-coding a list of channel choices that really has nothing to do with the current release (e.g. [openshift/console#1498](https://github.com/openshift/console/pull/1498)).

Yay, that'll be a UX and developer improvement at the same time! Slightly off-topic in the context of Cincinnati but maybe there needs to be a change to the CVO too. It [currently stores a single channel string per available update](https://github.com/openshift/cluster-version-operator/blob/0725bd53c732c8e3a490d1080f03edbe7e579b3a/pkg/cvo/availableupdates.go#L71-L79).

/lgtm

Thanks for this!

Retrying the devshift test:
[test]

steveej (approved):

wking:
> ... but maybe there needs to be a change to the CVO too. It currently stores a single channel string per available update.

Yeah, we probably need that to feed in-cluster consumers like the console.

PR #168, opened 2019-10-03
Add scripts for e2e tests

vrutkovs:
This commit adds two helper scripts:

* Fetch channels data from the instance and store it in a file
* Compare two files using `jq` to ensure the deploy didn't change the output

/cc @lucab @steveeJ


hack/e2e_collect.sh
```diff
@@ -0,0 +1,20 @@
...
+
+type -f curl > /dev/null || {
+  echo Please install curl
+  exit 1
+}
+
+test -z "${1}" && echo "Usage: ./e2e-collect.sh CINCINNATI_URL <tempfile prefix>" && exit 1
+
+TMPFILETEMPLATE="tmp.${2}XXXX.json"
+TMPFILE="$(mktemp -u --tmpdir=. -u ${TMPFILETEMPLATE})"
```
steveej: Why the usage of a template and the CWD?

hack/e2e_compare.sh
```diff
@@ -0,0 +1,12 @@
...
+set -e
+
+# Compare 2 files generated by e2e_collect. This ensures elements are sorted by version
+
+type -f jq > /dev/null || {
+  echo Please install jq
+  exit 1
+}
+
+diff <(jq -S '.nodes |= sort_by(.version)' ${1}) <(jq -S '.nodes |= sort_by(.version)' ${1}) || exit 0
```
steveej: Would you mind adding a comment to describe what the expected is for the `jq` script?

hack/e2e_collect.sh
vrutkovs: Might be useful for local testing - e.g. prefix stage and prod outputs separately

steveej: Wouldn't that be as easy as assigning the output of this script to a variable in the outer scope? e.g.

```shell
GRAPH_STAGE=$(./e2e_collect.sh https://api.openshift.com:443/api/upgrades_info/v1/graph?channel='prerelease-4.1'\&arch='amd64')
GRAPH_PROD=$(./e2e_collect.sh https://api.stage.openshift.com:443/api/upgrades_info/v1/graph?channel='prerelease-4.1'\&arch='amd64')
```

vrutkovs: Initially I didn't want to output to stdout for comparison (as graphs may be large), but it seems its more intuitive.

Note that `GRAPH_STAGE` and `GRAPH_PROD` still have to saved into a file for comparison via `e2e_compare`, but in that case the user gets to choose the file names

hack/e2e_compare.sh
vrutkovs: Done

hack/e2e_collect.sh
steveej: > Initially I didn't want to output to stdout for comparison

I'm not suggesting to do that. The variables would contain filenames in my suggestion.

steveej: > the user gets to choose the file names

I don't think we need that if the path is stored in a variable. I think we should clean them up afterwards though, so maybe we need a cohesive script which does the getting and the comparing and cleans up unconditionally.

vrutkovs: Ah, I see. I wanted to have a temp file template so that stage / prod results for several tests (e.g. multiarch / channels) would be easy to distinct. This is probably too complex, removed this part

hack/e2e_compare.sh
```diff
@@ -0,0 +1,14 @@
...
+# Compare 2 files generated by e2e_collect. This ensures elements are sorted by version
+
+type -f jq > /dev/null || {
+  echo Please install jq
+  exit 1
+}
+
+# jq would output the whole graph with elements in "nodes" sorted by "version"
+# to ensure the json output is identical for the equal graphs
+diff <(jq -S '.nodes |= sort_by(.version)' ${1}) <(jq -S '.nodes |= sort_by(.version)' ${1}) || exit 0
```
steveej: Is one of the `${1}` is supposed to be `${2}`?

hack/e2e_compare.sh
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env bash
+
+set -e
+
+# Compare 2 files generated by e2e_collect. This ensures elements are sorted by version
+
```
steveej, line 6: Please add a check to ensure the argument(s) are passed in.

steveej (changes requested):

hack/e2e_collect.sh
```diff
@@ -0,0 +1,19 @@
...
+# * 2 - prefix
+
+type -f curl > /dev/null || {
+  echo Please install curl
+  exit 1
+}
+
+test -z "${1}" && echo "Usage: ./e2e-collect.sh CINCINNATI_URL" && exit 1
+
+TMPFILE="$(mktemp -u --tmpdir=.)"
```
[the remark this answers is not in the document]
vrutkovs: Local dir would be the easiest place to find those. `tmp.*` files in `hack/` were added to gitignore so they won't be commited

hack/e2e_compare.sh
vrutkovs: doh! Fixed

hack/e2e_compare.sh
vrutkovs: Fixed

hack/e2e_collect.sh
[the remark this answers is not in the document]
steveej: When using `mktemp` we don't care about finding them, as we just want a transient file with a unique name. Since the script emits the path I don't see why it needs to live and be ignored in the workdir.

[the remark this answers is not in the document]
vrutkovs: Okay, using standard tmpdir now

hack/e2e_compare.sh
steveej: Where?

hack/e2e_compare.sh
```diff
@@ -0,0 +1,14 @@
...
+type -f jq > /dev/null || {
+  echo Please install jq
+  exit 1
+}
+
+# jq would output the whole graph with elements in "nodes" sorted by "version"
+# to ensure the json output is identical for the equal graphs
+diff <(jq -S '.nodes |= sort_by(.version)' ${1}) <(jq -S '.nodes |= sort_by(.version)' ${2}) || exit 0
```
steveej: I suggest not to mask the exit state here and leave this decision to the caller. If the caller wants to treat this as an error, it'd have to rely on the callee's stdout instead to see if there was a difference.

steveej (changes requested):

hack/e2e_compare.sh
vrutkovs: Sorry, forgot to push it. `test "$#" -ne 2 && echo "Usage: ./e2e_compare.sh FILE1 FILE2" && exit 1` would implement that check

hack/e2e_compare.sh
vrutkovs: Okay, makes sense.

Seems default diff exit codes would work here too - `Exit status is 0 if inputs are the same, 1 if different, 2 if trouble`

steveej (approved):
/lgtm

Thanks!

steveej:
[test]

PR #169, opened 2019-10-03
add app-sre as approvers

maorfr:
done as part of https://jira.coreos.com/browse/APPSRE-481

we do not plan to use these powers, but only in case of emergency.

### DIFF
--- a/OWNERS_ALIASES
+++ b/OWNERS_ALIASES
@@ -7,4 +7,15 @@ aliases:
     - lucab
     - wking
     - vrutkovs
+  app-sre:
+    - aditya-konarde
+    - jfchevrette
+    - jmelis
+    - jakedt
+    - jzelinskie
+    - maorfr
+    - mmclanerh
+    - pbergene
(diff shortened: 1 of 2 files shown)


OWNERS_ALIASES
```diff
@@ -7,4 +7,15 @@ aliases:
     - lucab
     - wking
     - vrutkovs
+  app-sre:
```
steveej, line 10: Would be nice if this could be defined in a common place and pulled in from there. [The OWNERS doc](https://github.com/kubernetes/community/blob/master/contributors/guide/owners.md) doesn't talk about such a feature though.

steveej (approved):
/lgtm

@maofr 
In case you want to work on customization for the OWNERS file to pull in external alias definitions:
/hold
Otherwise please feel free to cancel the hold.

maorfr:
I may write an external automation for this at some point.
In the meanwhile, thanks a lot!

/hold cancel

PR #172, opened 2019-10-17
Revert "Add scripts for e2e tests"

vrutkovs:
Reverts openshift/cincinnati#168

There is nothing with those, but these have to be duplicated internally for now - and we'll revert this PR to avoid duplication and confusion


steveej (approved):
/lgtm

PR #173, opened 2019-10-21
hack/graph: Port graph.sh for 🍎

lisa:
I was running into a Bash error with the script,

```
./hack/graph.sh: line 12: unexpected EOF while looking for matching `)'
```

This commit removes the subshell, its associated heredoc, as well as the escaped `$` and `\` to enable the script to run on Bash 3.2.57(1)-release (for MacOS 10.14).

### DIFF
--- a/hack/graph.sh
+++ b/hack/graph.sh
@@ -9,26 +9,25 @@
 #   curl -sH 'Accept:application/json' 'https://api.openshift.com/api/upgrades_info/v1/graph?channel=prerelease-4.1' | graph.sh | dot -Tsvg >graph.svg
 set -e
 
-JQ_SCRIPT="$(cat <<EOF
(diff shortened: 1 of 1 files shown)


steveej (commented):
Could we alternatively change the header to `#!/usr/bin/env bash`, and would that help for :apple:?

lisa:
> Could we alternatively change the header to `#!/usr/bin/env bash`, and would that help for 🍎?

@steveeJ I tried executing with `bash hack/graph.sh` and changed the shebang to `#!/bin/bash` (to force the issue) and the error still remained.

steveej (approved):
> I tried executing with `bash hack/graph.sh` and changed the shebang to `#!/bin/bash` (to force the issue) and the error still remained.

Alright, thanks for trying and fixing. 

/lgtm

PR #174, opened 2019-10-23
graph-builder/registry: concurrent tag processing

steveej:
This introduces concurrent processing of tags to potentially reduce
scrape time significantly. This causes the resulting release vector to
be ordered non-deterministically, which had to be considered in the
tests.

Other notable changes:
* Standardize log output throughout all operations when processing a tag
* Introduce graph_builder::registry::cache for convenience

---

TODO
* [x] Depends on https://github.com/camallo/dkregistry-rs/pull/120
* [x] Depends on #195 
* [x] Depends on #206 
* [x] Includes #207 
* [x] minimize critical section for cache locking
* [x] allow unordered processing of the tags
* [x] Make parallelization configurable?

### DIFF
--- a/dist/openshift/cincinnati.yaml
+++ b/dist/openshift/cincinnati.yaml
@@ -70,6 +70,7 @@ objects:
                 "--upstream.registry.url", "$(REGISTRY)",
                 "--upstream.registry.repository", "$(REPOSITORY)",
                 "--upstream.registry.credentials_path", "/etc/secrets/registry-credentials",
+                "--upstream.registry.fetch_concurrency", "${GB_UPSTREAM_FETCH_CONCURRENCY}",
               ]
               ports:
                 - name: graph-builder
@@ -289,3 +290,6 @@ parameters:
   - name: PE_BINARY
     value: /usr/bin/policy-engine
     displayName: Path to policy-engine binary
+  - name: GB_UPSTREAM_FETCH_CONCURRENCY
+    value: "16"
+    displayName: Concurrency when fetching the upstream metadata
--- a/graph-builder/src/config/options.rs
+++ b/graph-builder/src/config/options.rs
@@ -75,6 +75,10 @@ pub struct DockerRegistryOptions {
     /// Metadata key where to record the manifest-reference
     #[structopt(long = "upstream.registry.manifestref_key")]
     pub manifestref_key: Option<String>,
+
+    /// Concurrency for graph fetching
+    #[structopt(long = "upstream.registry.fetch_concurrency")]
+    pub fetch_concurrency: Option<usize>,
 }
 
 impl MergeOptions<Option<ServiceOptions>> for AppSettings {
@@ -109,6 +113,7 @@ impl MergeOptions<Option<DockerRegistryOptions>> for AppSettings {
             assign_if_some!(self.repository, registry.repository);
             assign_if_some!(self.credentials_path, registry.credentials_path);
             assign_if_some!(self.manifestref_key, registry.manifestref_key);
+            assign_if_some!(self.fetch_concurrency, registry.fetch_concurrency);
         }
         Ok(())
     }
--- a/graph-builder/src/graph.rs
+++ b/graph-builder/src/graph.rs
@@ -164,7 +164,7 @@ impl HasRegistry for State {
 }
 
 #[allow(clippy::useless_let_if_seq)]
-pub async fn run<'a>(settings: &'a config::AppSettings, state: &State) -> ! {
+pub async fn run(settings: &config::AppSettings, state: &State) -> ! {
     // Grow-only cache, mapping tag (hashed layers) to optional release metadata.
     let mut cache = HashMap::new();
 
--- a/graph-builder/src/graph.rs
+++ b/graph-builder/src/graph.rs
@@ -25,7 +25,7 @@ use lazy_static;
 pub use parking_lot::RwLock;
 use prometheus::{self, histogram_opts, labels, opts, Counter, Gauge, Histogram, IntGauge};
 use serde_json;
-use std::collections::{HashMap, HashSet};
+use std::collections::HashSet;
 use std::sync::Arc;
 use std::thread;
 
@@ -166,7 +166,7 @@ impl HasRegistry for State {
 #[allow(clippy::useless_let_if_seq)]
 pub async fn run(settings: &config::AppSettings, state: &State) -> ! {
     // Grow-only cache, mapping tag (hashed layers) to optional release metadata.
-    let mut cache = HashMap::new();
+    let mut cache = registry::cache::new();
 
     let registry = Registry::try_from_str(&settings.registry)
         .unwrap_or_else(|_| panic!("failed to parse '{}' as Url", &settings.registry));
--- a/graph-builder/src/graph.rs
+++ b/graph-builder/src/graph.rs
@@ -210,6 +210,7 @@ pub async fn run(settings: &config::AppSettings, state: &State) -> ! {
             password.as_ref().map(String::as_ref),
             &mut cache,
             &settings.manifestref_key,
+            settings.fetch_concurrency,
         )
         .await;
         UPSTREAM_SCRAPES.inc();
--- a/graph-builder/src/registry.rs
+++ b/graph-builder/src/registry.rs
@@ -195,6 +195,7 @@ pub async fn fetch_releases(
     password: Option<&str>,
     cache: &mut crate::registry::cache::Cache,
     manifestref_key: &str,
+    concurrency: usize,
 ) -> Result<Vec<Release>, Error> {
     let authenticated_client = dkregistry::v2::Client::configure()
         .registry(&registry.host_port_string())
@@ -218,7 +219,7 @@ pub async fn fetch_releases(
         Arc::new(FuturesMutex::new(Vec::with_capacity(estimated_releases)))
     };
 
-    tags.try_for_each_concurrent(8, |tag| {
+    tags.try_for_each_concurrent(concurrency, |tag| {
         let authenticated_client = authenticated_client.clone();
         let cache = cache.clone();
         let releases = releases.clone();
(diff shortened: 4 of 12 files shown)


lucab:
> This involved switching from tokio's current_thread::Runtime to its multi-threaded runtime

I'm not sure I follow the reasoning behind this. Why was concurrency (i.e. buffering futures) not enough and you had to introduce parallelism at the runtime level?

steveej:
> I'm not sure I follow the reasoning behind this. Why was concurrency (i.e. buffering futures) not enough and you had to introduce parallelism at the runtime level?

I didn't have to switch the runtime, but I wanted to. It enables true parallelism which is more than asynchronous I/O.

steveej:
[test]

steveej:
@vrutkovs I think something's broken with the CI setup here.

/retest

vrutkovs:
```
Could not retrieve mirrorlist http://mirrorlist.centos.org/?release=7&arch=x86_64&repo=os&infra=container error was
12: Timeout on http://mirrorlist.centos.org/?release=7&arc... Less than 1000 bytes/sec transferred the last 30 seconds')
```
in images

/retest

steveej:
Another network related failure it seems.

/retest

graph-builder/src/registry.rs
```diff
@@ -380,8 +408,9 @@ async fn find_first_release(
 ) -> Fallible<(String, Option<Release>)> {
     let tag = repo_tag.clone();
 
+    // TODO(steveeJ): concurrently process the layer digests
```
LalatenduMohanty: Can we add a Jira card for this and remove the TODO here?

LalatenduMohanty (changes requested):

graph-builder/src/config/options.rs
```diff
@@ -75,6 +75,10 @@ pub struct DockerRegistryOptions {
     /// Metadata key where to record the manifest-reference
     #[structopt(long = "upstream.registry.manifestref_key")]
     pub manifestref_key: Option<String>,
+
+    /// Concurrency for graph fetching
+    #[structopt(long = "upstream.registry.fetch_concurrency")]
```
vrutkovs, line 80: Lets update a configmap in `dist/openshift/cincinnati.yaml` with the default value so that it would not be hidden in the source code

graph-builder/src/graph.rs
```diff
@@ -164,9 +164,9 @@ impl HasRegistry for State {
 }
 
 #[allow(clippy::useless_let_if_seq)]
-pub async fn run<'a>(settings: &'a config::AppSettings, state: &State) -> ! {
```
vrutkovs, line 167: Not sure why lifetime is changed here

graph-builder/src/registry.rs
```diff
@@ -42,6 +43,21 @@ impl Into<cincinnati::Release> for Release {
...
+    use crate::registry::Release;
+    use futures_locks_pre::RwLock as FuturesRwLock;
+    use std::collections::HashMap;
+
+    /// The cache type to hold the release cache
+    pub type Cache = FuturesRwLock<HashMap<u64, Option<Release>>>;
+
+    /// Instantiates a new release cache
+    pub fn new() -> Cache {
+        FuturesRwLock::new(HashMap::new())
```
vrutkovs, line 57: I'm a bit worried about cache growth - do we know if its significant? Should it be periodically purged? Should we use some kind of LRU here?

graph-builder/src/registry.rs
```diff
@@ -191,107 +208,116 @@ pub async fn fetch_releases<S: ::std::hash::BuildHasher>(
...
+                .insert(manifestref_key.to_owned(), manifestref);
+
+            // Process the manifest architecture if given
+            if let Some(arch) = arch {
+                // Encode the architecture as SemVer information
+                release.metadata.version.build =
+                    vec![semver::Identifier::AlphaNumeric(arch.to_string())];
+
+                // Attach the architecture for later processing
+                release.metadata.metadata.insert(
```
vrutkovs, line 305: TODO: double `metadata` is confusing, this should be reworked (in a follow up PR)?

graph-builder/tests/e2e/mod.rs
```diff
@@ -6,6 +6,86 @@ use std::env;
...
+        .unwrap()
+        .as_array_mut()
+        .unwrap()
+        .iter_mut()
+        .for_each(|ref mut edge: &mut serde_json::Value| {
+            if edge.as_array().unwrap().len() < 2 {
+                return;
+            }
+
+            macro_rules! rewrite_edge_index {
```
vrutkovs, line 71: Does it have to be defined in `for_each`? Could it be defined once instead?

graph-builder/tests/net/quay_io/mod.rs
```diff
@@ -131,6 +152,7 @@ fn fetch_release_private_with_credentials_must_succeed() {
             password.as_ref().map(String::as_ref),
             &mut cache,
             MANIFESTREF_KEY,
+            8,
```
vrutkovs: Lets use a `static` for this magic number (preferably defined in `graph-builder/src/config/options.rs`)

graph-builder/src/graph.rs
steveej: This was part of the compiler warning fix. I noticed it while working on this file.

graph-builder/src/registry.rs
steveej: I think the discussion isn't in the scope of this PR, as the cache semantics aren't changed here, but only how the cache is accessed.

graph-builder/src/registry.rs
steveej: Not in scope for this PR, as the code you're commenting on is not actually new but just changed its position and indentation.

graph-builder/tests/e2e/mod.rs
steveej: The macro makes direct use of the variable `edge` which is local to this `for_each`. I also don't foresee the usage of the macro outside of this location.

graph-builder/src/registry.rs
vrutkovs: Agree - lets watch the memory consumption on stage and act if it keeps growing

steveej:
Thanks for the review @LalatenduMohanty and @vrutkovs. I addressed all the comments, and, notably,  changed the default concurrency to 16.

vrutkovs:
LGTM! I'll merge this once e2e tests pass

dist/openshift/cincinnati.yaml
```diff
@@ -289,3 +290,6 @@ parameters:
   - name: PE_BINARY
     value: /usr/bin/policy-engine
     displayName: Path to policy-engine binary
+  - name: GB_UPSTREAM_FETCH_CONCURRENCY
+    value: 16
```
```suggestion
    value: "16"
```

This causes
`
 error: unable to load template file "dist/openshift/cincinnati.yaml": unable to decode "dist/openshift/cincinnati.yaml": v1.Template.Parameters: []v1.Parameter: v1.Parameter.Value: ReadString: expects " or n, but found 1, error found in #10 byte of ...|,"value":16}]}|..., bigger context ...|","name":"GB_UPSTREAM_FETCH_CONCURRENCY","value":16}]}|...
`
in e2e tests

vrutkovs (changes requested):

vrutkovs:
/hold
template is broken

dist/openshift/cincinnati.yaml
steveej: Oh, thanks for catching this.

steveej:
>  template is broken

Should be fixed now, thanks for catching!

vrutkovs:
graph-builder [fails to start](https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_cincinnati/174/pull-ci-openshift-cincinnati-master-e2e/89/artifacts/e2e/pods/cincinnati-e2e_cincinnati-1-lx7zz_cincinnati-graph-builder.log):
```
error: Invalid value for '--upstream.registry.fetch_concurrency <fetch_concurrency>': invalid digit found in string
```

steveej:
> `error: Invalid value for '--upstream.registry.fetch_concurrency <fetch_concurrency>': invalid digit found in string`

Does this mean that the concurrency value is `<fetch_concurrency>`, which means that there's still a mistake in the template or how it's evaluated.

steveej:
> > `error: Invalid value for '--upstream.registry.fetch_concurrency <fetch_concurrency>': invalid digit found in string`
> 
> Does this mean that the concurrency value is `<fetch_concurrency>`, which means that there's still a mistake in the template or how it's evaluated.

I think I fixed it with this patch which uses the correct braces:

```diff
diff --git a/dist/openshift/cincinnati.yaml b/dist/openshift/cincinnati.yaml
index 2b87a87..0d6cb0d 100644
--- a/dist/openshift/cincinnati.yaml
+++ b/dist/openshift/cincinnati.yaml
@@ -70,7 +70,7 @@ objects:
                 "--upstream.registry.url", "$(REGISTRY)",
                 "--upstream.registry.repository", "$(REPOSITORY)",
                 "--upstream.registry.credentials_path", "/etc/secrets/registry-credentials",
-                "--upstream.registry.fetch_concurrency", "$(GB_UPSTREAM_FETCH_CONCURRENCY)",
+                "--upstream.registry.fetch_concurrency", "${GB_UPSTREAM_FETCH_CONCURRENCY}",
               ]
               ports:
                 - name: graph-builder
```

vrutkovs:
e2e passes, great job!

/hold cancel

vrutkovs (approved):
/lgtm

LalatenduMohanty:
> Thanks for the review @LalatenduMohanty and @vrutkovs. I addressed all the comments, and, notably, changed the default concurrency to 16.

Any specific reason you changed it to 16 from 8.

graph-builder/src/registry.rs
vrutkovs: yep, we no longer fit in 256Mb on stage. Not sure if its due to HashMap or concurrency though

steveej:
> Any specific reason you changed it to 16 from 8.

I wanted to increase the effects and side effects.

PR #176, opened 2019-10-29
plugins/arch-filter: use simple fallback-to-default logic and remove the version/threshold handling

steveej:
Depends on
* [x] https://github.com/camallo/dkregistry-rs/pull/117
* [x] #179 

TODO
* [x] discuss adjustment of the global required parameters

---

### Successor to #180

### DIFF
--- a/cincinnati/src/plugins/internal/arch_filter.rs
+++ b/cincinnati/src/plugins/internal/arch_filter.rs
@@ -29,9 +29,6 @@ pub struct ArchFilterPlugin {
 
     #[default(DEFAULT_DEFAULT_ARCH.to_string())]
     pub default_arch: String,
-
-    #[default(DEFAULT_DEFAULT_ARCH_THRESHOLD_VERSION.to_string())]
-    pub default_arch_threshold_version: String,
 }
 
 impl PluginSettings for ArchFilterPlugin {
@@ -54,18 +51,10 @@ impl ArchFilterPlugin {
     }
 }
 
-/// Evaluate an architecture from the combination of "arch" and "version" parameters.
-///
-/// Assuming that version has been sent since the beginning of all versions,
-/// this function fails if `arch.is_none() && version > default_arch_threshold_version`.
-fn infer_version(
-    arch: Option<String>,
-    version: Option<String>,
-    default_arch: String,
-    default_arch_threshold_version: String,
-) -> Result<String, GraphError> {
-    match (arch, version) {
-        (Some(arch), _) => {
+/// Evaluate an architecture from the given "arch" parameters.
+fn infer_arch(arch: Option<String>, default_arch: String) -> Result<String, GraphError> {
+    match arch {
+        Some(arch) => {
             if !ARCH_VALIDATION_REGEX_RE.is_match(&arch) {
                 return Err(GraphError::InvalidParams(format!(
                     "arch '{}' does not match regex '{}'",
@@ -74,36 +63,14 @@ fn infer_version(
             };
             Ok(arch.to_string())
         }
-        (None, Some(version)) => {
-            let (semantic_version, threshold_semantic_version) = {
-                use std::str::FromStr;
-                (
-                    semver::Version::from_str(&version)
-                        .map_err(|e| GraphError::ArchVersionError(e.to_string()))?,
-                    semver::Version::from_str(&default_arch_threshold_version)
-                        .map_err(|e| GraphError::ArchVersionError(e.to_string()))?,
-                )
-            };
-
-            if semantic_version > threshold_semantic_version {
-                return Err(GraphError::MissingParams(
-                    vec!["arch"].into_iter().map(String::from).collect(),
-                ));
-            }
-
+        None => {
             debug!(
-                "no architecture given. inferring {} by version {} <= {}",
-                default_arch, version, default_arch_threshold_version
+                "no architecture given. assuming the default {}",
+                default_arch
             );
 
             Ok(default_arch)
         }
-        (None, None) => Err(GraphError::MissingParams(
-            vec!["arch", "version"]
-                .into_iter()
-                .map(String::from)
-                .collect(),
-        )),
     }
 }
 
@@ -117,17 +84,10 @@ lazy_static! {
 
 impl InternalPlugin for ArchFilterPlugin {
     fn run_internal(self: &Self, internal_io: InternalIO) -> AsyncIO<InternalIO> {
-        let arch = {
-            let interesting_params =
-                try_get_multiple_values!(internal_io.parameters, "arch", "version");
-
-            infer_version(
-                interesting_params.0.map(std::string::ToString::to_string),
-                interesting_params.1.map(std::string::ToString::to_string),
-                self.default_arch.clone(),
-                self.default_arch_threshold_version.clone(),
-            )
-        }
+        let arch = infer_arch(
+            internal_io.parameters.get("arch").map(|s| s.to_string()),
+            self.default_arch.clone(),
+        )
         .map_err(failure::Error::from);
 
         let future_result = futures::future::result(arch)
@@ -276,7 +236,6 @@ mod tests {
             key_prefix: "release".to_string(),
             key_suffix: "arch".to_string(),
             default_arch: "amd64".to_string(),
-            default_arch_threshold_version: "1.0.0".to_string(),
         })
         .run_internal(InternalIO {
             graph: input_graph.clone(),
@@ -294,26 +253,18 @@ mod tests {
     }
 
     #[test]
-    fn ensure_infer_version() -> Fallible<()> {
-        // (arch, version, default_arch, default_arch_threshold), expecteded_arch
+    fn ensure_infer_arch() -> Fallible<()> {
+        // (arch, default_arch), expecteded_arch
         [
-            ((None, Some("1.0.0"), "amd64", "2.0.0"), "amd64"),
-            ((Some("arm64"), Some("1.0.0"), "amd64", "2.0.0"), "arm64"),
-            ((None, Some("2.0.0-rc.0"), "amd64", "2.0.0"), "amd64"),
-            (
-                (Some("arm64"), Some("2.0.0-rc.0"), "amd64", "2.0.0"),
-                "arm64",
-            ),
+            ((None, "amd64"), "amd64"),
+            ((Some("arm64"), "amd64"), "arm64"),
+            ((None, "amd64"), "amd64"),
+            ((Some("arm64"), "amd64"), "arm64"),
         ]
         .iter()
         .try_for_each(|(params, expected_arch)| -> Fallible<()> {
             assert_eq!(
-                infer_version(
-                    params.0.map(str::to_string),
-                    params.1.map(str::to_string),
-                    params.2.to_string(),
-                    params.3.to_string()
-                )?,
+                infer_arch(params.0.map(str::to_string), params.1.to_string(),)?,
                 expected_arch.to_string()
             );
 


cincinnati/src/plugins/internal/arch_filter.rs
```diff
@@ -0,0 +1,271 @@
...
+            if !ARCH_VALIDATION_REGEX_RE.is_match(&arch) {
+                return Err(GraphError::InvalidParams(format!(
+                    "arch '{}' does not match regex '{}'",
+                    arch, ARCH_VALIDATION_REGEX_STR
+                )));
+            };
+
+            let mut graph = internal_io.graph;
+
+            // find all releases with a different the arch
```
vrutkovs:
```suggestion
            // find similar releases with different arch suffixes and rewrite version to create a single multiarch release
```
Not entirely sure this function does that though

cincinnati/src/plugins/internal/edge_add_remove.rs
```diff
@@ -168,15 +173,17 @@ impl EdgeAddRemovePlugin {
         graph
             .find_by_metadata_key(&format!("{}.{}", self.key_prefix, "previous.add"))
             .into_iter()
-            .try_for_each(|(to, to_string, from_csv)| -> Fallible<()> {
-                for from_string in from_csv.split(',').map(str::trim) {
-                    if let Some(from) = graph.find_by_version(&from_string) {
-                        info!("[{}]: adding {} {}", &to_string, "previous", &from_string);
+            .try_for_each(|(to, to_version, from_csv)| -> Fallible<()> {
+                for from_version in from_csv.split(',').map(str::trim) {
+                    let from_version = try_annotate_semver_build(&mut graph, from_version, &to)?;
```
vrutkovs: Not sure if its a good idea to replace `from_version` here

cincinnati/src/plugins/internal/edge_add_remove.rs
```diff
@@ -186,15 +193,17 @@ impl EdgeAddRemovePlugin {
...
             .find_by_metadata_key(&format!("{}.{}", self.key_prefix, "next.add"))
             .into_iter()
-            .try_for_each(|(from, from_string, to_csv)| -> Fallible<()> {
-                for to_string in to_csv.split(',').map(str::trim) {
-                    if let Some(to) = graph.find_by_version(&to_string) {
-                        info!("[{}]: adding {} {}", &from_string, "next", &to_string);
-                        handle_add_edge!("next", from, to, from_string, to_string);
+            .try_for_each(|(from, from_version, to_csv)| -> Fallible<()> {
+                for to_version in to_csv.split(',').map(str::trim) {
+                    let to_version = try_annotate_semver_build(&mut graph, to_version, &from)?;
```
vrutkovs: Same here - I'm not sure if this replacement is intended, but it would probably complicate debugging

cincinnati/src/plugins/internal/arch_filter.rs
```diff
@@ -0,0 +1,271 @@
...
+                    trace!("rewriting version {} ->  {}", release.version(), version);
+
+                    match &mut release {
+                        crate::Release::Abstract(release) => release.version = version,
+                        crate::Release::Concrete(release) => release.version = version,
+                    };
+
+                    Ok(())
+                })
+                .map_err(|e| GraphError::InternalError(e.to_string()))?;
```
vrutkovs: Let make it more specific - `GraphError::ArchVersionError`?

steveej: Maybe `SemVerParseError`?

vrutkovs (approved):
looks good to me so far

steveej:
This PR is ready for review now.

@wking @vrutkovs @crawford I added some version inference logic described [here](https://github.com/openshift/cincinnati/pull/176/files#diff-297f403f38104d073b9549d51175a889R57-R60) which I'd like some feedback on.

vrutkovs (approved):

graph-builder/tests/images/test-public-multiarch-manual-0.0.0/Dockerfile
```diff
@@ -0,0 +1,2 @@
+FROM scratch
+COPY release-metadata release-manifests/release-metadata
```
wking: Do we expect these to diverge between images?  Might want to just bake this into the script from d091fac:

```sh
for target in "${@}"; do
  ...
  cat <<-EOF >"${target}/Dockerfile"
    FROM scratch
    COPY release-metadata release-manifests/release-metadata
    EOF
  ...
done
```

Or skip the Dockerfile and `buildah bud ...` altogether and do:

```sh
container="$(buildah from scratch)"
buildah copy "${container}" "${target}/release-metadata" /release-manifests/release-metadata
```

or whatever (untested).

wking:
Still working through this, but the "bucket of baby commits" approach is making review a lot easier than when this was a bigger, monolithic commit.  Thanks for taking the time to break it down :).

cincinnati/src/lib.rs
```diff
@@ -756,10 +756,11 @@ mod tests {
 
         let nodes: Vec<daggy::NodeIndex> = (start..(start + count))
             .map(|i| {
-                let metadata = metadata.remove(&i).unwrap_or_default();
+                let mut metadata = metadata.remove(&i).unwrap_or_default();
+                let suffix = metadata.remove("version_suffix").unwrap_or_default();
 
                 let release = Release::Concrete(ConcreteRelease {
-                    version: format!("{}.0.0", i),
+                    version: format!("{}.0.0{}", i, suffix),
```
wking: All of our exported graphs are going to be single-arch.  Do we need arch-specific names?  Or can we recycle "4.4.0" (say) to mean "the amd64 4.4.0" when the requested arch is amd64 and "the s390x 4.4.0" when the requested arch is s390x?

cincinnati/src/plugins/internal/arch_filter.rs
```diff
@@ -0,0 +1,356 @@
+//! This plugin can be used to filter a graph by a architecture.
```
wking: nit: "a architecture" -> "architecture"

cincinnati/src/plugins/internal/arch_filter.rs
```diff
@@ -0,0 +1,356 @@
...
+//! It reads the requested architecture from the parameters value at key "arch",
+//! and the value must match the regex specified at ARCH_VALIDATION_REGEX_STR
```
wking: We only hit this code after we've loaded the graph, right?  I'd rather drop the regexp and just track the set of all architectures seen during graph building.  Then if someone requests an arch we haven't seen we can say "invalid arch {whatever-they-gave-us}, try one of these {arches-we've-seen}".  If they give us a version string, we may even be able to filter that set and just return arch choices for that particular version, falling back to the set of all seen architectures if we don't recognize the version.

cincinnati/src/plugins/internal/arch_filter.rs
```diff
@@ -0,0 +1,356 @@
...
+        ensure!(!plugin.key_suffix.is_empty(), "empty arch-key suffix");
+
+        Ok(Box::new(plugin))
+    }
+}
+
+/// Evaluate an architecture from the combination of "arch" and "version" parameters.
+///
+/// Assuming that version has been sent since the beginning of all versions,
+/// this funcstion `if arch.is_none() && version > default_arch_threshold_version`.
```
wking: nit: "funcstion" -> "function"

cincinnati/src/plugins/internal/arch_filter.rs
```diff
@@ -0,0 +1,356 @@
...
+                // 2. collect all other releases to be removed
+                let to_remove = {
+                    graph
+                        .find_by_fn_mut(|release| {
+                            match release {
+                                crate::Release::Concrete(concrete_release) => concrete_release
+                                    .metadata
+                                    .get_mut(&format!("{}.{}", key_prefix, key_suffix))
+                                    .map_or(true, |values| {
+                                        if values.split(',').any(|value| value.trim() == arch) {
```
wking: This should be looking at the arch extracted from the release image via camallo/dkregistry-rs#117, right? Or is the graph builder using the dkregistry-rs code to fill in the arch metadata being consumed here?

cincinnati/src/plugins/internal/arch_filter.rs
```diff
@@ -0,0 +1,356 @@
...
+                        })
+                        .collect()
+                };
+
+                // remove all matches from the Graph
+                let removed = graph.remove_releases(to_remove);
+
+                trace!("removed {} releases", removed);
+
+                // remove the build suffix from the version
```
wking: I'd rather never have arch suffixes on the version strings.  That information can be extracted from the image and placed in release metadata and never be stuffed onto the signed version string.  Release _tags_ (which are outside the image and not signed) can include the arch if release admins want, but that shouldn't matter to Cincinnati.

cincinnati/src/plugins/internal/arch_filter.rs
steveej: > We only hit this code after we've loaded the graph, right?

The plugin receives the graph as part of its input when being executed in the context of a single request. I consider the plugins to be able to have certain requirements towards their input. I don't see the necessity to introduce a state here for keeping a history of all architectures, as the graph contains that state already. If a user requests an architecture which validates but isn't in the graph, they'll get an empty graph, which is fine IMO.

cincinnati/src/lib.rs
steveej: This method is used for generating a test graph with arch suffixes, which are (will be) used in production in the processing chain and are only removed by the final arch filter plugin. Eventually we might ship them on a custom endpoint, or on requests which specifically ask for it, to enable multi-arch clusters.

graph-builder/tests/images/test-public-multiarch-manual-0.0.0/Dockerfile
steveej: They are shared between a lot of images but not all of them. Some have multiple layers which are crafted to find issues with layer ordering or race-conditions.

cincinnati/src/plugins/internal/arch_filter.rs
steveej: > Or is the graph builder using the dkregistry-rs code to fill in the arch metadata being consumed here?

Yes

cincinnati/src/plugins/internal/arch_filter.rs
steveej: > I'd rather never have arch suffixes on the version strings.

The way the cincinnati library is written requires all version strings to be unique, hence I chose to encode the arch in it.



> That information can be extracted from the image and placed in release metadata and never be stuffed onto the signed version string.

That's what the graph-builder/registry code is doing.

>  Release _tags_ (which are outside the image and not signed) can include the arch if release admins want, but that shouldn't matter to Cincinnati.

It doesn't matter the way the graph-builder is implemented. As you mentioned earlier and I'm repeating now for completeness, we extract the architecture from the manifest, the tag is an opaque string to Cincinnati.

graph-builder/tests/images/test-public-multiarch-manual-0.0.0/Dockerfile
wking: > Some have multiple layers which are crafted to find issues with layer ordering or race-conditions.

Can we unit-test that separately?

cincinnati/src/lib.rs
wking: > Eventually we might ship them on a custom endpoint, or on requests which specifically ask for it, to enable multi-arch clusters.

Heterogenous clusters are going to need to use manifest-lists, at which point Cincinnati can go back to not caring about cluster architecture as far as nodes are concerned.  We might still care about cluster architecture as far as edge-stability is concerned.

cincinnati/src/plugins/internal/arch_filter.rs
wking: > If a user requests an architecture which validates but isn't in the graph, they'll get an empty graph, which is fine IMO.

That's fair.  But then why place validation constraints on the architecture string at all?

cincinnati/src/lib.rs
steveej: Gotcha, thanks for the outlook ;-) Can we conclude that we'll change now for the internal implementation, and we might change it back later when manifest-lists are generally available?

cincinnati/src/plugins/internal/arch_filter.rs
steveej: So that the user knows the difference between requesting an invalid architecture and an empty graph. The graph might be empty for a multitude of reasons which leaves the user guessing.

steveej:
[test]

cincinnati/src/plugins/internal/arch_filter.rs
wking: > So that the user knows the difference between requesting an invalid architecture and an empty graph.

What is the difference?  If you want to return an empty graph if the user asks for `arch=nevergonnahappen`, I don't see why you wouldn't also return an empty graph (and no error) for users who request `arch=never+gonna+happen`.

cincinnati/src/plugins/internal/arch_filter.rs
```diff
@@ -0,0 +1,343 @@
...
+        Ok(())
+    }
+
+    #[test]
+    fn ensure_infer_version() -> Fallible<()> {
+        // (arch, version, default_arch, default_arch_threshold), expecteded_arch
+        [
+            ((None, Some("1.0.0"), "amd64", "2.0.0"), "amd64"),
+            ((Some("arm64"), Some("1.0.0"), "amd64", "2.0.0"), "arm64"),
+            ((None, Some("2.0.0-rc0"), "amd64", "2.0.0"), "amd64"),
```
wking: nit: SemVer (e.g. [in 2.0.0-rc.2][1]) and us (e.g. with 4.2.0-rc.0) put a dot after the `rc`.  Pre-releases [sort earlier][2], so doesn't really matter, but you may want `2.0.0-rc.0` here and below for consistency.

[1]: https://semver.org/spec/v2.0.0-rc.2.html
[2]: https://semver.org/spec/v2.0.0.html#spec-item-9

cincinnati/src/plugins/internal/arch_filter.rs
```diff
@@ -0,0 +1,343 @@
...
+
+        Ok(Box::new(plugin))
+    }
+}
+
+/// Evaluate an architecture from the combination of "arch" and "version" parameters.
+///
+/// Assuming that version has been sent since the beginning of all versions,
+/// this function fails if `arch.is_none() && version > default_arch_threshold_version`.
+fn infer_version(
```
wking: nit: `infer_arch_from_version`?

graph-builder/src/graph.rs
```diff
@@ -296,15 +296,38 @@ pub fn run<'a>(settings: &'a config::AppSettings, state: &State) -> ! {
...
         .try_for_each(|release| {
-            let previous = release.metadata.previous.clone();
-            let next = release.metadata.next.clone();
+            let previous = release
+                .metadata
+                .previous
+                .iter()
+                .cloned()
+                .map(|mut previous| {
+                    previous.build = release.metadata.version.build.clone();
```
wking: I don't think we need this `build` mutation.  Can we ignore architecture when creating edges, link all versions regardless of `build`, and let the arch plugin filter any cross-arch edges out for us?

graph-builder/src/registry.rs
```diff
@@ -224,15 +224,27 @@ pub fn fetch_releases<S: ::std::hash::BuildHasher>(
...
         .collect();
-    let tagged_layers = thread_runtime.block_on(tags)?;
+    let manifest_and_ref_by_tag = thread_runtime.block_on(tags)?;
 
-    let mut releases = Vec::with_capacity(tagged_layers.len());
+    let mut releases = Vec::with_capacity(manifest_and_ref_by_tag.len());
+
+    for (tag, manifest, manifestref) in manifest_and_ref_by_tag {
+        let layers_digests = manifest
+            .layers_digests(None)
```
wking: Why don't we pass in our architecture explicitly here?  I see your `manifest.architectures()` call bit further down.

graph-builder/src/registry.rs
```diff
@@ -262,6 +274,34 @@ pub fn fetch_releases<S: ::std::hash::BuildHasher>(
...
+
+                // Encode the architecture as SemVer information
+                release.metadata.version.build =
+                    vec![semver::Identifier::AlphaNumeric(arch.clone())];
+
+                // Attach the architecture for later processing
+                release
+                    .metadata
+                    .metadata
+                    .insert("io.openshift.upgrades.graph.release.arch".to_owned(), arch);
```
wking: Should we be using the constants from the arch filter's `key_prefix` and `key_suffix` (both of which are `pub`, right?) instead of hard-coding `io.openshift.upgrades.graph.release.arch`?

cincinnati/src/plugins/internal/edge_add_remove.rs
```diff
@@ -205,6 +221,29 @@ impl EdgeAddRemovePlugin {
     }
 }
 
+/// Try to find the architecture metadata and add it to the version String assuming SemVer.
+///
+/// If the referenced ReleaseId doesn't have the arch metadata, the version
+/// string will be passed through unchanged.
+fn try_annotate_semver_build(
```
wking: As in [this comment][1], I'd rather draw edges to all matching versions regardless of arch, and let the arch-plugin go through and cull any cross-arch stuff at the end.  Easiest way to do that is probably to use a new `graph.iter_by_version(version: &str)` that iterates over all nodes and returns any where the *original* version (before we inject the uniquifying arch into `build`) matches.  Then you can iterate over the result and add/remove edges to/from all matching releases. 

[1]: https://github.com/openshift/cincinnati/pull/176#discussion_r347744320

graph-builder/tests/images/test-public-multiarch-manual-0.0.0/Dockerfile
steveej: I think so, yes. Let's consider that in the test refactoring we have planned and on which I'll be working on next. It's not a multi-arch specific issue, so I would prefer not to block this PR on that.

cincinnati/src/plugins/internal/arch_filter.rs
steveej: How about `arch=amd-64` vs `arch=amd64`? I'm thinking of the case where the user provides an arch which will eventually emit a non-empty graph for the request which he's making now, but he has a typo in the string which we could tell them about.

cincinnati/src/plugins/internal/arch_filter.rs
steveej: I didn't want to over-specify, as we might include other data -- e.g. telemetry data from the clusterID -- later on.

graph-builder/src/graph.rs
steveej: > I don't think we need this `build` mutation.

This mutates the `build` of the previous (and further down next) edge declaration. The nodes in the graph they refer to are already arch-annotated through their build version. Without the mutation here we wouldn't find the correct edge endpoint. If an edge endpoint isn't found, [it creates an abstract release as a place-holder][abstract-create]. This is required because the insertion process has to handle the insertion of a node which declares edges to versions of nodes which are not yet inserted in the graph. If we skip the mutation discussed here, we'll end up with edges to abstract releases in the final graph, and fail to connect the edge to the desired annotated concrete releases.

[abstract-create]: https://github.com/steveeJ-forks/openshift-cincinnati/blob/0d5823ab6c58e25cd966d737ec1915ef5bbb0525/graph-builder/src/graph.rs#L336-L338

> Can we ignore architecture when creating edges, link all versions regardless of `build`, and let the arch plugin filter any cross-arch edges out for us?

IIUC that would be a different implementation of the same logic, where instead of creating arch-annotated release nodes, we'd have arch-annotated edges. The reason why I chose the former is simply because it was easier to do with the existing code.

graph-builder/src/registry.rs
steveej: I wanted to postpone the handling of this, but since you brought it up I addressed it now. The code enforces that we only get one architecture per manifest until we support manifest lists.

cincinnati/src/plugins/internal/arch_filter.rs
steveej: Coming back to this, I failed to read this properly. I almost agree with your nit. My intention was to name it `infer_arch` and not specify the information considered for the inference in the function name.

cincinnati/src/plugins/internal/arch_filter.rs
wking: > I'm thinking of the case where the user provides an arch which will eventually emit a non-empty graph for the request which he's making now, but he has a typo in the string which we could tell them about.

I'm fine with that to.  But a generic regexp approach is only going to catch so many of those.  For example, it's not going to catch `nevergonnahappen` unless folks are manually maintaining the regexp to list expected arches.  And what would you put in a list of expected arches besides the arches you have in your graph?  So I see too paths:

a. The "user knows what they're doing" path, where we just accept whatever they gave for `arch` and return an empty graph with no error.
b. The "our graph has the set of arches" path, where we error out in this plugin with "unrecognised arch: `${whatever}`" if arch-filtering by the given `arch` empties the graph.

Trying to sit somewhere between those two seems like manual effort for questionable benefit.

graph-builder/src/graph.rs
wking: > Without the mutation here we wouldn't find the correct edge endpoint.

It wouldn't be found if you use `graph.find_by_version`.  But you'd find (potentially several) endpoints if you used a new `graph.iter_by_version(version: &str)` like I floated [here][1], right?

> This is required because the insertion process has to handle the insertion of a node which declares edges to versions of nodes which are not yet inserted in the graph.

Wait, why are we processing edges before we have added all the nodes?  Can't we slurp in the nodes (with whatever metadata, including edge requests), and then leave edge processing to the edge-add-remove plugin?

> IIUC that would be a different implementation of the same logic, where instead of creating arch-annotated release nodes, we'd have arch-annotated edges.

They wouldn't be annotated edges.  They'd be removed by the arch plugin because when it removed a non-matching node it would also remove all edges touching that node.

[1]: https://github.com/openshift/cincinnati/pull/176#discussion_r347750531

cincinnati/src/plugins/internal/arch_filter.rs
steveej: I think (a) and (b) can be combined. We definitely know and will tell the user if they provided an architecture which is incorrect by means of the underlying technology used, namely SemVer build annotations, which we can inform the user about. We could additionally tell them that the graph doesn't contain their syntactically correct architecture as an error, I'd be fine with that. But I wouldn't track the list of architectures over time. Rather I'd compile it for each request from the input graph of the plugin.

graph-builder/src/graph.rs
steveej: >  But you'd find (potentially several) endpoints if you used a new `graph.iter_by_version(version: &str)` like I floated [here](https://github.com/openshift/cincinnati/pull/176#discussion_r347750531), right?

That entails finding and connecting releases of different architectures as the one which is currently looked at. That doesn't translate the input data correctly (i.e. it breaks) in cases where architectures don't have the same edge constellation.

cincinnati/src/plugins/internal/edge_add_remove.rs
steveej: For reasons explained in https://github.com/openshift/cincinnati/pull/176#discussion_r348570055 I'd like to keep the Cincinnati core library correct and do the fiddling in the plugins.

steveej:
@wking thanks for another review! I added a few comments and a commit on top for further discussion. I would like to mention that the code works according to my understanding of the intention, which should be covered by the tests. It also tries to be correct for all of the possible Cincinnati use-cases, not just the OpenShift specific one.

cincinnati/src/plugins/internal/arch_filter.rs
wking: > We definitely know and will tell the user if they provided an architecture which is incorrect by means of the underlying technology used, namely SemVer build annotations, which we can inform the user about.

I'm having trouble parsing this.  Can you give a concrete example?

> We could additionally tell them that the graph doesn't contain their syntactically correct architecture as an error, I'd be fine with that.

I don't think there's a need to separate out "syntactically correct"; just say "error!  No matches for your provided `arch=${ARCH}`".

> But I wouldn't track the list of architectures over time. Rather I'd compile it for each request from the input graph of the plugin.

So we agree?  Can we drop `ARCH_VALIDATION_REGEX_STR`?

graph-builder/src/graph.rs
wking: > That entails finding and connecting releases of different architectures as the one which is currently looked at.

Yup.  But that's fine, we'll filter mismatches out in the arch plugin when we drop the non-matching nodes.

> That doesn't translate the input data correctly (i.e. it breaks) in cases where architectures don't have the same edge constellation.

What's an edge constellation?

> I slightly changed the code to first add the releases and subsequently add the edges.

Is this 92f6cb3?  I think we can probably drop edge handling entirely except for the edge-add-remove plugin and removing associated edges whenever we remove a node.  But that doesn't have to happen in this PR.

wking:
Still some ongoing discussion in [these][1] and [threads][2].  But we're out of time to bikeshed in this PR, so punting those to follow up work.  I'll push the WIP 92f6cb309962 off this branch, and we can get the rest of this landed for staging.

[1]: https://github.com/openshift/cincinnati/pull/176#discussion_r345516455
[2]: https://github.com/openshift/cincinnati/pull/176#discussion_r347744320

cincinnati/src/plugins/internal/arch_filter.rs
```diff
@@ -99,10 +99,13 @@ fn infer_version(
             Ok(default_arch)
         }
         (None, None) => Err(GraphError::MissingParams(
-            vec!["arch", "version"]
-                .into_iter()
-                .map(String::from)
-                .collect(),
+            vec![&format!(
+                "arch or version=<{}",
+                default_arch_threshold_version
```
steveej: @wking I'm not sure if we want to include the threshold in the error message

cincinnati/src/plugins/internal/arch_filter.rs
steveej: > I'm having trouble parsing this. Can you give a concrete example?

If the user provides an architecture which contains an invalid character according to the [SemVer specification](https://semver.org/#spec-item-10https://semver.org/#spec-item-10) we can let them know. This may be caused due to a simple typo, e.g. "amd64.". This is different from an architecture name which is valid but has no releases in the graph.

> I don't think there's a need to separate out "syntactically correct"; just say "error! No matches for your provided `arch=${ARCH}`".

We don't *need* to do anything :smile: I'd do it to give the possibly most precise error message to the user. I don't understand where your resistance towards distinguishing these two cases is coming from :man_shrugging:

> So we agree? Can we drop `ARCH_VALIDATION_REGEX_STR`?

Not really. We could have the regex check *and* the non-existing arch check. The latter would naturally take precedence as it's cheaper than looking through the graph. But if you really feel so strongly about it I can also remove the regex and only leave the non-existing arch check in.

cincinnati/src/plugins/internal/arch_filter.rs
wking: this is about folks updating their callers, so I think it's fine to just say "set `arch`".  Our supporting `version=4.1.z` as an alternative can be an undocumented internal workaround.

graph-builder/src/graph.rs
steveej: > Yup. But that's fine, we'll filter mismatches out in the arch plugin when we drop the non-matching nodes.

I'm not sure if it's fine. It's inaccurate, as it loses detail which the input data provides.



> What's an edge constellation?

What I meant by that is that, according to the input data, one architecture could have different edges than another architecture.



> Is this [92f6cb3](https://github.com/openshift/cincinnati/commit/92f6cb3099629b7d2656f25fe1942fd03be07e49)? I think we can probably drop edge handling entirely except for the edge-add-remove plugin and removing associated edges whenever we remove a node. But that doesn't have to happen in this PR.

That would require to copy all previous/next information from the release-metadata files into the release metadata which ends up in the graph, because that plugin doesn't have access to the former.

wking: > > But that's fine, we'll filter mismatches out in the arch plugin when we drop the non-matching nodes.
>
> I'm not sure if it's fine. It's inaccurate, as it loses detail which the input data provides.

But the input data does not contain arch-specific names, so we're not losing detail by creating edges to all possible versions regardless of arch.

> What I meant by that is that, according to the input data, one architecture could have different edges than another architecture.

I think that's fine, since it's temporary internal state.

> That would require to copy all previous/next information from the release-metadata files into the release metadata which ends up in the graph, because that plugin doesn't have access to the former.

That's fine, right?  I thought the idea was "we don't care if it was primary or secondary, it's all just metadata".

steveej: > But the input data does not contain arch-specific names, so we're not losing detail by creating edges to all possible versions regardless of arch.

The images we scrape are arch specific thus all metadata, including its previous/next declarations, is specific to that arch. As we scrape all images of a given repository we add their arch-specific releases to the graph.

> I think that's fine, since it's temporary internal state.

I don't think it's fine, as the graph could end up with edges which weren't defined by any metadata.

> That's fine, right? I thought the idea was "we don't care if it was primary or secondary, it's all just metadata".

That should be fine. I suggest we handle that in a separate PR.

steveej:
[test]

steveej:
@LalatenduMohanty this PR now only contains the changes we discussed out-of-band. PTAL

cincinnati/src/plugins/internal/arch_filter.rs
```diff
@@ -74,36 +63,14 @@ fn infer_version(
...
-                    vec!["arch"].into_iter().map(String::from).collect(),
-                ));
-            }
-
+        None => {
             debug!(
-                "no architecture given. inferring {} by version {} <= {}",
-                default_arch, version, default_arch_threshold_version
+                "no architecture given. assuming the default {}",
+                default_arch
```
LalatenduMohanty, line 69: :+1:

LalatenduMohanty (approved):
/lgtm

steveeJ:
[testp

steveej:
[test]

PR #179, opened 2019-11-14
cincinnati: expose customizable testgraph generator

steveej:
This code is useful for generating Graphs for testing purposes and can
be used in other crates as well. In addition to exposing the existing
functionality a builder with a few possible customizations have been
added.

### DIFF
--- a/cincinnati/src/lib.rs
+++ b/cincinnati/src/lib.rs
@@ -702,15 +702,10 @@ impl From<Graph> for plugins::interface::Graph {
     }
 }
 
-#[cfg(test)]
-mod tests {
-    extern crate serde_json;
-
+pub mod testing {
     use super::*;
 
-    type TestResult<T> = Result<T, Box<dyn std::error::Error>>;
-
-    pub(crate) fn generate_graph() -> Graph {
+    pub fn generate_graph() -> Graph {
         let mut graph = Graph::default();
         let v1 = graph.dag.add_node(Release::Concrete(ConcreteRelease {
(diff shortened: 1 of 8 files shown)


cincinnati/src/lib.rs
```diff
@@ -734,47 +729,134 @@ mod tests {
...
+            self.edges = edges;
+            self
+        }
+
+        pub fn with_version_template(mut self, version_template: &str) -> Self {
+            self.version_template = version_template.to_string();
+            self
+        }
+
+        pub fn enable_payload_suffix(mut self, enable: bool) -> Self {
```
vrutkovs, line 781: It seems this doesn't get tested yet, right?

vrutkovs (approved):
LGTM. Not sure if its worth including multiarch bits right now, can be done in follow up PRs

cincinnati/src/lib.rs
steveej: Right, i's not used in any tests as of the commit it's being added in.

steveej (commented):
> LGTM. Not sure if its worth including multiarch bits right now, can be done in follow up PRs

If you're referring to the suffix functionality, it's not restricted to multi-arch, although that's certainly the first use-case for it. I'm fine introducing these options here as opposed to stripping this part out into the other PR. 

Would you mind turning your LGTM into an actual one so it can be merged? :wink:

vrutkovs:
Okay, I think its safe to get merged now, more commits on this are coming.

/lgtm

PR #182, opened 2019-11-25
dist/openshift: inline configmap in template

steveej:
I was part of a longer discussion with @maorfr and @pbergene. It may or may not become possible for us to deploy the configmap from the repository.

Besides that, @maorfr is this syntactically correct?


dist/openshift/cincinnati.yaml
```diff
@@ -199,6 +199,23 @@ objects:
...
+    kind: ConfigMap
+    metadata:
+      name: cincinnati
+    type: Opaque
+    data:
+      gb.address: "0.0.0.0"
+      gb.status.address: "0.0.0.0"
+      gb.registry: "quay.io"
+      gb.repository: "openshift-release-dev/ocp-release"
+      gb.log.verbosity: ${{GB_LOG_VERBOSITY}}
```
maorfr, line 212: Using `${{VAR}}` is for variables which are evaluated as ints, while `${VAR}` is for strings. Notice how double curly brackets are used for ports in this file.

dist/openshift/cincinnati.yaml
```diff
@@ -199,6 +199,23 @@ objects:
...
+      pe.address: "0.0.0.0"
+      pe.status.address: "0.0.0.0"
+      pe.upstream: "http://localhost:8080/v1/graph"
+      pe.log.verbosity: ${{PE_LOG_VERBOSITY}}
```
maorfr, line 216: same here

maorfr (changes requested):
two minor comments.

in addition, you will need to add the new parameters to the saas repo as well, as we've discussed: https://gitlab.cee.redhat.com/service/saas-cincinnati/blob/master/cincinnati-services/cincinnati.yaml

nice work!

dist/openshift/cincinnati.yaml
steveej: Thanks for that insight!

dist/openshift/cincinnati.yaml
steveej: same [as here](https://github.com/openshift/cincinnati/pull/182#discussion_r350620303) :smile:

dist/openshift/cincinnati.yaml
steveej: I double-checked this in [the documentation on parameters](https://docs.openshift.com/container-platform/3.11/dev_guide/templates.html#writing-parameters), and wanted to add my findings here for completeness.

Double curly brackets use the json/yaml value as is, and can't be used as part of a string; it has to be used as a single value. E.g. `value: ${{PARAM}}` vs `value: "${PARAM1}-${PARAM2}"`. 

In conclusion, it's my understanding it's not the distinction between integers and strings per-se. Although the double curly brackets can indeed be used to pass an integer value through, one could as well pass through a string value.

maorfr:
/lgtm

totally fine by me, thanks for the doc reference!

PR #185, opened 2019-12-06
policy-engine/src/main: Add request logging

wking:
The default log format [is][1]:

```
%a %t "%r" %s %b "%{Referer}i" "%{User-Agent}i" %T
```

This will give us a user-agent to help debug broken clients.  Before this commit, logs looked like:

```
[2019-12-06T20:14:16Z ERROR policy_engine::graph] Error serving request with parameters '"channel=stable-4.1&version=4.2.4"': mandatory client parameters missing
[2019-12-06T20:14:29Z ERROR policy_engine::graph] Error serving request with parameters '"channel=stable-4.2&version=4.2.4"': mandatory client parameters missing
[2019-12-06T20:14:52Z ERROR policy_engine::graph] Error serving request with parameters '"channel=fast-4.2&version=4.2.4"': mandatory client parameters missing
```

which makes it hard to track down the broken clients.  Ideally we'd only include the user-agent logs for invalid requests, but I'm not clear on how to set that up.

Not sure how to test this either...

[1]: https://actix.rs/docs/middleware/#usage

### DIFF
--- a/policy-engine/src/main.rs
+++ b/policy-engine/src/main.rs
@@ -80,6 +80,7 @@ fn main() -> Result<(), Error> {
     registry.register(Box::new(BUILD_INFO.clone()))?;
     HttpServer::new(move || {
         App::new()
+            .wrap(actix_web::middleware::Logger::default())
             .register_data(actix_web::web::Data::new(RegistryWrapper(registry)))
             .service(
                 actix_web::web::resource("/metrics")


policy-engine/src/main.rs
```diff
@@ -80,6 +80,7 @@ fn main() -> Result<(), Error> {
     registry.register(Box::new(BUILD_INFO.clone()))?;
     HttpServer::new(move || {
         App::new()
+            .wrap(Logger::default())
```
steveej:
```suggestion
            .wrap(actix_web::middleware::Logger::default())
```

steveej: This suggestion will make it compile but it won't affect the log messages we produce manually, only the ones generated internally in the actix_web components.

wking: Pushed the fix with efdfd8e -> ea5c4e8.

> This suggestion will make it compile but it won't affect the log messages we produce manually...

Is there a way to add the user-agent prefix to *all* the log messages we produce manually while handling a request?  I'd rather not instrument each of our log lines directly like #186.

wking: > Is there a way to add the user-agent prefix to all the log messages we produce manually...

Sounds like #186 [does this][1], although I'm not clear on the mechanism.

/close

[1]: https://github.com/openshift/cincinnati/pull/186#discussion_r355347588

PR #186, opened 2019-12-06
pe/graph: verbose-log request on error

steveej:
This will make the output look like

```
[2019-12-06T21:57:22Z ERROR policy_engine::graph] Error serving request ' HttpRequest HTTP/1.1 GET:/api/upgrades_info/v1/graph   query: ?"channel=stable-4.1"   headers:     "accept": "application/json"     "host": "localhost:8081"     "user-agent": "curl/7.65.3" ' from '127.0.0.1:34370': MissingParams(["arch", "version"])
```

when a request lead to an error.

Alternative to #185.

### DIFF
--- a/policy-engine/src/graph.rs
+++ b/policy-engine/src/graph.rs
@@ -95,8 +95,11 @@ pub(crate) fn index(req: HttpRequest) -> Box<dyn Future<Item = HttpResponse, Err
 
             if let Err(e) = &r {
                 error!(
-                    "Error serving request with parameters '{:?}': {}",
-                    req.query_string(),
+                    "Error serving request '{}' from '{}': {:?}",
+                    format!("{:?}", &req).replace("\n", " ").replace("\t", " "),
+                    &req.peer_addr()
+                        .map(|addr| addr.to_string())
+                        .unwrap_or("<not available>".into()),
                     e
                 );
             }


policy-engine/src/graph.rs
```diff
@@ -95,8 +95,11 @@ pub(crate) fn index(req: HttpRequest) -> Box<dyn Future<Item = HttpResponse, Err
 
             if let Err(e) = &r {
                 error!(
-                    "Error serving request with parameters '{:?}': {}",
-                    req.query_string(),
+                    "Error serving request '{}' from '{}': {:?}",
+                    format!("{:?}", &req).replace("\n", " ").replace("\t", " "),
+                    &req.peer_addr()
+                        .map(|addr| addr.to_string())
+                        .unwrap_or("<not available>".into()),
```
wking, line 102: this is not the only type of 400 error we generate, right?  I'd like to have User-Agents on all the errors we spit out when returning a 400 response.

steveej: Following the chain of combinators like `and_then`, `map`, etc., you'll be able to notice that this  `then`, in which we log the error, is unconditionally attached at the end. The logging thus applies to all responses which contain an error regardless of the specific type. The `index` function which this is a part of is used as the handler for the `v1/graph` endpoint, so it handles all errors going out of that. If you can generate a request which leads to an error and is not covered by this log output I'll consider it a bug.

LalatenduMohanty (approved):
/lgtm

vrutkovs (approved):
/lgtm

PR #188, opened 2019-12-10
graph-builder/create_graph: add all releases before adding edges

steveej:
This gives the algorithm the chance to detect and warn about edges which
refer to a non-existing release.

/cc @LalatenduMohanty

### DIFF
--- a/cincinnati/src/plugins/internal/graph_builder/release.rs
+++ b/cincinnati/src/plugins/internal/graph_builder/release.rs
@@ -16,8 +16,9 @@ use crate as cincinnati;
 
 use self::cincinnati::MapImpl;
 
+use failure::Fallible;
 use itertools::Itertools;
-use log::trace;
+use log::{trace, warn};
 use semver::Version;
 use serde::{Deserialize, Serialize};
 use std::fmt;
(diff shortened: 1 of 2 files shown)


LalatenduMohanty (changes requested):
unit tests?

graph-builder/src/graph.rs
```diff
@@ -306,49 +306,55 @@ pub fn create_graph(releases: Vec<registry::Release>) -> Result<Graph, Error> {
...
-                .metadata
-                .next
-                .iter()
-                .cloned()
+                .try_for_each(|version| {
+                    let previous = match graph.find_by_version(&version.to_string()) {
+                        Some(id) => id,
+                        None => {
+                            warn!("Adding abstract release for {}", version.to_string());
+                            graph.add_release(Release::Abstract(AbstractRelease {
```
wking: if you add all the releases first, do you still need the concept of `AbstractRelease`?

steveej: @crawford thinks they could be useful if an edge refers to a release which was never officially released, but might have made it to clusters somehow.

steveej:
Looks like a network issue in the CI cluster:

```
failures:

---- net::test_public_stream_active_tags stdout ----
thread 'net::test_public_stream_active_tags' panicked at 'called `Result::unwrap()` on an `Err` value: reqwest::Error { kind: Request, url: "https://quay.io/api/v1/repository/redhat/openshift-cincinnati-test-public-manual/tag?onlyActiveTags=true", source: hyper::Error(Connect, ConnectError("dns error", Custom { kind: Other, error: "failed to lookup address information: Name or service not known" })) }', /rustc/b8cedc00407a4c56a3bda1ed605c6fc166655447/src/libcore/ops/function.rs:154:5
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

/retest

cincinnati/src/plugins/internal/graph_builder/release.rs
```diff
@@ -80,70 +81,89 @@ pub fn create_graph(releases: Vec<Release>) -> Result<cincinnati::Graph, failure
...
+                version: semver::Version::from((0, 0, 1)),
+                next: Default::default(),
+                previous: vec![semver::Version::from((0, 0, 0))],
+                metadata: Default::default(),
+            },
+        }];
+
+        let mut graph = create_graph(releases).unwrap();
+
+        assert_eq!(graph.prune_abstract(), 1);
```
LalatenduMohanty, line 165: Why do we need to call `graph.prune_abstract()`? We do not use it in the `pub fn create_graph` so using it here is confusing.

LalatenduMohanty (changes requested):

cincinnati/src/plugins/internal/graph_builder/release.rs
steveej: `Graph::prune_abstract` prunes and returns the number of abstract releases in the graph. The test uses this to assert that there was indeed 1 abstract release there. I found this the easiest way to ensure the currently expected behavior.

steveej: We are not using the method anywhere in the code because we currently don't explicitly prune abstract releases from the graph. Due to their lack of metadata they are implicitly filtered out by plugins such as the channel filter or the architecture filter.

steveej:
(flake) Hit the AWS rate limit.

/retest

cincinnati/src/plugins/internal/graph_builder/release.rs
LalatenduMohanty: Got it, thanks.

LalatenduMohanty (approved):
/lgtm

PR #191, opened 2020-01-08
 Add scripts for upstream e2e tests 

vrutkovs:
This adds a minimal e2e test, which starts a new Cincinnati instance, 
pointing to test data, and attempts to fetch channel data.

In order to point test Cincy to a different repo the template is modified so that GB repo can be updated

Related to https://github.com/openshift/release/pull/6600

### DIFF
--- a/dist/Dockerfile.e2e/Dockefile
+++ b/dist/Dockerfile.e2e/Dockefile
@@ -1,0 +1,14 @@
+FROM centos:7
+
+ENV HOME="/root"
+
+RUN mkdir -p "${HOME}/cincinnati"
+WORKDIR "${HOME}/cincinnati"
+
+# Get oc CLI
+RUN mkdir -p ${HOME}/bin && \
+    curl https://mirror.openshift.com/pub/openshift-v4/clients/oc/latest/linux/oc.tar.gz 2>/dev/null | tar xzf - -C "${HOME}/bin/" oc
+ENV PATH="${PATH}:${HOME}/bin"
+COPY . .
+
+ENTRYPOINT ["hack/e2e.sh"]
--- a/dist/openshift/cincinnati.yaml
+++ b/dist/openshift/cincinnati.yaml
@@ -265,10 +265,10 @@ parameters:
     displayName: Graph builder status port
   - name: PE_PORT
     value: "8081"
-    displayName: Policy enigine port
+    displayName: Policy engine port
   - name: PE_STATUS_PORT
     value: "9081"
-    displayName: Policy enigine status port
+    displayName: Policy engine status port
   - name: PE_PATH_PREFIX
     value: "/api/upgrades_info"
     displayName: Policy engine path prefix
--- a/dist/openshift/cincinnati.yaml
+++ b/dist/openshift/cincinnati.yaml
@@ -208,7 +208,7 @@ objects:
       gb.address: "0.0.0.0"
       gb.status.address: "0.0.0.0"
       gb.registry: "quay.io"
-      gb.repository: "openshift-release-dev/ocp-release"
+      gb.repository: ${{GB_CINCINNATI_REPO}}
       gb.log.verbosity: ${{GB_LOG_VERBOSITY}}
       pe.address: "0.0.0.0"
       pe.status.address: "0.0.0.0"
@@ -278,3 +278,6 @@ parameters:
   - name: PE_LOG_VERBOSITY
     value: "vv"
     displayName: Policy engine log verbosity
+  - name: GB_CINCINNATI_REPO
+    value: "openshift-release-dev/ocp-release"
+    displayName: Graph builder quay repo
--- a/hack/e2e.sh
+++ b/hack/e2e.sh
@@ -1,0 +1,24 @@
+#!/usr/bin/env bash
+#
+# This script assumes its running from a pod and does the following:
+# * Processes the template to setup a local cincy instance
+# * Prepares a graph using test data
+# * Ensures generated graph is valid
+
+# Debug information
+oc whoami
+oc project
+
+# Apply oc template
+oc new-app -f ../dist/openshift/cincinnati.yaml \
+  -p IMAGE=${CINCINNATI_E2E_IMAGE:-pipeline} \
+  -p IMAGE_TAG=${CINCINNATI_E2E_IMAGE_TAG:-deploy} \
+  -p PE_PORT=${CINCINNATI_PE_PORT:-8081} \
+  -p GB_CINCINNATI_REPO="redhat/openshift-cincinnati-test-public-manual"
+
+# Wait for dc to rollout
+oc wait --for=condition=available deploymentconfig/cincinnati
+
+# Check that policy engine returns channel data respond
+GRAPH_URL="http://${CINCINNATI_PE_BASE_URL:-cincinnati-policy-engine}:${CINCINNATI_PE_PORT:-8081}/v1/graph"
+curl --verbose --header 'Accept:application/json' "${GRAPH_URL}?channel=a"


dist/Dockerfile.e2e/Dockefile
```diff
@@ -0,0 +1,15 @@
+FROM centos:7
+
+ENV HOME="/root"
+
+# Get oc CLI
+RUN mkdir -p ${HOME}/bin && \
+    curl https://mirror.openshift.com/pub/openshift-v4/clients/ocp-dev-preview/4.3.0-0.nightly-2019-11-13-233341/openshift-client-linux-4.3.0-0.nightly-2019-11-13-233341.tar.gz 2>/dev/null | tar xzf - -C /tmp/bin/ oc
```
wking: Do we really need a nightly `oc`?  I'd rather use 4.2.13's or the [floating latest][1] or something.

[1]: https://mirror.openshift.com/pub/openshift-v4/clients/oc/latest/linux/

wking: Also, this seems to be unpacking into `/tmp/bin`, not `${HOME}/bin`?

dist/Dockerfile.e2e/Dockefile
```diff
@@ -0,0 +1,15 @@
...
+ENV PATH="${PATH}:${HOME}/bin"
+
+RUN mkdir -p "${HOME}/cincinnati"
+WORKDIR ${HOME}/cincinnati
```
wking: nit: moving these eariler will mean we can preserve those layers when we bump `oc`.  Unlikely to matter, but I like having more-stable steps earlier in the `Dockerfile` when we can do that without making things awkward.

vrutkovs: Sounds like a good idea, implemented

dist/Dockerfile.e2e/Dockefile
vrutkovs: Floating latest would be best, right, updated

dist/Dockerfile.e2e/Dockefile
```diff
@@ -0,0 +1,14 @@
+FROM centos:7
+
+ENV HOME="/root"
+
+RUN mkdir -p "${HOME}/cincinnati"
+WORKDIR ${HOME}/cincinnati
+
+# Get oc CLI
+RUN mkdir -p ${HOME}/bin && \
+    curl https://mirror.openshift.com/pub/openshift-v4/clients/oc/latest/linux/oc.tar.gz 2>/dev/null | tar xzf - -C /tmp/bin/ oc
```
wking: [Still][1] unpacking into `/tmp/bin`?

[1]: https://github.com/openshift/cincinnati/pull/191#discussion_r364451185

wking: Fixed with 9e4f7d1

dist/Dockerfile.e2e/Dockefile
```diff
@@ -0,0 +1,14 @@
+FROM centos:7
+
+ENV HOME="/root"
+
+RUN mkdir -p "${HOME}/cincinnati"
+WORKDIR ${HOME}/cincinnati
+
+# Get oc CLI
+RUN mkdir -p ${HOME}/bin && \
+    curl https://mirror.openshift.com/pub/openshift-v4/clients/oc/latest/linux/oc.tar.gz 2>/dev/null | tar xzf - -C ${HOME}/bin/ oc
```
wking: Unlikely to matter (who would put shell-sensitive chars in `HOME`?), but would be nice to quote `"${HOME}/bin"` here and in the line above.

dist/Dockerfile.e2e/Dockefile
vrutkovs: Oops, fixed

dist/Dockerfile.e2e/Dockefile
vrutkovs: Added quotes (to WORKDIR as well)

dist/openshift/cincinnati.yaml
```diff
@@ -278,3 +278,6 @@ parameters:
   - name: PE_LOG_VERBOSITY
     value: "vv"
     displayName: Policy engine log verbosity
+  - name: GB_CINCINNATI_REPO
```
wking, line 281: Took me a bit to look down at `displayName` in order to unpack `GB` to "graph builder".  I see we already have `PE` for "policy engine".  I'd like to expand both to avoid abbreviations, but we can do that in follow up work if you don't want to pick it up here.

hack/e2e.sh
```diff
@@ -0,0 +1,22 @@
...
+# * Processes the template to setup a local cincy instance
+# * Prepares a graph using test data
+# * Ensures generated graph is valid
+
+# Debug information
+oc whoami
+oc project
+
+# Apply oc template
+oc process ../dist/openshift/cincinnati.yaml \
```
LalatenduMohanty: We do not need `-f` for pointing to template file name?

vrutkovs: Fixed, thanks!

hack/e2e.sh
```diff
@@ -0,0 +1,22 @@
...
+# * Ensures generated graph is valid
+
+# Debug information
+oc whoami
+oc project
+
+# Apply oc template
+oc process -f ../dist/openshift/cincinnati.yaml \
+  -p IMAGE=${CINCY_E2E_IMAGE:-pipeline} \
+  -p IMAGE_TAG=${CINCY_E2E_IMAGE:-deploy} \
```
wking: nit: `CINCY` -> `CINCINNATI`.  I like "Cincy" when I'm in a hurry, but would rather avoid abbreviations in this write (hopefully) once, read-many source.

dist/openshift/cincinnati.yaml
vrutkovs: That would require updating existing params - and iirc these are used as is in internal deployment. Can be done in a new PR

hack/e2e.sh
vrutkovs: Fixed (also ensured `IMAGE_TAG` reads `CINCINNATI_E2E_IMAGE_TAG`)

hack/e2e.sh
```diff
@@ -0,0 +1,22 @@
...
+oc process -f ../dist/openshift/cincinnati.yaml \
+  -p IMAGE=${CINCINNATI_E2E_IMAGE:-pipeline} \
+  -p IMAGE_TAG=${CINCINNATI_E2E_IMAGE_TAG:-deploy} \
+  -p GB_CINCINNATI_REPO="redhat/openshift-cincinnati-test-public-manual"
+
+# Wait for dc to rollout
+oc wait --for=condition=available deploymentconfig/cincinnati
+
+# Check that policy engine returns channel data respond
+curl --verbose --header 'Accept:application/json' http://cincinnati-policy-engine:8080/v1/graph\?channel=a
```
wking: How will `http://cincinnati-policy-engine` work with parallel tests?  Won't we need this to use a dynamic, injected job-name or some such to avoid collisions?  Or is this magically running in the same Pod where the Cincinnati-under-test is running?

vrutkovs: Every test (in Prow) runs in separate namespaces. e2e pod would use the service from this namespace, as its running in the temp namespace too.

Perhaps base URL should be a config option so that e2e could be run on dev machines?

hack/e2e.sh
```diff
@@ -0,0 +1,22 @@
...
+
+# Debug information
+oc whoami
+oc project
+
+# Apply oc template
+oc process -f ../dist/openshift/cincinnati.yaml \
+  -p IMAGE=${CINCINNATI_E2E_IMAGE:-pipeline} \
+  -p IMAGE_TAG=${CINCINNATI_E2E_IMAGE_TAG:-deploy} \
+  -p GB_CINCINNATI_REPO="redhat/openshift-cincinnati-test-public-manual"
```
LalatenduMohanty, line 17: AFAIU you are trying to run the Cincy instance here, right? then `oc process` does not create an application from a template.

LalatenduMohanty (changes requested):

hack/e2e.sh
LalatenduMohanty: https://docs.openshift.com/container-platform/3.11/dev_guide/application_lifecycle/new_app.html#specifying-a-template

hack/e2e.sh
vrutkovs: Fixed URL (reads `CINCINNATI_PE_BASE_URL`) and PE port

hack/e2e.sh
vrutkovs: ah, good catch. Fixed by adding `| oc create -f`

wking: From those docs:

> To directly use a template in your local file system, without first storing it in OpenShift Container Platform, use the `-f|--file` argument...

which is what we have, right?

LalatenduMohanty (approved):
/lgtm

PR #193, opened 2020-01-10
OWNERS_ALIASES: Add rporres in app-sre team

rporres:


OWNERS_ALIASES
```diff
@@ -20,4 +20,5 @@ aliases:
     - skryzhny
     - tparikh
     - rrati
+    - rporres
```
LalatenduMohanty, line 23: We already have a lot of folks part of the app-sre approvers? Do we need `approve` access for everyone already in the file or we need to update the list? Do you need the access because you are going to work on Cincinnati?

LalatenduMohanty (changes requested):

OWNERS_ALIASES
wking: #169 has this team as a safety valve, but I'm not clear on why it needs to be large.  If app-sre wants to maintain a safety valve in case core Cincy maintainers are asleep, I think existing app-sre approvers should handle approval for PRs adjusting the set of app-sre approvers.

pbergene: It's the people participating in the on-call rotation, to ensure that an on-call responding would have the required access when responding out of hours.  Chances for doing changes are probably slim slim, but it needs to apply to the entire on-call team if it is to have value.

LalatenduMohanty: @pbergene how many folks will be in total for the on-call rotation?

wking: Bumping this thread since it went quiet.  @pbergene, you're already in the `app-sre` alias with approval power.  If you think @rporres needs to be included too, you or another `app-sre` approver can comment `/lgtm` on this PR to land it.

wking:
Doesn't seem to be much interest here.  I'm going to close it, but whenever app-sre approvers want to pick this back up, they can re-open and [approve it themselves][1].

/close

[1]: https://github.com/openshift/cincinnati/pull/193#discussion_r369880758

PR #197, opened 2020-01-22
commons/src/lib: Relax Accept header parsing

wking:
~WIP because I haven't actually relaxed anything yet ;).~ implemented-ish since e9b3d5c -> ddff926

Before this pull request:

```console
$ curl -isH 'Accept: */*' 'https://api.openshift.com/api/upgrades_info/v1/graph?channel=stable-4.2'
HTTP/1.1 406 Not Acceptable
Date: Sat, 18 Jan 2020 05:51:02 GMT
Server: Apache/2.4.6 (CentOS) OpenSSL/1.0.2k-fips
content-length: 72
content-type: application/json
Set-Cookie: ...; path=/; HttpOnly; Secure
Set-Cookie: ...; path=/; HttpOnly; Secure
{"kind":"invalid_content_type","value":"invalid Content-Type requested"}
```

and:

```console
$ curl -is 'https://api.openshift.com/api/upgrades_info/v1/graph?channel=stable-4.2'
HTTP/1.1 406 Not Acceptable
Date: Sat, 18 Jan 2020 05:51:11 GMT
Server: Apache/2.4.6 (CentOS) OpenSSL/1.0.2k-fips
content-length: 72
content-type: application/json
Set-Cookie: ...; path=/; HttpOnly; Secure
Set-Cookie: ...; path=/; HttpOnly; Secure
{"kind":"invalid_content_type","value":"invalid Content-Type requested"}
```

But both of those should be valid requests for "I accept all content types.  Server, pick your favorite" ([spec][1]).  This pull-request relaxes our media-type handling to support clients who set no constraints, or makes a request that includes `*`, `*/*`, `application/*`, or `application/json`.  This will make it easier for new Cincy clients to make successful Cincy requests.

[1]: https://tools.ietf.org/html/rfc7231#section-5.3.2

### DIFF
--- a/commons/src/lib.rs
+++ b/commons/src/lib.rs
@@ -91,21 +91,38 @@ pub fn ensure_query_params(
     Ok(())
 }
 
-/// Make sure client requested a valid content type.
-pub fn ensure_content_type(
+/// Make sure the client can accept the provided media type.
+pub fn validate_content_type(
     headers: &HeaderMap,
     content_type: &'static str,
 ) -> Result<(), GraphError> {
-    let content_json = header::HeaderValue::from_static(content_type);
+    let header_value = match headers.get(header::ACCEPT) {
+        None => return Ok(()),
+        Some(v) => v,
+    };
 
-    if !headers
-        .get(header::ACCEPT)
-        .map(|accept| accept == content_json)
-        .unwrap_or(false)
-    {
-        Err(GraphError::InvalidContentType)
-    } else {
+    let full_type = header::HeaderValue::from_static(content_type);
+    let wildcard = header::HeaderValue::from_static("*");
+    let double_wildcard = header::HeaderValue::from_static("*/*");
+    let top_type = content_type.split("/").next().unwrap_or("");
+    let top_type_wildcard = header::HeaderValue::from_str(&format!("{}/*", top_type));
+    assert!(
+        top_type_wildcard.is_ok(),
+        format!("could not form top-type wildcard from {}", top_type)
+    );
+
+    let acceptable_content_types: Vec<actix_web::http::HeaderValue> = vec![
+        full_type,
+        wildcard,
+        double_wildcard,
+        top_type_wildcard.unwrap(),
+    ];
+
+    // FIXME: this is not a full-blown Accept parser
+    if acceptable_content_types.iter().any(|c| c == header_value) {
         Ok(())
+    } else {
+        Err(GraphError::InvalidContentType)
     }
 }
 
@@ -152,10 +169,21 @@ mod tests {
     }
 
     #[test]
-    fn test_ensure_content_type() {
+    fn test_validate_content_type() {
         let mut headers = actix_web::http::HeaderMap::new();
-        headers.insert(header::ACCEPT, "application/json".parse().unwrap());
-        ensure_content_type(&headers, "application/json").unwrap();
-        ensure_content_type(&headers, "text/html").unwrap_err();
+        validate_content_type(&headers, "application/json").unwrap(); // if the request leaves Accept empty, we can return whatever we want
+        headers.insert(
+            header::ACCEPT,
+            //"application/json, text/*; q=0.2".parse().unwrap(), // prefer JSON, but also accept any text/* after an 80% markdown in quality.  FIXME: needs a smarter parser in validate_content_type
+            "application/json".parse().unwrap(),
+        );
+        validate_content_type(&headers, "application/json").unwrap();
+        headers.insert(
+            // FIXME: drop once validate_content_type gets a smarter parser and the previous insert can include the text/* entry
+            header::ACCEPT,
+            "text/*".parse().unwrap(),
+        );
+        validate_content_type(&headers, "text/plain").unwrap();
+        validate_content_type(&headers, "image/png").unwrap_err();
     }
 }
--- a/policy-engine/src/graph.rs
+++ b/policy-engine/src/graph.rs
@@ -45,7 +45,7 @@ pub(crate) async fn index(
     V1_GRAPH_INCOMING_REQS.inc();
 
     // Check that the client can accept JSON media type.
-    commons::ensure_content_type(req.headers(), CONTENT_TYPE)?;
+    commons::validate_content_type(req.headers(), CONTENT_TYPE)?;
 
     // Check for required client parameters.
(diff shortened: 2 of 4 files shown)


lucab:
The semantics you are trying to change is part of the protocol though: https://github.com/openshift/cincinnati/blob/master/docs/design/cincinnati.md#request.

wking:
> The semantics you are trying to change is part of the protocol ...

I don't think we should try to redefine standard HTTP constructs.  I'll reroll to bump those docs.

wking:
Rebased onto master with 7e8ef2f -> a7b4aa7.

>  I'll reroll to bump those docs.

Docs reroll happened back [here][1].

[1]: https://github.com/openshift/cincinnati/pull/197#event-2994648718

PratikMahajan:
lgtm

policy-engine/src/graph.rs
```diff
@@ -45,7 +45,7 @@ pub(crate) async fn index(
     V1_GRAPH_INCOMING_REQS.inc();
 
     // Check that the client can accept JSON media type.
-    commons::ensure_content_type(req.headers(), CONTENT_TYPE)?;
+    commons::validate_content_type(req.headers(), CONTENT_TYPE)?;
```
vrutkovs, line 48: Lets rework `CONTENT_TYPES` into a vector of accepted content-type regexps? This would make this a bit more explicit

commons/src/lib.rs
```diff
@@ -91,16 +91,34 @@ pub fn ensure_query_params(
...
 
-/// Make sure client requested a valid content type.
-pub fn ensure_content_type(
+/// Make sure the client can accept the provided media type.
+pub fn validate_content_type(
     headers: &HeaderMap,
     content_type: &'static str,
 ) -> Result<(), GraphError> {
-    let content_json = header::HeaderValue::from_static(content_type);
+    if !headers.contains_key(header::ACCEPT) {
```
vrutkovs, line 99: Not necessary - duplicated by line 133 (`get` return an `Option`, `map` would throw `Result` if its `None` and `unwrap_or(false)` would result into "Invalid content type" error)

commons/src/lib.rs
```diff
@@ -91,16 +91,34 @@ pub fn ensure_query_params(
...
+        return Ok(());
+    }
+
+    let full_type = header::HeaderValue::from_static(content_type);
+    let wildcard = header::HeaderValue::from_static("*");
+    let double_wildcard = header::HeaderValue::from_static("*/*");
+    let top_type = content_type.split("/").next().unwrap_or("");
```
vrutkovs, line 107: Not sure if I understand - if we pass `application/foo` it would still match as the top type matches? Why would we want that?

commons/src/lib.rs
```diff
@@ -91,16 +91,34 @@ pub fn ensure_query_params(
...
+    let top_type_wildcard = header::HeaderValue::from_str(&format!("{}/*", top_type));
+    assert!(
```
vrutkovs, line 109: we can unwrap this wildcard here and throw error if it cannot be constructed:
```
top_type_wildcard = match top_type_wildcard {
	Ok(t) => t,
	Err(_) => return Err(GraphError::InvalidContentType)
}
```
so that later it won't have to be unwrapped

commons/src/lib.rs
```diff
@@ -91,16 +91,34 @@ pub fn ensure_query_params(
...
+        top_type_wildcard.is_ok(),
+        format!("could not form top-type wildcard from {}", top_type)
+    );
 
     if !headers
```
vrutkovs: Build a vector of acceptable content types 

```suggestion
    let acceptable_content_types: Vec<&str> =
        vec![full_type, wildcard, double_wildcard, top_type_wildcard];
    let header_value = headers.get(header::ACCEPT);

	if acceptable_content_types.iter().any(|c| c == header_value) {
		Ok(())
	} else {
		Err(GraphError::InvalidContentType)
	}
```

commons/src/lib.rs
wking: `top_type` isn't used directly, it just feeds `top_type_wildcard`.  If some crazy person sends `Accept: application/*`, then we can serve them `application/json`.  If they send `Accept: application/foo`, we should 400 them.

policy-engine/src/graph.rs
wking: I'd rather feed it a list of types we were comfortable returning, and have `validate_content_type` handle as much of Accept header parsing as we expect to do, and return the accepted type.  At the moment, we only support returning `application/json`, so I'm fine passing in that single type string and not returning anything besides errors.  Can we punt passing in a list of types into a future if/when we support returning multiple types?

commons/src/lib.rs
wking: I wanted to panic here instead of returning a `GraphError`, because it's not the client's fault if we can't construct `top_type_wildcard`.  Is there a better way to distinguish between "dev(-ops) issue" and "client issue" besides `assert` panics?

commons/src/lib.rs
wking: done in ade9cb68 -> 73e8e06a

commons/src/lib.rs
vrutkovs: >it's not the client's fault if we can't construct top_type_wildcard

The only input for it is `content_type`, so its certainly client's fault if error occurs here

policy-engine/src/graph.rs
vrutkovs: Okay, sure thing

commons/src/lib.rs
wking: > The only input for it is `content_type`, so its certainly client's fault if error occurs here

Client is responsible for `headers`.  We are responsible for `content_type`:

```console
$ git --no-pager grep commons::ensure_content_type origin/master
origin/master:graph-builder/src/graph.rs:    commons::ensure_content_type(req.headers(), CONTENT_TYPE)?;
origin/master:policy-engine/src/graph.rs:    commons::ensure_content_type(req.headers(), CONTENT_TYPE)?;
$ git --no-pager grep 'const CONTENT_TYPE' origin/master
origin/master:cincinnati/src/lib.rs:pub const CONTENT_TYPE: &str = "application/json";
```

vrutkovs (approved):
/lgtm

PR #198, opened 2020-01-24
Adding justfile and adding instructions to make local development easier

LalatenduMohanty:
just is similar to make but it is simpler. It will greatly simplify developer workflows

https://github.com/casey/just

### DIFF
--- a/Justfile
+++ b/Justfile
@@ -1,0 +1,106 @@
+format:
+	cargo fmt --all
+
+commit +args="": format
+	git commit {{args}}
+
+build:
+	cargo build
+
+_coverage:
+	cargo kcov --verbose --all --no-clean-rebuild --open
+
+coverage: test _coverage
+
+test-pwd +args="":
+	#!/usr/bin/env bash
+	set -e
+	export RUST_BACKTRACE=0 RUST_LOG="graph-builder=trace,cincinnati=trace,dkregistry=trace"
+	pushd {{invocation_directory()}}
+	cargo test {{args}}
+
+test: format
+	#!/usr/bin/env bash
+	set -e
+	export RUST_BACKTRACE=1 RUST_LOG="graph-builder=trace,cincinnati=trace,dkregistry=trace"
+	cargo test --all
+
+_test component features='test-net,test-net-private' rustcargs='--ignored':
+	#!/usr/bin/env bash
+	set -xe
+	(pushd {{component}} && cargo test --features {{features}} -- {{rustcargs}})
+
+
+test-net-private:
+	#!/usr/bin/env bash
+	set -e
+	just _test quay "test-net,test-net-private" ""
+	just _test cincinnati "test-net,test-net-private" ""
+	just _test graph-builder "test-net,test-net-private" ""
+
+run-ci-tests:
+	#!/usr/bin/env bash
+	set -e
+	hack/run-all-tests.sh
+
+path_prefix = "api/upgrades_info/"
+
+
+get-and-display-graph:
+	just get-graph-gb | jq -cM | {{invocation_directory()}}/hack/graph.sh | dot -Tsvg > graph.svg; xdg-open graph.svg
+
+run-graph-builder registry="https://quay.io" repository="openshift-release-dev/ocp-release" credentials_file="${HOME}/.docker/config.json":
+	#!/usr/bin/env bash
+	export RUST_BACKTRACE=1
+	export RUST_LOG="graph_builder=trace,graph-builder=trace,cincinnati=trace"
+	# export RUST_LOG="${RUST_LOG},dkregistry=trace"
+	# strace -f -D -o gb.strace.lol
+	cargo run --package graph-builder -- --upstream.registry.pause_secs 30 --address 0.0.0.0 --registry {{registry}} --repository {{repository}} -vvv --service.path_prefix {{path_prefix}} --credentials-file {{credentials_file}} # --disable-quay-api-metadata
+
+
+run-graph-builder-satellite:
+	just run-graph-builder 'sat-r220-02.lab.eng.rdu2.redhat.com' 'default_organization-custom-ocp'
+
+run-policy-engine:
+	#!/usr/bin/env bash
+	export RUST_BACKTRACE=1 RUST_LOG="policy_engine=trace,cincinnati=trace,actix=trace,actix_web=trace"
+	cargo run --package policy-engine -- -vvvv --service.address 0.0.0.0 --service.path_prefix {{path_prefix}} --upstream.cincinnati.url 'http://localhost:8080/{{path_prefix}}v1/graph' --service.mandatory_client_parameters='channel'
+
+kill-daemons:
+	pkill graph-builder
+	pkill policy-engine
+
+run-daemons:
+	#!/usr/bin/env bash
+	just run-graph-builder #"https://quay.io" "redhat/openshift-cincinnati-test-public-manual" ~/.docker/config.json 2>&1 &
+	PG_PID=$!
+
+	just run-policy-engine 2>&1 &
+	PE_PID=$!
+
+	trap "kill $PG_PID $PE_PID" EXIT
+	while true; do sleep 30; done
+
+get-graph port channel arch host="http://localhost":
+	curl --header 'Accept:application/json' {{host}}:{{port}}/{{path_prefix}}v1/graph?channel='{{channel}}'\&arch='{{arch}}' | jq .
+
+get-graph-gb:
+	just get-graph 8080 "" ""
+
+get-graph-pe channel='' arch='':
+	just get-graph 8081 "{{channel}}" "{{arch}}"
+
+get-graph-pe-staging channel='stable-4.1' arch='amd64':
+	just get-graph 443 "{{channel}}" "{{arch}}" https://api.stage.openshift.com
+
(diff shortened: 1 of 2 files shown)


LalatenduMohanty:
The `justfile` is provided by @steveeJ and I added the docs.  I am planning to simplify it further and then update the Readme.

Justfile
```diff
@@ -0,0 +1,155 @@
...
+	curl --get https://quay.io/v2/{{repo}}/manifests/{{tag}} | jq .
+
+get-quay-labels tag repo='openshift-release-dev/ocp-release':
+	#!/usr/bin/env bash
+	set -xe
+	manifestref=`curl --head https://quay.io/v2/{{repo}}/manifests/{{tag}} | rg 'docker-content-digest: (sha256:[a-z0-9]{64}).*$' -r '$1'`
+	curl --get https://quay.io/api/v1/repository/{{repo}}/manifest/${manifestref}/labels | jq .
+
+
+testrg:
```
steveej: I think we can remove this recipe

Justfile
```diff
@@ -0,0 +1,155 @@
...
+	#!/usr/bin/env bash
+	set -xe
+	var=`echo 'hello\nrg\n' | rg '^(hello).*$' -r '$1'`
+
+_promquery query:
+	curl --get -g --header "Authorization: Bearer $(pass show Webservices/console.app-sre.openshift.com | jq -r '.token')" 'https://infogw-data.api.openshift.com/api/v1/query' --data-urlencode 'query={{query}}' --data-urlencode 'time=2019-03-08T14:45:34.000067500+00:00'
```
steveej: I think this is too specific to my system. Do you want to read the bearer token from the environment maybe?

steveej: Also, I think it's not of much use because of the hardcoded timestamp. Maybe we can remove the promquery recipes altogether. Opinions?

Justfile
```diff
@@ -0,0 +1,155 @@
...
+		avg by (version) (count by (version) (cluster_version{type="failure"}) / on (version) count by (version) (cluster_version)) \
+	'
+
+# count by (version) (cluster_version{type="failure"}) / on (version) count by (version) (cluster_version) \
+
+
+test-prometheus:
+	#!/usr/bin/env bash
+	set -e
+	export PROMETHEUS_API_TOKEN=`pass show Webservices/console.app-sre.openshift.com | jq -r '.token'`
```
steveej: Again a reference very specific to my setup. Here the API token is already referenced by variable in the command later, so this line can be removed.

docs/developer/developing.md
```diff
@@ -0,0 +1,49 @@
...
+TBD
+
+## Running tests locally
+
+To run unit tests
+
+```
+just test
+```
+
```
steveej, line 41: Looks like we also want `just test-net` without the private ones. Would you like to add that recipe and mention it here?

docs/developer/developing.md
```diff
@@ -0,0 +1,49 @@
...
+Here is the command to get graph for "stable-4.2" and for "amd64" architecture
+
+```
+just get-graph-pe "stable-4.2" "amd64"
+```
+
+## Building it locally
+
+TBD
```
steveej: What exactly do you mean by *it*? And do you mean just to build the binaries? Or a container image?

LalatenduMohanty: Both. Also planning to add the steps where we can build the image locally and push it to quay.io and then use it to run on an OpenShift instance.

docs/developer/developing.md
```diff
@@ -0,0 +1,49 @@
...
+
+* Git
+* Rust
+* [just](https://github.com/casey/just)
+
+*Note:* We recommend using [Rustup](https://github.com/rust-lang/rustup/blob/master/README.md) for Rust toolchain management. 
+
+## Running it locally
+
+Please below commands in different terminal windows
```
wking: Maybe:

> Run the following in parallel (for example, in different terminals).

docs/developer/developing.md
```diff
@@ -0,0 +1,49 @@
...
+* Git
+* Rust
+* [just](https://github.com/casey/just)
+
+*Note:* We recommend using [Rustup](https://github.com/rust-lang/rustup/blob/master/README.md) for Rust toolchain management. 
+
+## Running it locally
+
+Please below commands in different terminal windows 
+```
```
wking: This and others should be:

    ```console

to get console syntax highlighting because they include explicit `$` prompts.

docs/developer/developing.md
```diff
@@ -0,0 +1,49 @@
...
+$ just run-graph-bulider
```
wking: `bulider` -> `builder`

docs/developer/developing.md
```diff
@@ -0,0 +1,49 @@
...
+```
+
+```
+$ just run-policy-engine
+```
+
+Here is the command to get graph for "stable-4.2" and for "amd64" architecture
```
wking: "get graph" -> "get the graph".

And personally, I'd rather have `... to get the stable-4.2 amd64 graph` without the quotes and a reminder that amd64 was an architecture, but I'm fine if you want to keep the quotes and current order.

docs/developer/developing.md
```diff
@@ -0,0 +1,49 @@
...
+
+```
+just get-graph-pe "stable-4.2" "amd64"
```
wking, line 31: Earlier examples all included an explicit `$` prompt.  I like prompts, in which case we should have one here.  But it would also be internally consistent to remove all of the prompts and use:

    ```sh

Justfile
```diff
@@ -0,0 +1,155 @@
...
+
+get-graph-pe-production channel='stable-4.0' arch='amd64':
+	just get-graph 443 "{{channel}}" "{{arch}}" https://api.openshift.com
+
+get-openapi host port:
+	curl --header 'Accept:application/json' {{host}}:{{port}}/{{path_prefix}}v1/openapi | jq .
+
+get-openapi-staging:
+	just get-openapi https://api.stage.openshift.com 443
+	j
```
wking: typo?

wking:
I like POSIX and [make][1], so I didn't look over the `Justfile` too closely.  I'm fine with folks landing whatever they want in a `Justfile` though, if that makes their lives easier.

[1]: https://pubs.opengroup.org/onlinepubs/9699919799/utilities/make.html

Justfile
steveej: Most likely. It's also not in the version I have on my disk.

docs/developer/developing.md
LalatenduMohanty: I am going to make it shell.

docs/developer/developing.md
LalatenduMohanty: sure. My intention is to first get something working merged and then we can improved on this.

LalatenduMohanty: Also IMO we should be running `./hack/run-all-tests.sh` as part of the Justfile `just test`. we should be consistent with respect to what we run in CI vs on our laptop

steveej: We could have a recipe named `just run-ci-tests` which does that. I would like to keep a few recipes in there which ease development of specific components.

steveej:
/lgtm 
thanks! leaving @wking the opportunity to resolve his comments:
/hold

docs/developer/developing.md
```diff
@@ -0,0 +1,46 @@
...
+
+```shell
+just run-graph-builder
+```
+
+```shell
+just run-policy-engine
+```
+
+Here is the command to get the graph for stable-4.2 amd64 graph architecture
```
wking: Sorry if [my earlier recommendation][1] was unclear, but I think you can drop "architecture" here.  If you want to keep it, it should move closer to amd64 and not have "graph" in between.

[1]: https://github.com/openshift/cincinnati/pull/198#discussion_r370886260

docs/developer/developing.md
wking: nit: no need to quote stable-4.2 or amd64.  Neither contains shell-sensitive characters.

docs/developer/developing.md
```diff
@@ -0,0 +1,46 @@
...
+
+* Git
+* Rust
+* [just](https://github.com/casey/just)
+
+*Note:* We recommend using [Rustup](https://github.com/rust-lang/rustup/blob/master/README.md) for Rust toolchain management. 
+
+## Running it locally
+
+Please below commands in parallel (for e.g. in different terminals at the same time).
```
wking: nit: missing a verb.  Maybe "Please run the following commands..."?  Or "Please run the commands below ..."?  Although I personally [don't feel a need to say "please"][1], since folks can follow these directions or not without it bothering me either way ;).

[1]: https://github.com/openshift/cincinnati/pull/198#discussion_r370885704

docs/developer/developing.md
wking: continued in https://github.com/openshift/cincinnati/pull/198#discussion_r375427729

docs/developer/developing.md
wking: continued in https://github.com/openshift/cincinnati/pull/198#discussion_r375426001

docs/developer/developing.md
wking: Fixed in 1ed279c.

docs/developer/developing.md
wking: Fixed in 1ed279c.

PR #200, opened 2020-01-29
e2e improvements

vrutkovs:
This PR updates `hack/e2e.sh` script to setup latest Cincinnati version on a temporary Openshift cluster and runs e2e tests (currently just a simple curl)

https://issues.redhat.com/browse/OTA-103

### DIFF
--- a/hack/e2e.sh
+++ b/hack/e2e.sh
@@ -26,7 +26,9 @@ oc secrets link default ci-pull-secret --for=pull
 oc new-app -f dist/openshift/cincinnati.yaml \
   -p IMAGE="${IMAGE_FORMAT%/*}/stable" \
   -p IMAGE_TAG=deploy \
-  -p GB_CINCINNATI_REPO="redhat/openshift-cincinnati-test-public-manual"
+  -p GB_CINCINNATI_REPO="redhat/openshift-cincinnati-test-public-manual" \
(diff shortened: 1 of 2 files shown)


hack/e2e.sh
```diff
@@ -5,20 +5,52 @@
...
+  else
+    sleep ${DELAY}
+    ATTEMPTS=$((ATTEMPTS-1))
+  fi
+done
 
 # Check that policy engine returns channel data respond
-GRAPH_URL="http://${CINCINNATI_PE_BASE_URL:-cincinnati-policy-engine}:${CINCINNATI_PE_PORT:-8081}/v1/graph"
-curl --verbose --header 'Accept:application/json' "${GRAPH_URL}?channel=a"
+curl --header 'Accept:application/json' "${GRAPH_URL}"
```
steveej, line 56: Do we want to pipe it to `jq` so ensure it's valid JSON?

steveej (commented):
This looks like a very good start, thanks! We can postpone the more elaborate final graph verification if you prefer that.

hack/e2e.sh
vrutkovs: Test container doesn't have jq. I'd rather have a proper compiled test built here, using some kind of BDD testing - e.g. https://crates.io/crates/fluid or https://crates.io/crates/stainless2

PR #202, opened 2020-01-31
e2e: add cargo-based e2e test and run it after the instance is stood up

vrutkovs:
Add several tests which use `reqwest` to ensure returned test data graph is valid. These tests won't run during `cargo test`, instead they'd be a part of e2e suite

Part of OTA-103

### DIFF
--- a/graph-builder/tests/mod.rs
+++ b/graph-builder/tests/mod.rs
@@ -4,3 +4,6 @@ extern crate failure;
 
 #[cfg(feature = "test-net")]
 mod net;
+
+#[cfg(feature = "test-e2e")]
+mod e2e;
--- a/hack/e2e.sh
+++ b/hack/e2e.sh
@@ -38,14 +38,14 @@ oc wait --for=condition=available --timeout=5m deploymentconfig/cincinnati
 # Expose services
 oc expose service cincinnati-policy-engine --port=policy-engine
 PE_URL=$(oc get route cincinnati-policy-engine -o jsonpath='{.spec.host}')
-GRAPH_URL="http://${PE_URL}/api/upgrades_info/v1/graph?channel=a"
+export GRAPH_URL="http://${PE_URL}/api/upgrades_info/v1/graph"
 
 # Wait for route to become available
 ATTEMPTS=10
 DELAY=10
 
 while [ $ATTEMPTS -ge 0 ]; do
-  CODE=$(curl -s -o /dev/null -w "%{http_code}" --header 'Accept:application/json' "${GRAPH_URL}")
+  CODE=$(curl -s -o /dev/null -w "%{http_code}" --header 'Accept:application/json' "${GRAPH_URL}?channel=a")
   if [ "${CODE}" == "200" ]; then
     break
   else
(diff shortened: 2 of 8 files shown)


hack/e2e.sh
```diff
@@ -52,5 +52,6 @@ while [ $ATTEMPTS -ge 0 ]; do
   fi
 done
 
-# Check that policy engine returns channel data respond
-curl --header 'Accept:application/json' "${GRAPH_URL}"
+# Run e2e tests
+cd ../../graph-builder
+cargo test --features test-e2e e2e --no-default-features  -- --nocapture
```
vrutkovs: This would install test dependencies during the test run, which is suboptimal. I'll rework this in /release repo to ensure a test image is built first

steveej: I'm not sure what you mean by "install test dependencies", please explain.

vrutkovs: This test starts with a container built from `Dockerfile.builder`, so it has rustc and cargo installed, but doesn't have test prebuilt. So when `cargo test` is run here it would download crates and compile test - [log](https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_cincinnati/202/pull-ci-openshift-cincinnati-master-e2e/23/artifacts/e2e/container-logs/test.log)

I'll update this to have test built along with other artifacts in follow up PRs

graph-builder/tests/mod.rs
```diff
@@ -4,3 +4,12 @@ extern crate failure;
 
 #[cfg(feature = "test-net")]
 mod net;
+
+#[cfg(feature = "test-e2e")]
+#[macro_use]
+extern crate assert_json_diff;
```
steveej: This is not needed. Instead you can write `use assert_json_diff::name_of_the_macro` in the mod where the macro should be used.

graph-builder/tests/mod.rs
```diff
@@ -4,3 +4,12 @@ extern crate failure;
...
+#[cfg(feature = "test-e2e")]
+extern crate serde_json;
```
steveej: This is not needed.

steveej (changes requested):
Just a few nitpicks in the comments.

Extra note:
I'm not a fan of our using and with this PR duplicating the expected hashes from the manually maintained test repositories, see e.g. [here](https://github.com/openshift/cincinnati/blob/master/graph-builder/tests/net/quay_io/mod.rs#L258-L259). A first step is to have a common definition of these images and their expected hashes somewhere in shared module. We can postpone this to a future PR but I did want to note this down somewhere.

graph-builder/tests/mod.rs
vrutkovs: good catch, fixed

graph-builder/tests/mod.rs
vrutkovs: Fixed

steveej (approved):
/lgtm

Thanks!

I'm not sure if you wanted to do something about https://github.com/openshift/cincinnati/pull/202#discussion_r373414878 in this PR or postpone all of the things you mentioned, so:

/hold

vrutkovs:
>I'm not sure if you wanted to do something about #202 (comment) in this PR or postpone all of the things you mentioned, so:

This needs to happen in two repos - first proper images built in openshift/release then these images would be used in cincinnati's e2e.

I'll keep it on hold until e2e passes just to be sure

vrutkovs:
e2e passed \o/

/hold cancel

vrutkovs:
/override ci/prow/rustfmt-1.38.0

Outdated test

PR #203, opened 2020-01-31
dist/Dockerfile.deploy/Dockerfile: install openssl

vrutkovs:
Update existing packages and install openssl, since binaries depend on
it.

e2e images don't fail since they are using src container as a base (this would be changed later)


steveej (approved):
This is fine for now. Eventually we need to use something more officially maintained as a base container with all these goodies.

PR #206, opened 2020-02-03
 graph-builder/registry: asynchronously assemble image metadata

steveej:
Working towards having the tag processing being completely asynchronous.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -323,6 +323,44 @@ dependencies = [
 ]
 
 [[package]]
+name = "async-compression"
+version = "0.2.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "bytes 0.5.3 (registry+https://github.com/rust-lang/crates.io-index)",
+ "flate2 1.0.13 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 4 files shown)


aditya-konarde:
Is there a way to get ourselves excluded from the reviewers lottery that prow does? :)

vrutkovs:
@aditya-konarde sorry, app-sre have been added in https://github.com/openshift/cincinnati/pull/169 so that you could merge release-critical PRs.

I don't think there is a simple way to have you excluded, I'll see what we can do

vrutkovs:
https://github.com/openshift/cincinnati/pull/207 should fix e2e - nodes would now be sorted by version. @steveeJ could you check if cherrypicking this commit does the trick?

steveej:
> #207 should fix e2e

The e2e test failure looks unrelated to ordering, and I wouldn't expect this PR to change the order. It merely helps to not block the runtime thread this code runs in.

vrutkovs:
/retest

e2e cluster didn't start properly

vrutkovs (approved):
/lgtm

vrutkovs:
/hold

so that @LalatenduMohanty could have another look

PR #207, opened 2020-02-04
e2e: sort received keys by version

vrutkovs:
Cincinnati may return unsorted list of nodes if tags are processed
in parallel.

e2e tests would now sort nodes by version before comparison

### DIFF
--- a/graph-builder/tests/e2e/mod.rs
+++ b/graph-builder/tests/e2e/mod.rs
@@ -6,6 +6,20 @@ use std::env;
 use test_case::test_case;
 use url::Url;
 
+pub fn sort_by_version(v: &mut Value) {
+    if !v.is_object() {
+        return;
+    }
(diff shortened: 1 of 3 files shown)


steveej (changes requested):
I'd also apply the sorting to the testdata so the test doesn't rely on manually sorting it.

vrutkovs:
Added a commit which shuffles nodes so that we'd ensure at least expected in being sorted correctly

steveej (changes requested):
The current sorting only takes care of the nodes but it doesn't adjust the edges indices, so in cases where the sorting actually changes the order it'll break the consistency.

vrutkovs:
/hold

This doesn't sort `edges` so in case of async cincinnati these would be different as well

vrutkovs:
Included in #174

PR #211, opened 2020-02-05
Dockerfile.deploy: use multistage docker build

vrutkovs:
Build `deploy` container using builder and artifacts stages to make it slim

### DIFF
--- a/dist/Dockerfile.build/Dockerfile
+++ b/dist/Dockerfile.build/Dockerfile
@@ -32,5 +32,3 @@ RUN \
   mkdir -p $HOME/.cargo/git/ && \
   find $HOME/. -type d -exec chmod 777 {} \; && \
   find $HOME/. -type f -exec chmod ugo+rw {} \;
-
-WORKDIR /go/src/github.com/openshift/cincinnati
--- a/dist/Dockerfile.deploy/Dockerfile
+++ b/dist/Dockerfile.deploy/Dockerfile
@@ -1,3 +1,11 @@
+FROM cincinnati:build as builder
+# build
+COPY . .
+RUN cargo build --release && \
+    mkdir -p /opt/cincinnati/bin && \
+    cp -rvf target/release/graph-builder /opt/cincinnati/bin && \
+    cp -rvf target/release/policy-engine /opt/cincinnati/bin
+
 FROM centos:7
 
 ENV RUST_LOG=actix_web=error,dkregistry=error
@@ -6,6 +14,6 @@ RUN yum update -y && \
     yum install -y openssl && \
     yum clean all
 
-COPY graph-builder policy-engine /usr/bin/
+COPY --from=builder /opt/cincinnati/bin/* /usr/bin/
 
 ENTRYPOINT ["/usr/bin/graph-builder"]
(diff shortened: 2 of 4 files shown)


LalatenduMohanty:
Does podman supports multi-stage builds yet. I will be interested  build it with podman as it is more sustainable in long term for us.

vrutkovs:
podman supports multistage builds - and so does imagebuilder used in OCP

dist/Dockerfile.build/Dockerfile
```diff
@@ -32,3 +32,5 @@ RUN \
   mkdir -p $HOME/.cargo/git/ && \
   find $HOME/. -type d -exec chmod 777 {} \; && \
   find $HOME/. -type f -exec chmod ugo+rw {} \;
+
+WORKDIR /go/src/github.com/openshift/cincinnati
```
steveej: Is this GOPATH structure as the CI is built mainly for Go packages, and it clones the source accordingly?

dist/Dockerfile.deploy/Dockerfile
```diff
@@ -1,3 +1,11 @@
+FROM cincinnati:build as builder
+# build
+COPY . .
+RUN cargo build --release && \
+    mkdir -p /opt/cincinnati/bin && \
+    cp -rvf target/release/graph-builder /opt/cincinnati/bin && \
+    cp -rvf target/release/policy-engine /opt/cincinnati/bin && \
+
```
steveej, line 8: Either
```suggestion
    :
```
or remove the `&& \` in the previous line

vrutkovs: Yep, fixed, thanks

dist/Dockerfile.build/Dockerfile
vrutkovs: Yes, that's the path CI would put source at. I don't think its required though - removed

steveej:
/lgtm
Not sure about the timing for merging this PR, so:
/hold

hack/e2e.sh
```diff
@@ -23,12 +23,10 @@ oc create secret generic ci-pull-secret --from-file=.dockercfg=/tmp/cluster/pull
 oc secrets link default ci-pull-secret --for=pull
 
 # Apply oc template
-oc new-app -f dist/openshift/cincinnati.yaml \
+oc new-app -f ../template/cincinnati.yaml \
```
steveej: It's not clear to me why this path is relative to `hack/`

steveej (approved):
/lgtm
one question though:
/hold

hack/e2e.sh
vrutkovs: This is set by `WORKDIR` in dockerfile:
```
WORKDIR "${HOME}/cincinnati"
...
COPY --from=builder hack/e2e.sh hack/
COPY --from=builder dist/openshift/cincinnati.yaml template/
```

vrutkovs: Ah, I was wrong - `hack/e2e.sh` would not change pwd to `hack`. Fixed

vrutkovs:
/hold cancel

e2e test is now passing!

PR #212, opened 2020-02-06
e2e: use multistage dockerfile to prepare e2e image with artifacts

vrutkovs:
This ensures e2e image we're running would be slimmer, since tests are
being built in a separate stage.

Its safe to merge now, since its not yet used by e2e tests

### DIFF
--- a/dist/Dockerfile.e2e/Dockerfile
+++ b/dist/Dockerfile.e2e/Dockerfile
@@ -1,3 +1,14 @@
+FROM cincinnati:build as builder
+# build e2e tests
+COPY . .
+
+RUN cd graph-builder && \
+    cargo build --tests --features test-e2e --no-default-features && \
+    mkdir -p /opt/cincinnati/bin && \
+    cp -rvf ../target/debug/mod-* /opt/cincinnati/bin && \
+    rm -rf /opt/cincinnati/bin/mod-*.d && \
+    mv /opt/cincinnati/bin/mod-* /opt/cincinnati/bin/cincinnati-e2e-test
+
 FROM centos:7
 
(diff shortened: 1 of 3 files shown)


wking:
I like slim production images.  ~Do we care about slim e2e images?~ edit: dropping the dynamic rustup curl is enough for me ;).

dist/Dockerfile.e2e/Dockerfile
```diff
@@ -9,6 +16,9 @@ WORKDIR "${HOME}/cincinnati"
 RUN mkdir -p ${HOME}/bin && \
     curl https://mirror.openshift.com/pub/openshift-v4/clients/oc/latest/linux/oc.tar.gz 2>/dev/null | tar xzf - -C "${HOME}/bin/" oc
 ENV PATH="${PATH}:${HOME}/bin"
-COPY . .
+COPY --from=builder /go/src/github.com/openshift/cincinnati/target/debug/mod-* /usr/bin/
```
wking: Can you use `/usr/bin/cincinnati-e2e-test` here and skip the later `mv`?

vrutkovs:
>dropping the dynamic rustup curl is enough for me ;).

yeah, this mostly. Also we won't spin up the cluster if the test fails to build

dist/Dockerfile.e2e/Dockerfile
vrutkovs: Fixed

vrutkovs:
/retest

Cluster timeout error

vrutkovs:
Fails `e2e` since CI is not using `Dockerfile.e2e` yet

PR #213, opened 2020-02-07
template: update GB mem limit

vrutkovs:
Stage pod get killed with OOMKilled, lets see if we fit into 512MB

### DIFF
--- a/dist/openshift/cincinnati.yaml
+++ b/dist/openshift/cincinnati.yaml
@@ -229,9 +229,9 @@ parameters:
     displayName: cincinnati version
     description: cincinnati version which defaults to latest
   - name: GB_MEMORY_LIMIT
-    value: "256Mi"
+    value: "512Mi"
     displayName: "Graph-builder memory limit"
-    description: "Maximum amount of memory (bytes) allowed for graph-builder (default: 256Mi)"
+    description: "Maximum amount of memory (bytes) allowed for graph-builder (default: 523Mi)"
   - name: GB_CPU_LIMIT
     value: "750m"
     displayName: "Graph-builder CPU limit"


dist/openshift/cincinnati.yaml
```diff
@@ -229,9 +229,9 @@ parameters:
     displayName: cincinnati version
     description: cincinnati version which defaults to latest
   - name: GB_MEMORY_LIMIT
-    value: "256Mi"
+    value: "512Mi"
```
LalatenduMohanty, line 232: I am interested to find out why we need this bump. Is this expected from https://github.com/openshift/cincinnati/pull/174

LalatenduMohanty (changes requested):

dist/openshift/cincinnati.yaml
LalatenduMohanty: If we change the currency level to 8 can it be done with in 256Mi. @steveeJ ^

vrutkovs: in order to find out what's chewing RAM we need to stop OOMKills first

LalatenduMohanty: https://github.com/openshift/cincinnati/pull/174#discussion_r375927706 so the new cache needs more memory.

LalatenduMohanty (approved):
/lgtm

vrutkovs:
/hold

Lets see if e2e passes

dist/openshift/cincinnati.yaml
steveej: #174 doesn't increase the final cache size, it just fills it up concurrently which uses more memory. I'm curious if it fits into 512MiB as well, let's try it.

steveej (approved):
/lgtm 
as well

LalatenduMohanty:
> Lets see if e2e passes

Without e2e passing the `lgtm` bot will not merge the PR right?

vrutkovs:
>Without e2e passing the lgtm bot will not merge the PR right?

The bot will merge once all required tests have passed. e2e is not set to required yet

steveej:
> The bot will merge once all required tests have passed. e2e is not set to required yet

It will not merge unless we release the hold though ;-) We are waiting for the e2e tests to pass before we do that, correct?

PR #215, opened 2020-02-11
Changing the HTTP response of readiness probe to 503 when not available

LalatenduMohanty:

### DIFF
--- a/graph-builder/src/status.rs
+++ b/graph-builder/src/status.rs
@@ -8,12 +8,12 @@ use failure::Fallible;
 ///
 /// Status:
 ///  * Live (200 code): The upstream scrape loop thread is running
-///  * Not Live (500 code): everything else.
+///  * Not Live (503 code): everything else.
 pub async fn serve_liveness(app_data: actix_web::web::Data<State>) -> Fallible<HttpResponse> {
     let resp = if app_data.is_live() {
         HttpResponse::Ok().finish()
     } else {
-        HttpResponse::InternalServerError().finish()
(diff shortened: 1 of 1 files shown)


graph-builder/src/status.rs
```diff
@@ -23,12 +23,12 @@ pub async fn serve_liveness(app_data: actix_web::web::Data<State>) -> Fallible<H
 ///
 /// Status:
 ///  * Ready (200 code): a JSON graph as the result of a successful scrape is available.
-///  * Not Ready (500 code): no JSON graph available yet.
+///  * Service Unavailable (503 code): no JSON graph available yet.
```
lucab: The docstring is meant to describe the semantics of the result ("ready" or "not ready"), not the meaning of the HTTP code.

vrutkovs (changes requested):
Needs to update readiness check - and same applied to policy-engine

graph-builder/src/status.rs
steveej: To add to Luca's comment:

```suggestion
///  * Not Ready (503 code): no JSON graph available yet.
```

LalatenduMohanty:
> Needs to update readiness check - and same applied to policy-engine

I do not think I understand. Can you please elaborate?

vrutkovs:
>I do not think I understand. Can you please elaborate?

This PR updates only one check - readiness check in graph-builder. It should also update status returned by readiness check there - and update policy-engine binary too

vrutkovs (approved):

PR #216, opened 2020-02-11
e2e: set RUST_BACKTRACE in e2e tests and handle missing git commit data

vrutkovs:
* Update template to support RUST_BACKTRACE=1 (set to 0 by default) for
both containers
* Install `git` in Dockerfile.build so that `built_info::GIT_VERSION` would be set
* Handle missing commit info by setting "unknown" version

### DIFF
--- a/dist/Dockerfile.deploy/Dockerfile
+++ b/dist/Dockerfile.deploy/Dockerfile
@@ -1,6 +1,12 @@
 FROM quay.io/app-sre/cincinnati:builder AS builder
+
+WORKDIR /go/src/github.com/openshift/cincinnati/
+
 # build
 COPY . .
+# copy git information for built crate
+COPY .git/ ./.git/
+
 RUN cargo build --release && \
     mkdir -p /opt/cincinnati/bin && \
     cp -rvf target/release/graph-builder /opt/cincinnati/bin && \
--- a/dist/openshift/cincinnati.yaml
+++ b/dist/openshift/cincinnati.yaml
@@ -59,6 +59,11 @@ objects:
                     configMapKeyRef:
                       key: gb.log.verbosity
                       name: cincinnati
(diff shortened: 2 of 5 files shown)


vrutkovs:
>level=error msg="Error: Provider produced inconsistent result after apply"

/retest

dist/openshift/cincinnati.yaml
```diff
@@ -131,6 +136,11 @@ objects:
                     configMapKeyRef:
                       key: pe.mandatory_client_parameters
                       name: cincinnati
+                - name: "RUST_BACKTRACE"
+                  valueFrom:
+                    configMapKeyRef:
+                      key: pe.rust_backtrace
+                      name: cincinnati
```
LalatenduMohanty, line 143: This is temporary right?

vrutkovs:
Returns "unknown" build commit

dist/Dockerfile.deploy/Dockerfile
```diff
@@ -1,6 +1,9 @@
 FROM quay.io/app-sre/cincinnati:builder AS builder
 # build
 COPY . .
+# copy git information for built crate
+COPY .git/ ./.git/
```
LalatenduMohanty, line 8: Why is this required?  because when you copy a git repository in  the .git directory should get copied correctly in to `../.git/`.

vrutkovs: `COPY . .` doesn't copy hidden files or dirs

dist/openshift/cincinnati.yaml
vrutkovs: This is required for PR e2e. This won't be set on stage or prod

vrutkovs:
/hold cancel

e2e fixed

PR #218, opened 2020-02-12
Dockerfiles: use uppercase AS so that CI would replace these correctly

vrutkovs:
Imagebuilder doesn't recognize alias without this apparently. See rehearse results in https://github.com/openshift/release/pull/7133


steveej (approved):
This is surprising :man_shrugging: 

/lgtm

PR #219, opened 2020-02-12
Dockerfile.deploy: don't run `cargo build` if binaries are prepared

vrutkovs:
Internal builder is running `cargo build --release` outside of the container
and copies then to docker context. This ensures deploy build won't attempt
to run cargo build if the binaries (at least graph-builder) are already
present

### DIFF
--- a/dist/Dockerfile.deploy/Dockerfile
+++ b/dist/Dockerfile.deploy/Dockerfile
@@ -1,8 +1,8 @@
 FROM local/cincinnati-build:latest AS builder
 # build
 COPY . .
-RUN cargo build --release && \
-    mkdir -p /opt/cincinnati/bin && \
+RUN if [ ! -z target/release/graph-builder ]; then cargo build --release; fi
+RUN mkdir -p /opt/cincinnati/bin && \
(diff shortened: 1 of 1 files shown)


dist/Dockerfile.deploy/Dockerfile
```diff
@@ -1,8 +1,8 @@
 FROM local/cincinnati-build:latest AS builder
 # build
 COPY . .
-RUN cargo build --release && \
-    mkdir -p /opt/cincinnati/bin && \
+RUN if [ ! -z target/release/graph-builder ]; then cargo build --release; fi
```
steveej, line 4: How does

```
-z STRING
       the length of STRING is zero
 ``` 
work here? I'd expect 
```
-f FILE
    FILE exists and is a regular file
```

_(both are from `man test`)_

So maybe

```suggestion
RUN [ -f target/release/graph-builder ] || cargo build --release
```
?

vrutkovs:
No longer needed, https://github.com/openshift/cincinnati/pull/220 is the right way to solve this

PR #220, opened 2020-02-12
fix stage build after multistage rework

vrutkovs:
This is now happening in `Dockerfile.deploy` automatically

### DIFF
--- a/dist/Dockerfile.deploy/Dockerfile
+++ b/dist/Dockerfile.deploy/Dockerfile
@@ -1,4 +1,4 @@
-FROM quay.io/app-sre/cincinnati:base AS builder
+FROM quay.io/app-sre/cincinnati:builder AS builder
 # build
 COPY . .
 RUN cargo build --release && \
--- a/dist/Dockerfile.deploy/Dockerfile
+++ b/dist/Dockerfile.deploy/Dockerfile
@@ -1,4 +1,4 @@
-FROM local/cincinnati-build:latest AS builder
+FROM quay.io/app-sre/cincinnati:base AS builder
 # build
 COPY . .
 RUN cargo build --release && \
--- a/dist/build_deploy.sh
+++ b/dist/build_deploy.sh
@@ -1,5 +1,7 @@
 #!/usr/bin/env bash
 
+# TODO [vrutkovs]: rework this as Prow CI doesn't use it anymore 
+
 set -ex
 
 ABSOLUTE_PATH="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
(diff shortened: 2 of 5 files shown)


dist/Dockerfile.deploy/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM local/cincinnati-build:latest AS builder
+FROM quay.io/app-sre/cincinnati:base AS builder
```
wking: If we're going to pick a dummy name for the build container, can we use `quay.io/app-sre/cincinnati:builder`?  I think of `base` as the next image into which the built binaries are copied.

dist/build_deploy.sh
```diff
@@ -26,10 +26,6 @@ function cleanup() {
 }
 trap cleanup EXIT
 
-docker_cargo cargo build --release
-mkdir $RELEASE_OUTPUT_DIR
-cp ${RELEASE_DIR}/{graph-builder,policy-engine} $DOCKERFILE_DEPLOY  $RELEASE_OUTPUT_DIR/
```
wking, line 31: Does this mean we want to drop the other `DOCKERFILE_DEPLOY` references from this script?  It's not clear to me how it fits into the app-sre build pipeline.  Does it end up just being `dist/build.sh`?  Which is currently a wrapper of some sort around `build_deploy.sh`?  I'm fuzzy on how all of this is supposed to fit together.

vrutkovs: >Which is currently a wrapper of some sort around build_deploy.sh? I'm fuzzy on how all of this is supposed to fit together.

Previously we had a Jenkins job which ran `dist/build.sh`, it has been unsetting quay env vars so that the built image would not be pushed. The job has been removed so now `build_deploy.sh` is the only script used and it has some dead code.

I'd rather rework the internal CI procedure in a different PR really

dist/Dockerfile.deploy/Dockerfile
vrutkovs: Makes sense, done

dist/build_deploy.sh
steveej: > I'd rather rework the internal CI procedure in a different PR really

Please add a comment to the script which code is stale.

dist/Dockerfile.deploy/Dockerfile
steveej: Looks like the name here is still `base` as opposed to @wking's suggestion `builder`.

steveej (approved):
/lgtm

vrutkovs:
/skip
/hold cancel

e2e tests are still mysteriously failing

PR #222, opened 2020-02-13
build_deploy.sh: don't create temporary release dir

vrutkovs:
Its being handled by multistage build now. This fixes stage build errors

### DIFF
--- a/dist/build_deploy.sh
+++ b/dist/build_deploy.sh
@@ -11,24 +11,8 @@ DOCKERFILE_BUILD="$ABSOLUTE_PATH/Dockerfile.build/"
 ensure_build_container "${DOCKERFILE_BUILD}"
 
 DOCKERFILE_DEPLOY="$ABSOLUTE_PATH/Dockerfile.deploy/Dockerfile"
-RELEASE_DIR="${PROJECT_PARENT_DIR}/target/release"
-RELEASE_OUTPUT_DIR="${PROJECT_PARENT_DIR}/release-$(date +'%Y%m%d.%H%M%S')"
 
-function cleanup() {
-    set +e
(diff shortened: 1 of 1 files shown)


dist/build_deploy.sh
```diff
@@ -11,24 +11,8 @@ DOCKERFILE_BUILD="$ABSOLUTE_PATH/Dockerfile.build/"
...
-        docker_cargo cargo clean
-    fi
-    docker_cargo_stop_all
-    if [[ ! -n "$KEEP_RELEASE_OUTPUT" ]]; then
-        rm -f ${RELEASE_OUTPUT_DIR}/{graph-builder,policy-engine}
-        rm -f ${RELEASE_OUTPUT_DIR}/$(basename ${DOCKERFILE_DEPLOY})
-        rmdir ${RELEASE_OUTPUT_DIR}
-    fi
-}
-trap cleanup EXIT
```
steveej, line 29: Could you just remove this line and leave the cleanup function in place? It would be nice if we could just disable the trap on CI builds, because it's still useful for people running the testsuite locally in the container environment.

vrutkovs: it won't work, there is nothing to clean up because we don't create intermediate dir with binaries and dockerfile. This happens in builder stage of `Dockerfile.deploy`

LalatenduMohanty (approved):
/lgtm

dist/build_deploy.sh
LalatenduMohanty: Because we are using a multi-stage build.

steveej: I'm not convinced but I'd like to get the builds working again, so I'll take a  look at the end-result :man_shrugging:

steveej (approved):
/lgtm

PR #224, opened 2020-02-14
graph-builder: move registry into an internal plugin

steveej:
Moving the container registry scraper functionality into a plugin gives
very high flexibility to swap the release metadata out for something
else very easily. It also allows it to be reused by any code which knows
how to execute internal cincinnati plugins. To not make this change more
invasive, I decided to leave the code of the new plugin in the
graph-builder crate as opposed to moving it to the cincinnati crate.

Other notable changes:

* Changed metrics semantics and organization

  Because code which lives in the new plugin responsible for the
  `graph_upstream_raw_releases` metric, I found it necessary to
  introduce a mechanism to help ensure that certain metrics are
  registered after the plugin chain is loaded.
  This mechanism can be refined in the future, and could possibly be
  turned into a contract API between the main service and the plugins,
  which could specify what metrics are handled by which plugin.

* Moved the graph-builder/tests/net tests to the plugin

  Because these tests are all specific to fetching releases from a
  container registry, they belong to this plugin now. The registry tests
  were moved verbatim and corrected to match its environmment. One
  exception tot his is `create_graph_public_multiarch_manual_succeeds`,
  which also uses the higher-level `create_graph` function and was
  therefore renamed to `scrape_public_multiarch_manual_succeeds` and
  refactored to test the outcome of the plugin introduced in this
  change.

* Reverted async metadata processing

  Unfortunately it was necessary to revert recent code which used an async
  interface for the metadata assembly. The reason is that even though the
  code is async, it doesn't fulfill the `Sync + Send` trait bound required
  by the cincinnati plugin interface. For now, the synchronous metadata
  assembly is spawned using tokio's spawn_blocking [0], which offloads
  blocking task to a dedicated thread-pool.

  > Revert "Merge pull request #206 from steveeJ-forks/pr/gb-parse-metadata-async"

  > This reverts commit 3780abe,
  > reversing changes made to f957a5f.

  [0]: https://docs.rs/tokio/0.2.11/tokio/task/fn.spawn_blocking.html

### DIFF
--- a/graph-builder/src/config/mod.rs
+++ b/graph-builder/src/config/mod.rs
@@ -12,3 +12,6 @@ mod options;
 mod settings;
 
 pub use self::settings::AppSettings;
+
+/// Common prefix for graph-builder metrics.
+pub const METRICS_PREFIX: &str = "cincinnati_gb";
--- a/graph-builder/src/config/settings.rs
+++ b/graph-builder/src/config/settings.rs
@@ -37,11 +37,11 @@ pub struct AppSettings {
 
     // TODO(lucab): split this in (TLS, hostname+port).
     /// Target host for the registry scraper.
-    #[default("http://localhost:5000")]
+    #[default(crate::plugins::release_scrape_dockerv2::DEFAULT_SCRAPE_REGISTRY.to_string())]
     pub registry: String,
 
     /// Target image for the registry scraper.
-    #[default("openshift")]
+    #[default(crate::plugins::release_scrape_dockerv2::DEFAULT_SCRAPE_REPOSITORY.to_string())]
     pub repository: String,
 
     /// Listening address for the status service.
@@ -62,8 +62,15 @@ pub struct AppSettings {
     pub disable_quay_api_metadata: bool,
 
     /// Concurrency for graph fetching
-    #[default(16)]
+    #[default(crate::plugins::release_scrape_dockerv2::DEFAULT_FETCH_CONCURRENCY)]
     pub fetch_concurrency: usize,
+
+    /// Metrics which are required to be registered, to be specified without the `METRICS_PREFIX`.
+    /// If these are not registered by the time all plugins have been loaded an error will be thrown.
+    #[default([
+        "graph_upstream_raw_releases",
+    ].iter().cloned().map(Into::into).collect())]
+    pub metrics_required: HashSet<String>,
 }
 
 impl AppSettings {
--- a/graph-builder/src/graph.rs
+++ b/graph-builder/src/graph.rs
@@ -14,10 +14,9 @@
 
 use crate::built_info;
 use crate::config;
-use crate::registry::{self, Registry};
 use actix_web::{HttpRequest, HttpResponse};
 use cincinnati::plugins::prelude::*;
-use cincinnati::{AbstractRelease, Graph, Release, CONTENT_TYPE};
+use cincinnati::{AbstractRelease, Graph, CONTENT_TYPE};
 use commons::metrics::HasRegistry;
 use commons::GraphError;
 use failure::{Error, Fallible};
@@ -38,11 +37,6 @@ lazy_static! {
     static ref GRAPH_LAST_SUCCESSFUL_REFRESH: IntGauge = IntGauge::new(
         "graph_last_successful_refresh_timestamp",
         "UTC timestamp of last successful graph refresh"
-    )
-    .unwrap();
-    static ref GRAPH_UPSTREAM_RAW_RELEASES: IntGauge = IntGauge::new(
-        "graph_upstream_raw_releases",
-        "Number of releases fetched from upstream, before processing"
     )
     .unwrap();
     static ref UPSTREAM_ERRORS: Counter = Counter::new(
@@ -90,7 +84,6 @@ pub fn register_metrics(registry: &prometheus::Registry) -> Fallible<()> {
     commons::register_metrics(&registry)?;
     registry.register(Box::new(GRAPH_FINAL_RELEASES.clone()))?;
     registry.register(Box::new(GRAPH_LAST_SUCCESSFUL_REFRESH.clone()))?;
-    registry.register(Box::new(GRAPH_UPSTREAM_RAW_RELEASES.clone()))?;
     registry.register(Box::new(UPSTREAM_ERRORS.clone()))?;
     registry.register(Box::new(UPSTREAM_SCRAPES.clone()))?;
     registry.register(Box::new(GRAPH_UPSTREAM_INITIAL_SCRAPE.clone()))?;
@@ -170,17 +163,6 @@ impl HasRegistry for State {
 
 #[allow(clippy::useless_let_if_seq)]
 pub async fn run(settings: &config::AppSettings, state: &State) -> ! {
-    // Grow-only cache, mapping tag (hashed layers) to optional release metadata.
-    let mut cache = registry::cache::new();
-
-    let registry = Registry::try_from_str(&settings.registry)
-        .unwrap_or_else(|_| panic!("failed to parse '{}' as Url", &settings.registry));
-
-    // Read the credentials outside the loop to avoid re-reading the file
-    let (username, password) =
-        registry::read_credentials(settings.credentials_path.as_ref(), &registry.host)
-            .expect("could not read registry credentials");
-
     // Indicate if a panic happens
     let previous_hook = std::panic::take_hook();
     let panic_live = state.live.clone();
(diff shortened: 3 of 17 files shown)


graph-builder/src/config/settings.rs
```diff
@@ -37,11 +37,11 @@ pub struct AppSettings {
 
     // TODO(lucab): split this in (TLS, hostname+port).
```
vrutkovs, line 38: Not sure if this is still worth implementing, but this PR would be a good place for this

graph-builder/src/plugins/release_scrape_dockerv2/registry.rs
```diff
@@ -193,10 +181,10 @@ pub async fn fetch_releases(
     repo: &str,
     username: Option<&str>,
     password: Option<&str>,
-    cache: &mut crate::registry::cache::Cache,
+    cache: &mut cache::Cache,
```
vrutkovs: Could `cache::CacheAsync` be used here from the start?

graph-builder/src/plugins/release_scrape_dockerv2/registry.rs
```diff
@@ -330,44 +296,62 @@ pub async fn fetch_releases(
 ///
 /// Update Images with release metadata should be immutable, but
 /// tags on registry can be mutated at any time. Thus, the cache
-/// is keyed on the hash of tag layers.
-async fn cache_release(
+/// is keyed on the manifest reference.
+#[allow(clippy::too_many_arguments)]
```
vrutkovs: Create a struct 3 new params?

graph-builder/src/config/settings.rs
```diff
@@ -62,8 +62,19 @@ pub struct AppSettings {
     pub disable_quay_api_metadata: bool,
 
     /// Concurrency for graph fetching
-    #[default(16)]
+    #[default(crate::plugins::release_scrape_dockerv2::DEFAULT_FETCH_CONCURRENCY)]
     pub fetch_concurrency: usize,
+
+    /// Common prefix for graph-builder metrics.
+    #[default(Some("cincinnati_gb".to_string()))]
```
vrutkovs: Why does that need to be configurable?

vrutkovs (commented):
Looks fine to me, few questions / nitpicks

graph-builder/src/config/settings.rs
steveej: It doesn't necessarily need to be. Do you feel strongly about it not to be? If so, also about the `required_metrics`?

Just as a note, there's currently no way to actually configure this, because it hasn't been added to the [cli options](https://github.com/openshift/cincinnati/blob/ab9fed6c082131f8ffb67f38a82225751c672b6b/graph-builder/src/config/cli.rs#L10) or [file options](https://github.com/openshift/cincinnati/blob/ab9fed6c082131f8ffb67f38a82225751c672b6b/graph-builder/src/config/file.rs#L13).

graph-builder/src/config/settings.rs
steveej: In general I'd say no, the parser should handle non-TLS by now, see the test here: https://github.com/openshift/cincinnati/blob/f5a676548b8fb5d8c5218e7517ed97bf1b2408d0/graph-builder/src/registry.rs#L501-L507.

graph-builder/src/plugins/release_scrape_dockerv2/registry.rs
steveej: Yes. It would require to be cloned for every plugin invocation, which might be even better than using interior mutability as we do now. Any strong preferences?

graph-builder/src/plugins/release_scrape_dockerv2/registry.rs
steveej: I don't think this is a helpful lint. Creating a struct for free-standing functions parameter seems like an anti-pattern to me, but I can be convinced otherwise.

graph-builder/src/plugins/release_scrape_dockerv2/registry.rs
vrutkovs: I don't think there is any benefit in having (sync) `Cache` and `CacheAsync`, so there should be only one public - async one. Plugins should probably use their own cache copies

graph-builder/src/plugins/release_scrape_dockerv2/registry.rs
vrutkovs: Agree, lets keep it and clean it up later (along with `#[allow(clippy::useless_let_if_seq)]`)

graph-builder/src/config/settings.rs
vrutkovs: Right, this just adds a bit confusion as it can't in fact be changed. Also I can't imagine a situation where metrics would collide and a name change would be required.

`required_metrics` is useful - lets fail early if the app can't be monitored later

steveej: Alright, I'll add a constant to the settings then instead.

steveej: Done

graph-builder/src/plugins/release_scrape_dockerv2/registry.rs
steveej: This is reworked now. There's now only a single public cache type which can be instantiated with `cache::new`. I think this design is fine for the use-case of passing in a cache from the outside as well, which needs to be possible for implementing cache persistence in a separate module.

graph-builder/src/config/mod.rs
```diff
@@ -12,3 +12,6 @@ mod options;
 mod settings;
 
 pub use self::settings::AppSettings;
+
+/// Common prefix for graph-builder metrics.
+pub const METRICS_PREFIX: Option<&str> = Some("cincinnati_gb");
```
vrutkovs: Not sure if we want Option and Some here, but its okay to stay

steveej: Agreed. Let me see how much effort it is to change that.

graph-builder/src/plugins/release_scrape_dockerv2/mod.rs
```diff
@@ -0,0 +1,146 @@
...
+
+/// Metadata fetcher for quay.io API.
+#[derive(CustomDebug)]
+pub struct ReleaseScrapeDockerv2Plugin {
+    settings: ReleaseScrapeDockerv2Settings,
+    registry: registry::Registry,
+    cache: registry::cache::Cache,
+
+    #[debug(skip)]
+    graph_upstream_raw_releases: IntGauge,
```
steveej: leave `prometheus::` in place here

graph-builder/src/plugins/release_scrape_dockerv2/mod.rs
```diff
@@ -0,0 +1,146 @@
...
+}
+
+impl ReleaseScrapeDockerv2Plugin {
+    pub fn try_new(
+        mut settings: ReleaseScrapeDockerv2Settings,
+        cache: Option<registry::cache::Cache>,
+        prometheus_registry: Option<&prometheus::Registry>,
+    ) -> failure::Fallible<Self> {
+        let graph_upstream_raw_releases: IntGauge = IntGauge::new(
```
steveej: leave `prometheus::` in place here

steveej (commented):
Pair review session with @LalatenduMohanty.
In addition to the comments, I'll think about migrating the graph tests to the new plugin.

graph-builder/src/plugins/release_scrape_dockerv2/registry.rs
steveej: @vrutkovs resolved?

graph-builder/src/plugins/release_scrape_dockerv2/registry.rs
steveej: @vrutkovs resolved?

graph-builder/src/config/settings.rs
steveej: @vrutkovs resolved?

graph-builder/src/config/mod.rs
steveej: It wasn't much effort at all. PTAL

graph-builder/src/graph.rs
```diff
@@ -208,61 +190,23 @@ pub async fn run(settings: &config::AppSettings, state: &State) -> ! {
...
                 parameters: Default::default(),
             }),
         )
-        .await
-        {
-            Ok(internal_io) => internal_io.graph,
+        .await;
+        UPSTREAM_SCRAPES.inc();
+
+        let (graph, _) = match scrape {
```
vrutkovs: `internal_io.parameters` is being returned here and then thrown away?

vrutkovs (commented):
found another questionable part, other than that lgtm

graph-builder/src/graph.rs
steveej: Yes, awesome right? :-)

steveej:
@vrutkovs I addressed all comments. PTAL and resolve them respectively. Thanks for the many iterations :smile:

graph-builder/src/graph.rs
steveej: I have changed this to pass on the internal_io instance, so the outer code can finally decide which fields it needs from it.

vrutkovs (approved):
/hold

Looks good, lets see if e2e passes

PR #226, opened 2020-02-18
graph-builder: add plugin to fetch openshift/cincinnati-graph-data from github

steveej:
Add a github_openshift_secondary_metadata_scraper plugin:
* Implement github v3 API helper module with deserialization tests
* Implement plugin logic
* Test to fetch production data from
  https://github.com/openshift/cincinnati-graph-data

This plugin is in preparation for a subsequent plugin which will be able
to parse the fetched data.

---

* [x] Plugin to fetch and unpack repository
    * [x] Impl
    * [x] Network dependent tests
    * [ ] ~~Independent tests? e.g. mock Github API and have a custom generated tarball?~~
* [x] Add the legacy metadata to a raw metadata JSON in the data repository (https://github.com/openshift/cincinnati-graph-data/pull/74)
* [x] Double-check if any blocking code should be moved to the blocking thread-pool
* [ ] ~~Change graph-builder default configuration to use these plugins instead of the quay secondary metadata fetcher~~ moved to #232.

### DIFF
--- a/graph-builder/src/plugins/github_openshift_secondary_metadata_scraper/github_v3.rs
+++ b/graph-builder/src/plugins/github_openshift_secondary_metadata_scraper/github_v3.rs
@@ -1,0 +1,76 @@
+//! This is a helper module for accessing the [GitHub API v3][].
+//!
+//! [GitHub API v3]: https://developer.github.com/v3/
+
+/// Commit structure.
+#[derive(Default, Clone, Debug, Deserialize, PartialEq, Eq)]
+pub(crate) struct Commit {
+    pub(crate) sha: String,
+    pub(crate) url: String,
+}
+
+/// Branch structure.
+#[derive(Debug, Deserialize, PartialEq, Eq)]
+pub(crate) struct Branch {
+    pub(crate) name: String,
+    pub(crate) commit: Commit,
+    pub(crate) protected: bool,
+}
+
+/// Format the URL to request branch information.
+pub(crate) fn branches_url(org: &str, repo: &str) -> String {
+    format!(
+        "https://api.github.com/repos/{org}/{repo}/branches",
+        org = &org,
+        repo = &repo,
+    )
+}
+
+/// Format the URL to request a tarball URL.
+pub(crate) fn tarball_url(org: &str, repo: &str, commit: &Commit) -> String {
+    format!(
+        "https://api.github.com/repos/{org}/{repo}/tarball/{sha}",
+        org = org,
+        repo = repo,
+        sha = commit.sha,
+    )
+}
+
+/// Format a subdirectory name for a specific revision's tarball.
+pub(crate) fn archive_entry_directory_name(org: &str, repo: &str, commit: &Commit) -> String {
+    format!("{}-{}-{}", &org, &repo, &commit.sha[0..7],)
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn de_serialize_branch() {
+        let json = r#"
+            [
+                {
+                    "name": "master",
+                    "commit": {
+                        "sha": "fef06adb57b9d965bfc9ae0959bd038f3044207e",
+                        "url": "https://api.github.com/repos/openshift/cincinnati-graph-data/commits/fef06adb57b9d965bfc9ae0959bd038f3044207e"
+                    },
+                    "protected": true
+                }
+            ]
+            "#;
+
+        let branches = serde_json::from_str::<Vec<Branch>>(&json).unwrap();
+
+        let branches_expected = vec![Branch {
+                name: "master".to_string(),
+                commit: Commit {
+                        sha: "fef06adb57b9d965bfc9ae0959bd038f3044207e".to_string(),
+                        url: "https://api.github.com/repos/openshift/cincinnati-graph-data/commits/fef06adb57b9d965bfc9ae0959bd038f3044207e".to_string()
+                },
+                protected: true
+            }];
+
+        assert_eq!(branches_expected, branches);
+    }
+}
--- a/graph-builder/src/plugins/github_openshift_secondary_metadata_scraper/plugin.rs
+++ b/graph-builder/src/plugins/github_openshift_secondary_metadata_scraper/plugin.rs
@@ -1,0 +1,358 @@
+use super::github_v3;
+use async_trait::async_trait;
+use cincinnati::plugins::prelude::*;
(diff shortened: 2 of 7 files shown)


steveej:
the `OWNERS` files of the test fixtures were tripping up prow.

/verify-owners

graph-builder/src/plugins/openshift_secondary_metadata_parser/plugin.rs
```diff
@@ -0,0 +1,529 @@
...
+            ],
+        ),
+        (
+            ("4.2.13-s390x+s390x"),
+            vec![
+                (
+                    "io.openshift.upgrades.graph.previous.add",
+                    "4.2.10-s390x,4.2.12-s390x",
+                ),
+                ("io.openshift.upgrades.graph.previous.remove", "*"),
```
wking: This can go into cincinnati-graph-data

wking: Wait, this [is][1] in cincinnati-graph-data.  I don't think you need to manually inject it; might be a bug somewhere in the Rust parser?

[1]: https://github.com/openshift/cincinnati-graph-data/blob/b70334977789366606fcec18ab4207870c6df279/blocked-edges/4.2.13-s390x.yaml#L3

graph-builder/src/plugins/openshift_secondary_metadata_parser/plugin.rs
```diff
@@ -0,0 +1,529 @@
...
+                (
+                    "io.openshift.upgrades.graph.release.channels",
+                    "candidate-4.2,fast-4.2,stable-4.2",
```
wking: This is [in cincinnati-graph-data][1] too.  Might be another parser bug?  The graph-data entries use the raw `4.2.13-s390x` name, which means they apply to all releases whose internal name is `4.2.13-s390x` regardless of image architecture.  However, there is only a single release image named `4.2.13-s390x` and it's an s390x image.  So cincinnati-graph-data could use either `4.2.13-s390x` or `4.2.13-s390x+s390x` to talk about that release, and the plugin should generate the same graph in both cases.

[1]: https://github.com/openshift/cincinnati-graph-data/blob/b70334977789366606fcec18ab4207870c6df279/channels/stable-4.2.yaml#L31

steveej: The bug in this case was actually me putting this here. When I added the remaining parsing logic, this one stood out as duplicate channels in the assertion.

graph-builder/src/plugins/openshift_secondary_metadata_parser/plugin.rs
steveej: The issue is that the hack expanded the regex on Quay, in addition to this glob metadata. which means there are multiple labels for this metadata. This can be worked around by having the EdgeRemoveAdd plugin remove all it's processed metadata. This is required because the  parser plugin adds one `remove_regex` metadata entry, and while this will effectively be the same as long, a trivial metadata comparison will fail.

graph-builder/src/plugins/github_openshift_secondary_metadata_scraper/plugin.rs
```diff
@@ -0,0 +1,359 @@
...
+use cincinnati::plugins::prelude::*;
+use cincinnati::plugins::{InternalIO, InternalPlugin};
+use failure::{Fallible, ResultExt};
+use serde::Deserialize;
+use smart_default::SmartDefault;
+use std::path::PathBuf;
+use tokio::sync::Mutex as FuturesMutex;
+
+pub static DEFAULT_OUTPUT_WHITELIST: &[&str] = &[
+    "channels/.+\\.ya*ml",
```
vrutkovs: `ya+ml` (or maybe lets just make `yaml` as a hard requirement)

graph-builder/src/plugins/github_openshift_secondary_metadata_scraper/plugin.rs
```diff
@@ -0,0 +1,359 @@
...
+            .iter()
+            .try_fold(
+                Vec::with_capacity(settings.output_whitelist.len()),
+                |mut acc, cur| -> Fallible<_> {
+                    let re = regex::Regex::new(cur)?;
+                    acc.push(re);
+                    Ok(acc)
+                },
+            )
+            .context("Parsing output whhitelist strings as regex")?;
```
vrutkovs: typo: `whitelist`

graph-builder/src/plugins/github_openshift_secondary_metadata_scraper/github_v3.rs
```diff
@@ -0,0 +1,76 @@
...
+    format!(
+        "https://api.github.com/repos/{org}/{repo}/tarball/{sha}",
+        org = org,
+        repo = repo,
+        sha = commit.sha,
+    )
+}
+
+/// Format a subdirectory name for a specific revision's tarball.
+pub(crate) fn archive_entry_directsory_name(org: &str, repo: &str, commit: &Commit) -> String {
```
vrutkovs: typo: `archive_entry_directory_name`

vrutkovs (changes requested):
LGTM other than a few typos

vrutkovs:
I'm not too happy about binding Cincinnati to Github. Could it be reworked into a generic git repo (use clone instead of fetching data as a tarball)? This would make disconnected installs easier

steveej:
> I'm not too happy about binding Cincinnati to Github. Could it be reworked into a generic git repo (use clone instead of fetching data as a tarball)? This would make disconnected installs easier

I understand your concern and I can assure you that making GitHub a hard requirement is not a goal. 
The next step from here is to finish the graph-builder plugin configuration support.
That will allow us to configure the graph-builder so that the github scraper is disabled, and only the data parser is enabled. The data repository can be transferred and provided to the graph-builder by any means. 
This doesn't exclude building additional plugins which download the archive from a fixed URL, shell out to git, use a git library, etc.

vrutkovs:
>The data repository can be transferred and provided to the graph-builder by any means.

Ah, that makes sense (since data would be basically checked out files if I understood correctly)

steveej:
> Ah, that makes sense (since data would be basically checked out files if I understood correctly)

Yes. The plugin in this PR downloads and extracts them, while the plugin in #231 parses the extracted files. I will pick up #74 soon which will implement the missing piece.

vrutkovs (approved):
/lgtm

PR #227, opened 2020-02-19
WIP prow_rustfmt: run cargo build

vrutkovs:
This ensures the project builds in all 3 Rust versions, instead of
just the latest


steveej:
I don't agree that the project needs to build on all versions used for the format checks, as we may want to use features of the latest stable.

vrutkovs:
For some reason this doesn't exhibit the issue I had on rust 1.40, closing

PR #228, opened 2020-02-19
cargo_test: run `cargo kcov` to collect coverage in unittests

vrutkovs:
TODO:

* [x] Improve `cargo-test` to take less time
* [x] Save coverage report to test artifacts

kcov report would be stored in artifacts - `cargo-test/kcov-merged/index.html`.

[Current results](https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_cincinnati/228/pull-ci-openshift-cincinnati-master-cargo-test/927/artifacts/cargo-test/cov/kcov-merged/index.html)

### DIFF
--- a/dist/cargo_test.sh
+++ b/dist/cargo_test.sh
@@ -14,6 +14,7 @@ if [[ -n "${CINCINNATI_TEST_CREDENTIALS_PATH}" && -n "${CINCINNATI_TEST_QUAY_API
     cargo_test_flags["cincinnati"]+=",test-net-private"
     cargo_test_flags["graph-builder"]+=",test-net-private"
     cargo_test_flags["quay"]+=",test-net-private"
+
     export CINCINNATI_TEST_QUAY_API_TOKEN="$(cat ${CINCINNATI_TEST_QUAY_API_TOKEN_PATH})"
 fi
 
@@ -22,16 +23,50 @@ executors["cargo"]="execute_native"
 executors["docker"]="execute_docker"
 
 function run_tests() {
+  set -x
+  export CARGO_TARGET_DIR="$PWD/target"
+
+  if [[ $(type -f kcov) ]]; then
+    export HAS_KOV="${HAS_KOV:-1}"
+    rm -rf "${CARGO_TARGET_DIR}"/cov
+  fi
+
   for directory in ${!cargo_test_flags[*]}; do
+    if [[ ${HAS_KOV} -eq 1 ]]; then
+      # we want to prevent completely untested functions to be stripped
+      export RUSTFLAGS='-C link-dead-code'
+      export CARGO_ARGS="test --no-run"
+
+      # delete leftover test executibles
+      [[ ! -d "${CARGO_TARGET_DIR}"/debug ]] || find "${CARGO_TARGET_DIR}"/debug/ -maxdepth 1 -type f -executable -print -delete
+    fi
+
     (
-      set -x;
       ${1} /usr/bin/env bash -c "\
-        cd ${directory} && \
-        export CARGO_TARGET_DIR=\"../target\" && \
-        cargo ${CARGO_ARGS:-test --release} ${cargo_test_flags[${directory}]} && \
-      :"
+        cd ${directory}
+        cargo ${CARGO_ARGS:-test} ${cargo_test_flags[${directory}]}
+
+        if [[ ${HAS_KOV} -eq 1 ]]; then
+          rm -f \"${CARGO_TARGET_DIR}\"/debug/${directory}
+          find \"${CARGO_TARGET_DIR}\"/debug/ -maxdepth 1 -type f -executable -print -exec \
+          kcov \
+              --exclude-pattern=$HOME/.cargo \
+              --verify \
+              ${CARGO_TARGET_DIR}/cov \
+              {} \\;
+        fi
+      "
     )
+
   done
+
+  if [[ ${HAS_KOV} -eq 1 && -n "${ARTIFACTS_DIR}" ]]; then
+    mkdir -p "${ARTIFACTS_DIR}"
+    [[ ! -e "${ARTIFACTS_DIR}"/cov ]] || rm -rf "${ARTIFACTS_DIR}"/cov
+    cp -rf "${CARGO_TARGET_DIR}"/cov "${ARTIFACTS_DIR}"/
+  fi
+
+  set +x
 }
 
 function execute_native() {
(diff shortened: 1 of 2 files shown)


vrutkovs:
Requires https://github.com/openshift/release/pull/7273 to display artifacts

dist/cargo_test.sh
```diff
@@ -23,16 +23,29 @@ executors["cargo"]="execute_native"
...
+        export RUSTFLAGS='-C link-dead-code'
+        export CARGO_ARGS="kcov --verbose --all --no-clean-rebuild"
+      fi
+
       ${1} /usr/bin/env bash -c "\
         cd ${directory} && \
         export CARGO_TARGET_DIR=\"../target\" && \
-        cargo ${CARGO_ARGS:-test --release} ${cargo_test_flags[${directory}]} && \
+        cargo test --no-run &&
+        cargo ${CARGO_ARGS:-test} ${cargo_test_flags[${directory}]} && \
```
steveej: I'm curious why you removed `--release` here. Would you mind to explain?

vrutkovs: This follows "hack/coverage.sh" script - don't release optimizations remove debug info so kcov won't be able to collect info? I'll check that in the debug commit

dist/cargo_test.sh
```diff
@@ -23,16 +23,29 @@ executors["cargo"]="execute_native"
...
+      if [[ "${HAS_KOV}" ]]; then
+        # we want to prevent completely untested function to be stripped
+        export RUSTFLAGS='-C link-dead-code'
+        export CARGO_ARGS="kcov --verbose --all --no-clean-rebuild"
+      fi
+
       ${1} /usr/bin/env bash -c "\
         cd ${directory} && \
         export CARGO_TARGET_DIR=\"../target\" && \
+        cargo test --release --no-run &&
```
steveej: This compilation of the tests in `--release` mode is superfluous. The next line will compile and run the tests accordingly.

```suggestion
```

vrutkovs: Turns out `cargo kcov` needs it - it won't build tests itself

steveej (commented):
This doesn't (yet) deal with collecting the result as artifacts, is this correct?
It would be nice to get a coverage report by a bot for each PR, e.g. through https://coveralls.io/.

vrutkovs:
We do collect coverage results: open cargo-test details, Artifacts link -> artifacts / cargo-test / cov / index.html

Uploading to coveralls.io is out of scope - it would require at least a token and probably some tweaking to emulate Travis

LalatenduMohanty (changes requested):
Does it require `kcov` to be installed explicitly? If yes, we need to update https://github.com/openshift/cincinnati/blob/master/docs/developer/developing.md#prerequisites 
```
$ just run-ci-tests

xxxx
error: no such subcommand: `kcov`

        Did you mean `doc`?
error: Recipe `run-ci-tests` failed with exit code 101
```

vrutkovs:
Unit tests would run coverage collection only if `kcov` binary is installed.

I don't use `just` (its not a prerequisite really), so I'll leave updating this section to Lala

LalatenduMohanty:
> Unit tests would run coverage collection only if `kcov` binary is installed.

We need to keep track of prerequisites somewhere and at this point it is https://github.com/openshift/cincinnati/blob/master/docs/developer/developing.md#prerequisites 

If we do not like `Just` then that's a different issue and may be we can separate the perquisites in to two parts.

LalatenduMohanty:
With the recent changes to the PR, `kcov` is not a prerequisite anymore. So we do not need to update the doc.

steveej:
> We do collect coverage results: open cargo-test details, Artifacts link -> artifacts / cargo-test / cov / index.html
> 
> Uploading to coveralls.io is out of scope - it would require at least a token and probably some tweaking to emulate Travis

I wasn't aware that this is collected into the artifacts already, thanks for the clarification!
/lgtm

Letting you sort out the dependency issue with @LalatenduMohanty.
/hold

steveej:
This fails with

```
plugins::internal::graph_builder::openshift_secondary_metadata_parser::plugin::tests::compare_quay_result_fixture_20200220104838 stdout ----
Error: Os { code: 2, kind: NotFound, message: "No such file or directory" }

Reading "src/plugins/internal/graph_builder/openshift_secondary_metadata_parser/test_fixtures/20200220.104838/graph-gb-raw.json" to string
```

Hitting the unresolved https://github.com/kennytm/cargo-kcov/issues/17. There is currently no proper solution for this. As a workaround we can move the fixtures to the code as inline strings and write them to tempfiles for the test. If we agree on this workaround I can prepare a PR for it.

vrutkovs:
>If we agree on this workaround I can prepare a PR for it.

Yeah, I don't think there is a better way to have this fixed.

/hold

steveej:
> > If we agree on this workaround I can prepare a PR for it.
> 
> Yeah, I don't think there is a better way to have this fixed.

I think that changing _cargo-kcov_ would be a proper solution, but I understand that it's not a quick fix. OTOH, there are a lot of fixtures in the repo now, and turning them into inline strings is not a very quick fix either. I'll investigate this a bit more.

/assign

steveej:
I pushed two commits on top of your branch. In addition to switching to using `kcov` directly, I also made the artifacts directory optional and configurable. This means we need to set `ARTIFACTS_DIR` no in the release job config.

dist/cargo_test.sh
```diff
@@ -22,16 +23,52 @@ executors["cargo"]="execute_native"
...
+              --verify \
+              ${CARGO_TARGET_DIR}/cov \
+              {} \\;
+        fi
+      "
     )
+
   done
+
+  # TODO: remove this before merging the PR
```
vrutkovs: I don't see anything wrong about this for now really

steveej: I'd prefer not to have this in by default, rather only if it's passed in by the caller. I added it to give the CI job a chance to succeed. Now that we know that it works we can properly integrate it.

vrutkovs:
/hold cancel

Ready for review. ARTIFACTS_DIR has to land in this PR so that it would pass tests

steveej:
> ARTIFACTS_DIR has to land in this PR so that it would pass tests

This is not true at the moment, because we're setting a default [here](https://github.com/openshift/cincinnati/pull/228/files#diff-fcb40cdd14e5204dff404b456b6ea179R64). I would like it to be true though, i.e. I'd prefer the removal of that line according to its TODO line above it.

vrutkovs:
https://github.com/openshift/release/pull/7528 would ensure its being passed here and defaults can be removed

steveej:
> [openshift/release#7528](https://github.com/openshift/release/pull/7528) would ensure its being passed here and defaults can be removed.

Nice! Please let's get that one merged and address the TODO in the code here.

steveej (approved):
/lgtm

PR #229, opened 2020-02-20
graph-builder/tests/e2e: make sorting algorithm (hopefully) fully predictable

steveej:
This adds a fix to also sort the
edges array in itself. This is required to make the output fully
predictable in case a node has multiple edges, and thus makes the output assertable for equality.

The function has also been moved to a common place for reusage in other tests.

/cc @vrutkovs


steveej:
/retest
after a flaky e2e test, which was likely fixed by #230.

vrutkovs (approved):
/lgtm

PR #231, opened 2020-02-24
graph-builder: add plugin to parse openshift/cincinnati-graph-data

steveej:
Add openshift_secondary_metadata_parser plugin:
* Implement plugin logic
* Test using fixtures produced with a custom graph-builder to obtain a
  snapshot of the following triple:
    * a vanilla graph produced by the registry scraper plugin
    * a vanilla graph augmented by the quay secondary metadata plugin
    * the data from the openshift/cincinnati-graph-data repository
  The test asserts equality of two results of the EdgeAddRemovePlugin.
  The first result is obtianed by passing the raw graph through the new
  parser plugin and then through the EdgeAddRemovePlugin.
  The second passing the quay metadata graph fixture through the new
  parser plugin.
  This ensures compatibility between the new plugin and the
  QuayMetadataPlugin with respect to equality for subsequent processing
  of their produced metadata.

  Note: for this to work I had to add raw metadata to the fixtures,
  which represents metadata present on quay, but not present as part of
  the existing cincinnati-graph-data. This raw metadata has also been
  added to the upstream data repository.

---

Split off the original #226.

* [x] Plugin to parse the contents of the repository and mangle the Graph
    * [x] Impl
    * [ ] ~~Tests which rely on the previous plugin?~~
    * [x] Independent tests, i.e. generate a known file structure on-the-fly in the test
* [x] Double-check to either use `tokio::fs` functions or use `spawn_blocking`


cincinnati/src/plugins/internal/edge_add_remove.rs
```diff
@@ -19,6 +19,10 @@ pub struct EdgeAddRemovePlugin {
 
     #[default(DEFAULT_REMOVE_ALL_EDGES_VALUE.to_string())]
     pub remove_all_edges_value: String,
+
+    /// If true causes the removal of all processed metadata from the releases.
+    #[default(false)]
+    pub remove_consumed_metadata: bool,
```
wking, line 25: Do we need to make this tuneable?  Can we just always do it?

steveej: It can be helpful for debugging to leave them in, so see what annotations are on the release. Also for now it's still off by default, which is compatible with the existing deployments.

graph-builder/src/plugins/openshift_secondary_metadata_parser/plugin.rs
```diff
@@ -0,0 +1,513 @@
...
+                Ok(())
+            })?;
+
+        Ok(())
+    }
+
+    async fn process_channels(&self, io: &mut InternalIO) -> Fallible<()> {
+        let channels_dir = self.settings.data_directory.join("channels");
+        let channels: Vec<graph_data_model::Channel> =
+            deserialize_directory_files(&channels_dir, regex::Regex::new("ya+ml")?)
```
vrutkovs, line 309: Perhaps its worth sticking to `yaml` only until file extensions go out of control :)

cincinnati/src/plugins/internal/edge_add_remove.rs
```diff
@@ -53,7 +57,8 @@ impl PluginSettings for EdgeAddRemovePlugin {
 ///     2. next
 /// 2. *.remove
 ///     1. previous
```
vrutkovs, line 59: Seems both `previous` and `previous_regex` data would be applied. Shouldn't we prefer `previous_regex` instead?

cincinnati/src/plugins/internal/edge_add_remove.rs
```diff
@@ -127,11 +139,66 @@ impl EdgeAddRemovePlugin {
...
+                                release.version(),
+                            );
+                            true
+                        } else {
+                            false
+                        }
+                    });
+
+                    for (from, from_version) in froms {
+                        info!("[{}]: removing previous {}", to_version, from_version);
```
vrutkovs: "removing previous {} by regexp" so that it would be easier to distinguish code paths

cincinnati/src/plugins/internal/edge_add_remove.rs
```diff
@@ -127,11 +139,66 @@ impl EdgeAddRemovePlugin {
...
+                            .map(|metadata| metadata.remove(&previous_remove_regex_key))?;
+                    }
+
+                    if from_regex_string == ".*" {
+                        let parents: Vec<daggy::EdgeIndex> = graph
+                            .previous_releases(&to)
+                            .map(|(edge_index, _, _)| edge_index)
+                            .collect();
+
+                        trace!("removing parents for '{}': {:?}", to_version, parents);
```
vrutkovs: "removing parents for '{}': {:?} by regexp" or smth like that to distinguish code path

cincinnati/src/plugins/internal/edge_add_remove.rs
```diff
@@ -127,11 +139,66 @@ impl EdgeAddRemovePlugin {
...
+                        return graph.remove_edges_by_index(&parents);
+                    };
+
+                    let from_regex = regex::Regex::new(&from_regex_string)
```
vrutkovs: This should happen before other `trace!` calls so that remove_regexp code path would be more distinct

graph-builder/src/plugins/openshift_secondary_metadata_parser/plugin.rs
```diff
@@ -0,0 +1,513 @@
...
+
+                // add metadata to block edge using the `previous.remove_regex` metadata
+                io.graph
+                    .get_metadata_as_ref_mut(&release_id)
+                    .context(format!(
+                        "[blocked_edges] Getting mutable metadata for {}",
+                        &to.to_string()
+                    ))?
+                    .insert(
+                        format!("{}.{}", self.settings.key_prefix, "previous.remove_regex"),
```
vrutkovs, line 296: I'm not entirely sure this would be a regex - is it an exact version?

vrutkovs (commented):
Looks good, a few nitpicks and odd parts

dist/prow_yaml_lint.sh
```diff
@@ -11,4 +11,6 @@ if ! type -f "${YAML_LINTER}"; then
   exit 1
 fi
 
-find . -type f \( -name '*.yaml' -o -name '*.yml' \) -print0 | xargs -L 1 -0 "${YAML_LINT_CMD[@]}"
+find . \
+  -path "./graph-builder/src/plugins/openshift_secondary_metadata_parser/test_fixtures" -prune -o \
```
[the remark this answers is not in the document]
steveej: there's only one usage of it though, and I could see us wanting to prune more than just this one. usually I'm all for constants if they benefit the DRY principle, but in this case it would just add indirection.

dist/prow_yaml_lint.sh
```diff
@@ -11,4 +11,6 @@ if ! type -f "${YAML_LINTER}"; then
...
+  -type f \( -name '*.yaml' -o -name '*.yml' \) -print0 | xargs -L 1 -0 "${YAML_LINT_CMD[@]}"
```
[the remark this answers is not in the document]
steveej: I didn't intend to refactor the whole script ;-)

graph-builder/src/plugins/openshift_secondary_metadata_parser/plugin.rs
steveej: This is currently in line with #226. I think both (yml and yaml) are common extensions. Do you think we should make it configurable?

dist/prow_yaml_lint.sh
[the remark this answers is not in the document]
vrutkovs: Okay, it can be reworked later

graph-builder/src/plugins/openshift_secondary_metadata_parser/plugin.rs
vrutkovs: They are common, but I'd rather not leave a space for errors here.

@wking @LalatenduMohanty do you agree we should stick to "<file>.yaml" and ignore (possible) "<file>.yml" in graph-data repo?

cincinnati/src/plugins/internal/edge_add_remove.rs
steveej: I think the existing syntax can coexist with the additional one, so I don't see why we should remove the existing one. What makes you think we should?

vrutkovs: I'd expect we use either `previous` or `previous_regex` in `blocked-edges`, but current implementation allows both. This may lead to a few odd situations - lets throw error when both are specified?

graph-builder/src/plugins/openshift_secondary_metadata_parser/plugin.rs
steveej: If by *this* you mean `blocked_edge.from`, it is defined to be here: https://github.com/openshift/cincinnati/blob/89cd657bd552efdd1f8bd2c3c57a038e6cfecf9d/graph-builder/src/plugins/openshift_secondary_metadata_parser/plugin.rs#L20-L22

steveej:
> Looks good, a few nitpicks and odd parts

Thanks for the review! I addressed your comments either by code changes or answers.

cincinnati/src/plugins/internal/edge_add_remove.rs
steveej: I don't see why it would be an error. They could be used in conjunction as well. Why do you want to restrict it?

vrutkovs: Its not an error really, just feels that it might have undesirable consequences. Maybe I'm overreacting really as I can't come up with a potential issue on this one

steveej:
Rebased and last TODO item completed!

cincinnati/src/plugins/internal/edge_add_remove.rs
steveej: If it helps, I don't see a potential issue in this part of the code either, and I've thought about it quite a bit :slightly_smiling_face:

graph-builder/src/plugins/openshift_secondary_metadata_parser/plugin.rs
vrutkovs: Oh, correct, thanks!

cincinnati/src/plugins/internal/edge_add_remove.rs
LalatenduMohanty: Can this be used to force the plugin to discard previous processed metadata and start with fresh metadata scanning ?

graph-builder/src/plugins/openshift_secondary_metadata_parser/plugin.rs
LalatenduMohanty: Standard practice for yaml file uses  `.yaml , .yml` . So we should only support these two extensions. Just supporting `yaml` is fine too but then someone might request to add `yml` as well.  So supporting `.yaml , .yml` makes us future proof.

LalatenduMohanty:
@steveeJ Can we merge some of the commits with single line code changes still keep them logically separate? IMO all code changes related to the plugin should be just in one commit.

steveej:
> IMO all code changes related to the plugin should be just in one commit.

I suggest not squashing them, as it makes reverting commits more difficult in the future. Take ab718b4ab985601172e4345060990380693b2ba0 for example which exposes a constant. If this is squashed with the plugin commit, and we would want to revert the plugin commit, there is no guarantee the reversal won't break other code which in the meantime also depends on the public constant.

vrutkovs:
It seems squashing https://github.com/openshift/cincinnati/pull/231/commits/87a928f7c50e44cdfb4198342d2b30b7078ddb0b and https://github.com/openshift/cincinnati/pull/231/commits/ab718b4ab985601172e4345060990380693b2ba0 would be a good idea.

Other than that it looks good

steveej:
> It seems squashing [87a928f](https://github.com/openshift/cincinnati/commit/87a928f7c50e44cdfb4198342d2b30b7078ddb0b) and [ab718b4](https://github.com/openshift/cincinnati/commit/ab718b4ab985601172e4345060990380693b2ba0) would be a good idea.

Agreed, these two are not unreasonable to squash, done!

vrutkovs (approved):
/lgtm

cincinnati/src/plugins/internal/edge_add_remove.rs
steveej: This controls whether the metadata which is used by this plugin is removed after processing or not. As an example, if there is a `previous.remove` with the value `1.2.3` metadata on a release, this plugin will remove the edge, in the graph. If `remove_consumed_metadata` is `true` in the plugin configuration, it will also remove the metadata entry with the key `previous.remove`.

PR #232, opened 2020-02-25
graph-builder: use openshift/cincinnati-graph-data instead of quay labels

steveej:
This configures the graph-builder binary to use the
GithubOpenshiftSecondaryMetadataScraperPlugin and the
OpenshiftSecondaryMetadataParserPlugin in conjunction to use
https://github.com/openshift/cincinnati-graph-data as the new source for
secondary metadata.

---

- [x] Requires #226 
- [x] Requires #231 
- [x] Github token support in scraper https://github.com/openshift/cincinnati/pull/233
- [x] Configure the e2e test cluster to work with the new plugin (requires https://github.com/openshift/cincinnati/pull/234, and possibly more)
- [x] Based on #250 
- [x] Based on #258


vrutkovs:
We're now using prod metadata - but tests are still using channels "a" and "b"

steveej:
> We're now using prod metadata - but tests are still using channels "a" and "b"

Note that we're only using prod secondary metadata, while using test primary metadata :smile: 

When I started working on a solution for this locally, I stumbled upon a real issue:

```
[2020-02-27T10:06:44Z ERROR graph_builder::graph] Parsing {"message":"API rate limit exceeded for 82.197.161.223. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)","documentation_url":"https://developer.github.com/v3/#rate-limiting"} to Vec<Branch>
```

Our default delay of 30 seconds is too short for GitHub's API, the good news is we're not that far away:

> For unauthenticated requests, the rate limit allows for up to 60 requests per hour. Unauthenticated requests are associated with the originating IP address, and not the user making requests.

I.e., raising the delay to 60 seconds, assuming our deployments have different IPs, will do it until we have authentication against the GitHub API. If we have consensus I will raise the delay to 60 in this PR as well.

vrutkovs:
>If we have consensus I will raise the delay to 60 in this PR as well.

SGTM, perhaps we should invest into token support (most likely we'll use that on stage / prod anyway)

steveej:
> SGTM, perhaps we should invest into token support (most likely we'll use that on stage / prod anyway)

Definitely a mid-term goal, but I wouldn't block on it, as I think 60s, or even 120s is a small duration compared to the release frequency and the cluster sync frequency.

dist/openshift/cincinnati.yaml
```diff
@@ -308,3 +309,6 @@ parameters:
   - name: RUST_BACKTRACE
     value: "0"
     displayName: Set RUST_BACKTRACE env var
+  - name: GB_UPSTREAM_PAUSE_SECS
+    value: "120"
+    displayName: Pause between upstream scrapes
```
LalatenduMohanty: We should not use anonymous auth for github v3 API call. As you can easily hit the rate limits if you are behind NAT networking. Github considers all calls to github from NAT network as a single source. So we should implement a way to use OAuth2 tokens for github calls. Also without oauth2 this is not ready for production use and as we are following continuous delivery we should not differentiate between stage and production.

LalatenduMohanty (changes requested):

steveej:
As per [@LalatenduMohanty comment](https://github.com/openshift/cincinnati/pull/232#discussion_r385060528) I have changed my opinion on blocking on authentication. The comment reminded me of our intention to follow best continuous delivery  practices, which means that we shouldn't compromise our quality in master (which feeds stage) if not necessary.
In this case it can be prevented.

I implemented authentication in a new commit. Now we need to get secrets for CI, staging and production and wire them through in the deployment templates.

graph-builder/src/plugins/github_openshift_secondary_metadata_scraper/plugin.rs
```diff
@@ -73,10 +94,20 @@ impl GithubOpenshiftSecondaryMetadataScraperPlugin {
         let url = github_v3::branches_url(&self.settings.github_org, &self.settings.github_repo);
 
         trace!("Getting branches from {}", &url);
-        let bytes = reqwest::Client::new()
-            .get(&url)
-            .header(reqwest::header::USER_AGENT, USER_AGENT)
-            .header(reqwest::header::ACCEPT, "application/vnd.github.v3+json")
+
+        let request = {
+            let request = reqwest::Client::new()
```
vrutkovs: Create Client once in `try_new` and reuse it here to reuse the connection to Github

wking:
Needs a rebase?  This PR seems quiet; anything else outstanding around it?

steveej:
One of the e2e test cases currently fails due to a mismatch in edge handling between quay-metadata and github-metadata parsing:

```
stripped from expected: {
    "4.2.11-s390x": {
        "io.openshift.upgrades.graph.previous.remove": "\"*\"",
    },
    "4.2.22": {
        "io.openshift.upgrades.graph.previous.remove": "\"*\"",
    },
    "4.2.13-s390x": {
        "io.openshift.upgrades.graph.previous.remove": "\"*\"",
    },
}
stripped from actual: {
    "4.2.11-s390x": {
        "io.openshift.upgrades.graph.previous.remove_regex": "\".*\"",
    },
    "4.2.13-s390x": {
        "io.openshift.upgrades.graph.previous.remove_regex": "\".*\"",
    },
}
```
This needs to be investigated, hence putting this on hold.
/hold

cincinnati/src/plugins/internal/graph_builder/github_openshift_secondary_metadata_scraper/github_v3.rs
```diff
@@ -43,6 +43,14 @@ pub(crate) fn archive_entry_directory_name(org: &str, repo: &str, commit: &Commi
     format!("{}-{}-{}", &org, &repo, &commit.sha[0..7],)
 }
 
+/// Format a commit URL
+pub(crate) fn commit_url(org: &str, repo: &str, sha: &str) -> String {
+    format!(
+        "https://api.github.com/repos/{}/{}/commits/{}",
```
wking, line 49: This sort of thing will require restricted-network folks who want to use this plugin to use the GitHub API (vs. just hosting a Git repository over Git's native Git or HTTP(S) protocols).  I guess that's an existing requirement from #234, so we can always come back and revisit later, but just thought I'd mention it here.

steveej: This plugin is the GitHub specific scraper. The idea behind decoupling the scraper and the parser is that the scraper can be replaced by something else, or could in fact just be skipped, where the files for the parser would be managed out-of-process somehow. All the possibilities here ;-)

steveej:
This should be good to go now.

/hold cancel

hack/e2e.sh
```diff
@@ -54,29 +54,45 @@ oc secrets link default ci-pull-secret --for=pull
...
+
 # Apply oc template
 oc new-app -f dist/openshift/cincinnati.yaml \
   -p IMAGE="${IMAGE}" \
   -p IMAGE_TAG="${IMAGE_TAG}" \
   -p GB_CPU_REQUEST=50m \
   -p PE_CPU_REQUEST=50m \
   -p RUST_BACKTRACE="1" \
-  -p GB_PLUGIN_SETTINGS='
+  -p GB_PLUGIN_SETTINGS="$(envsubst <<'EOF'
```
wking: No need for `envsubst`:

```console
$ A=some-value
$ python -c 'import sys; print(sys.argv)' --example="$(cat <<EOF
> here it is, all expanded: ${A}
> EOF  
> )"
['-c', '--example=here it is, all expanded: some-value']
```

steveej: Can you still use `"` inside the given argument?

steveej: It seems so ;-) Thanks for the improvement suggestion!

cincinnati/src/lib.rs
```diff
@@ -452,7 +452,10 @@ impl Graph {
 
     /// Get the edges expressed as version -> versions
```
vrutkovs: Lets update this, `include_abstract=true` would filter out abstract (== legacy releases not in cincinnati-graph-data) releases?

cincinnati/src/plugins/internal/graph_builder/github_openshift_secondary_metadata_scraper/github_v3.rs
vrutkovs: We should investigate if its worth supporting Github Enterprise - or a plain git plugin is more common. I think its fine to go as is atm

graph-builder/Cargo.toml
```diff
@@ -41,6 +41,7 @@ custom_debug_derive = "^0.1.7"
 built = "^0.3.2"
 
 [dev-dependencies]
+cincinnati = { path = "../cincinnati", features = ["test"] }
```
vrutkovs: Not sure why are we enabling features here (also feature name is too generic imo)

hack/e2e.sh
```diff
@@ -54,29 +54,40 @@ oc secrets link default ci-pull-secret --for=pull
...
+
 # Apply oc template
 oc new-app -f dist/openshift/cincinnati.yaml \
   -p IMAGE="${IMAGE}" \
   -p IMAGE_TAG="${IMAGE_TAG}" \
   -p GB_CPU_REQUEST=50m \
   -p PE_CPU_REQUEST=50m \
   -p RUST_BACKTRACE="1" \
-  -p GB_PLUGIN_SETTINGS='
+  -p GB_PLUGIN_SETTINGS="$(cat <<-EOF
```
vrutkovs, line 70: bash :/

Lets use the following:

* create a plain param file:
```
IMAGE="${IMAGE}"
IMAGE_TAG="${IMAGE_TAG}"
...
```

* use `sed`/`envsubst`/moar bash magic to template it
* run `oc new-app -f dist/openshift/cincinnati.yaml -p param-file`

vrutkovs (changes requested):
looks good overall, `hack/e2e.sh` is growing way too complex though.

I'd prefer to merge this after #249 so that we could compare performance of quay and github scrapers

hack/e2e.sh
steveej: I'd rather not rework this now as part of this PR because it works as is. I do agree that we want to change this in the future, but refactoring it is out of this PRs scope.

graph-builder/Cargo.toml
steveej: Certain test-related functions in the cincinnati crate are now gated by the `test` feature. This is a workaround to the limitation that code gated with `#[cfg(test)]` is not exposed to the tests of dependent crates. That's why I used `#[cfg(any(test, feature = "test"))]` to gate these functions in the cincinnati crate.

vrutkovs: Oh, I see, that makes sense

cincinnati/src/lib.rs
steveej: Abstract releases are created by edges which refer to non-existing versions. I'd rather not discuss this here, but refer you to #188 and #256.

hack/e2e.sh
vrutkovs: okay, lets rework it later once we introduce different e2e tests

cincinnati/src/lib.rs
steveej: (Changing the docstring is fine of course)

cincinnati/src/plugins/internal/graph_builder/github_openshift_secondary_metadata_scraper/github_v3.rs
steveej: > or a plain git plugin is more common

I think a plain git plugin would be most useful in the long-term, as it's easy to transfer to and set up in a network-restricted environment while preserving the metadata ( at _.git_). It will be easy to swap out this plugin out simply by configuration in the future, so we can leave it in place now and revisit later.

wking:
[cargo-test][1]:

```
error: error creating buildah builder: Error parsing image configuration: Get https://storage.googleapis.com/openshift-gce-devel-origin-ci-registry/docker/registry/v2/blobs/sha256/f3/f3c24bf7fc6253c6974077ff63b59705cbb525f4f39c9ba512bbe398d24f1559/data?Expires=1585339175&GoogleAccessId=api-ci%40openshift-ci-infra.iam.gserviceaccount.com&Signature=Uj4LFTCUbu9QchNKuYZj10D0rnmIe%2FpXpOEhFt8AxfUCgCBoyae1fhLMcOEygZfXDdWl%2Funi1WQncac9lHpE6MAcDdEFVfdgdrM%2BxRcgJLEknB1cL3gxu0vgSa2Q4atJ1Qtd17V6RySf7MTot74dl9uOoZ7P6O9K%2Fna2rblTJLglWQ5skfT3SCpe6Gcxv1SWQBzu%2BisghffeOvlRYXr9o%2B9WkInLvir38ydf81OKwMkEObmcBtBARA23NH2HkJoCYi0R9HHxvBkhPLB3aEUcpF2gv5SPK8oBwz%2BhuaYBnIAJ%2FTFbhDxHbzOvAItqGFjrP9YlkppirXrHNrBzszxhwQ%3D%3D: read tcp 10.131.62.97:47196->172.217.15.112:443: read: connection reset by peer
```

Maybe Google is just sad.

/retest

[1]: https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/pr-logs/pull/openshift_cincinnati/232/pull-ci-openshift-cincinnati-master-cargo-test/1167

vrutkovs:
SLO tests failed
/retest

vrutkovs:
IIUC this still checks `a` and `b` channels in load-tests - I guess it should use `stable-4.2` and `stable-4.3` now? 

Huge performance dip:
```
Bucket           #      %       Histogram
[0s,     50ms]   85495  59.00%  ############################################
[50ms,   100ms]  1836   1.27%   
[100ms,  500ms]  3481   2.40%   #
[500ms,  1s]     4911   3.39%   ##
[1s,     5s]     23311  16.09%  ############
[5s,     10s]    14361  9.91%   #######
[10s,    +Inf]   11505  7.94%   #####
```

At 500 rps Cincinnati starts throwing errors:
```
Testing workers 10, rate 500 -> /tmp/tmp.B5jntvd3Cp/rate-500-workers-10.bin
Requests      [total, rate, throughput]         15000, 500.03, 161.26
Duration      [total, attack, wait]             48.444s, 29.998s, 18.446s
Latencies     [min, mean, 50, 90, 95, 99, max]  14.727ms, 3.664s, 1.856s, 10.4s, 13.308s, 24.589s, 30.004s
Bytes In      [total, mean]                     15664937, 1044.33
Bytes Out     [total, mean]                     0, 0.00
Success       [ratio]                           52.08%
Status Codes  [code:count]                      0:190  200:7812  500:116  502:2009  503:4873  
```

steveej:
It doesn't make sense to me that a change in the graph-builder would result in drastically decreased performance on the policy-engine endpoint. I think we're dealing with load fluctuation on the CI cluster here.

/retest

graph-builder/Cargo.toml
```diff
@@ -42,8 +42,6 @@ built = "^0.3.2"
 
 [dev-dependencies]
 twoway = "^0.2"
-assert-json-diff = "1.0.0"
-test-case = "1.0.0"
```
LalatenduMohanty, line 46: Does `cargo update` removes the stale entries? If not which command?

steveej: `cargo update` doesn't alter the `Cargo.toml` at all, but only `Cargo.lock`.

dist/openshift/cincinnati.yaml
```diff
@@ -318,11 +318,15 @@ parameters:
       credentials_path = "/etc/secrets/registry_credentials_docker.json"
 
       [[plugin_settings]]
-      name = "quay-metadata"
-      repository = "openshift-release-dev/ocp-release"
+      name = "github-secondary-metadata-scrape"
+      github_org = "openshift"
+      github_repo = "cincinnati-graph-data"
+      reference_branch = "master"
```
LalatenduMohanty, line 324: :+1:

graph-builder/Cargo.toml
LalatenduMohanty: What is the command for detecting stale entries in cargo toml file?

Justfile
```diff
@@ -82,41 +156,65 @@ run-graph-builder registry="https://quay.io" repository="openshift-release-dev/o
...
 
 		[[plugin_settings]]
-		name="quay-metadata"
-		repository="{{repository}}"
+		name = "github-secondary-metadata-scrape"
+		github_org = "openshift"
+		github_repo = "cincinnati-graph-data"
+		branch = "master"
+		output_directory = "${TMPDIR}"
+		{{metadata_reference}}
```
LalatenduMohanty, line 164: :+1:

LalatenduMohanty (changes requested):

graph-builder/src/config/settings.rs
```diff
@@ -146,12 +154,30 @@ impl AppSettings {
...
-                ("name", QuayMetadataFetchPlugin::PLUGIN_NAME),
-                ("repository", &self.repository),
-                ("manifestref_key", &self.manifestref_key)
+            GithubOpenshiftSecondaryMetadataScraperSettings::deserialize_config(toml::from_str(
+                &format!(
+                    r#"
+                        github_org = "openshift"
+                        github_repo = "cincinnati-graph-data"
+                        branch = "master"
+                        output_directory = {:?}
```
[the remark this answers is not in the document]
steveej: This function (`default_openshift_plugin_settings`) is not used at all if any plugin configuration is provided. Customization of the plugin config is part of the graph-builders TOML config.

graph-builder/Cargo.toml
steveej: I'm not aware of one. I simply knew that they weren't used anymore.

steveej:
Amending to [my previous comment](https://github.com/openshift/cincinnati/pull/232#issuecomment-605927616), we're not dealing with actual fluctuation of the CI, but I'm certain that this PR doesn't degrade the the performance of the Cincinnati stack either. This PR does change the e2e deployment to use production data, which is the likely cause of performance changes.

vrutkovs:
This LGTM, lets squash in a single commit

vrutkovs (approved):
/lgtm

:tada:

steveej:
> New changes are detected. LGTM label has been removed.

Just reworded a commit message, no content changes:

```shell
$ git diff 814bee1 a8cddc8 | wc -l
0
```

cincinnati/src/plugins/internal/graph_builder/github_openshift_secondary_metadata_scraper/github_v3.rs
LalatenduMohanty: We should revisit this from air-gap cluster point of view for sure.

PR #233, opened 2020-02-27
graph-builder/plugins/github scraper: add token authorization handling

steveej:
This implements reading a token from a file and setting the authorization header on the request towards the GitHub API.

It also adjusts the deployment template.

- [x] Github API token for CI https://github.com/openshift/release/pull/7356
- [x] Github API token for stage https://gitlab.cee.redhat.com/service/app-interface/merge_requests/3392
- [x] Based on #245

### DIFF
--- a/cincinnati/src/plugins/internal/graph_builder/github_openshift_secondary_metadata_scraper/plugin.rs
+++ b/cincinnati/src/plugins/internal/graph_builder/github_openshift_secondary_metadata_scraper/plugin.rs
@@ -13,6 +13,9 @@ pub static DEFAULT_OUTPUT_WHITELIST: &[&str] = &[
     "raw/metadata.json",
 ];
 
+/// Environment variable name for the Oauth token path
+pub static GITHUB_SCRAPER_TOKEN_PATH_ENV: &str = "CINCINNATI_GITHUB_SCRAPER_OAUTH_TOKEN_PATH";
+
 static USER_AGENT: &str = "openshift/cincinnati";
 
 /// Plugin settings.
@@ -28,6 +31,7 @@ pub struct GithubOpenshiftSecondaryMetadataScraperSettings {
     /// An empty vector is regarded as a configuration error.
     #[default(DEFAULT_OUTPUT_WHITELIST.iter().map(|s| (*s).to_string()).collect())]
     output_whitelist: Vec<String>,
+    oauth_token_path: Option<PathBuf>,
 }
 
 impl GithubOpenshiftSecondaryMetadataScraperSettings {
@@ -61,6 +65,7 @@ pub struct GithubOpenshiftSecondaryMetadataScraperPlugin {
 
     #[default(FuturesMutex::new(Default::default()))]
     state: FuturesMutex<State>,
+    oauth_token: Option<String>,
 }
 
 impl GithubOpenshiftSecondaryMetadataScraperPlugin {
@@ -81,9 +86,25 @@ impl GithubOpenshiftSecondaryMetadataScraperPlugin {
             )
             .context("Parsing output whitelist strings as regex")?;
 
+        let oauth_token = (&settings.oauth_token_path)
+            .clone()
+            .map(|path| {
+                std::fs::read_to_string(&path)
+                    .context(format!("Reading Oauth token from {:?}", &path))
+            })
+            .transpose()?
+            .map(|token| {
+                token
+                    .lines()
+                    .next()
+                    .map(|first_line| first_line.trim().to_owned())
+            })
+            .flatten();
+
         Ok(Self {
             settings,
             output_whitelist,
+            oauth_token,
 
             ..Default::default()
         })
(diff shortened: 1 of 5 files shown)


dist/openshift/cincinnati.yaml
```diff
@@ -64,6 +64,8 @@ objects:
                     configMapKeyRef:
                       key: gb.rust_backtrace
                       name: cincinnati
+                - mame: "CINCINNATI_GITHUB_SCRAPER_OAUTH_TOKEN_PATH"
```
```suggestion
                - name: "CINCINNATI_GITHUB_SCRAPER_OAUTH_TOKEN_PATH"
```
typo

graph-builder/src/plugins/github_openshift_secondary_metadata_scraper/plugin.rs
```diff
@@ -73,10 +94,20 @@ impl GithubOpenshiftSecondaryMetadataScraperPlugin {
...
-            .header(reqwest::header::USER_AGENT, USER_AGENT)
-            .header(reqwest::header::ACCEPT, "application/vnd.github.v3+json")
+
+        let request = {
+            let request = reqwest::Client::new()
+                .get(&url)
+                .header(reqwest::header::USER_AGENT, USER_AGENT)
+                .header(reqwest::header::ACCEPT, "application/vnd.github.v3+json");
+            if let Some(token) = &self.oauth_token {
+                request.header(reqwest::header::AUTHORIZATION, format!("token {}", token))
```
vrutkovs: Isn't it `Bearer <token>`?

Update: it is token though - https://developer.github.com/v3/#oauth2-token-sent-in-a-header

steveej:
e2e flake: level=fatal msg="failed to fetch Cluster: failed to generate asset \"Cluster\": failed to create cluster: failed to apply using Terraform"

/retest

steveej:
more flakes: level=error msg="Error: Error waiting for AMI (ami-0b526b366cf36f573) to be ready: timeout while waiting for state to become 'available' (last state: 'pending', timeout: 40m0s)"

/retest

dist/openshift/cincinnati.yaml
```diff
@@ -174,7 +176,7 @@ objects:
           volumes:
             - name: secrets
               secret:
-                secretName: cincinnati-registry-credentials
+                secretName: cincinnati-credentials
```
vrutkovs: `hack/e2e.sh` is still using `cincinnati-registry-credentials`.

I think this needs two template options:
* CREDENTIALS_SECRET_NAME (set by default to stage github token secret)
* CINCINNATI_GITHUB_SCRAPER_OAUTH_TOKEN_PATH (set to github token path by default)

and have `hack/e2e.sh` adjusted to create a dummy secret and pass `-p CINCINNATI_GITHUB_SCRAPER_OAUTH_TOKEN_PATH=""` in `oc new-app`.

This way we can have:
* CREDENTIALS_SECRET_NAME in PR e2e set to docker credentials and CINCINNATI_GITHUB_SCRAPER_OAUTH_TOKEN_PATH set to none. e2e test for PRs won't use github token for authentication
* stage / prod deploy would mount the secret and use token-based authentication

This is similar to RUST_BACKTRACE env var

steveej: Since `CINCINNATI_GITHUB_SCRAPER_OAUTH_TOKEN_PATH=""` will lead to the error, I tried to find a way to leave this variable unset in the e2e environment. Please see the latest commit and let me know if it's a viable approach or not.

vrutkovs: That would work, but wouldn't template look cleaner if Cincinnati support empty path in `CINCINNATI_GITHUB_SCRAPER_OAUTH_TOKEN_PATH`?

steveej: In my mind that's a workaround and I would prefer to indicate the absence of the secret by the absence of the environment variable, instead of special casing the `""` value.

If you feel strongly against this exercise I can be convinced otherwise.

cincinnati/src/plugins/internal/graph_builder/github_openshift_secondary_metadata_scraper/plugin.rs
```diff
@@ -94,10 +115,20 @@ impl GithubOpenshiftSecondaryMetadataScraperPlugin {
         let url = github_v3::branches_url(&self.settings.github_org, &self.settings.github_repo);
 
         trace!("Getting branches from {}", &url);
-        let bytes = reqwest::Client::new()
-            .get(&url)
-            .header(reqwest::header::USER_AGENT, USER_AGENT)
-            .header(reqwest::header::ACCEPT, "application/vnd.github.v3+json")
+
+        let request = {
+            let request = reqwest::Client::new()
```
vrutkovs: Re-use Client here

vrutkovs: Re-use existing Client instead of creating new one

vrutkovs (changes requested):

LalatenduMohanty (approved):
/hold for @vrutkovs  to take another look. 

/lgtm

vrutkovs:
/hold cancel

LGTM

PR #239, opened 2020-03-03
WIP add scripts for load testing

vrutkovs:
cc @steveeJ

### DIFF
--- a/hack/load-testing.sh
+++ b/hack/load-testing.sh
@@ -3,20 +3,25 @@
 # This script is used to load test Cincinnati instance.
 # It uses vegeta - `go get -u github.com/tsenart/vegeta`
 
-PE_URL=$(oc -n cincinnati-e2e get route cincinnati-policy-engine -o jsonpath='{.spec.host}')
-export GRAPH_URL="http://${PE_URL}/api/upgrades_info/v1/graph"
+# PE_URL=$(oc -n cincinnati-e2e get route cincinnati-policy-engine -o jsonpath='{.spec.host}')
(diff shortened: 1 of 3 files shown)


hack/load-testing.sh
```diff
@@ -0,0 +1,27 @@
...
+    echo "Testing workers ${workers}, rate ${rate} -> ${file}"
+    sed "s,GRAPH_URL,${GRAPH_URL},g" vegeta.targets | \
+      vegeta attack -format http -workers=${workers} -rate=${rate} -duration ${duration} > ${file}
+    vegeta report -type=text ${file}
+  done
+done
+
+vegeta report -type='hist[0,50ms,100ms,500ms,1s,5s,10s]' /tmp/results/*.bin
+
+sleep infinity
```
steveej, line 27: Why can't it exit?

vrutkovs: I've added to inspect pod output (this script was running in a deployment). The final version won't have this

LalatenduMohanty (changes requested):
I would suggest a directory named `load-test` inside `hack` and all the files should be moved there.

vrutkovs:
>I would suggest a directory named load-test inside hack and all the files should be moved there.

We already have `dist` for "public" files and `hack` for internal scripts (at least as I understood it). Load testing probably would be used by stage/prod scripts - why would need a new dir for that?

vrutkovs:
Also this is superseded by https://github.com/openshift/cincinnati/pull/249

PR #241, opened 2020-03-04
Sort releases by Alphabetical prior to sorting by release number.

### LINKED ISSUE #236: test plugins::openshift_secondary_metadata_parser::plugin::tests::compare_quay_result_fixture_20200220104838 ... FAILED
Plugin test failure
From git commit  7f68cf5ceadd4cee22b7cc8ae319097954c860e9

cargo clean; cargo build; cargo test
...
failures:

---- plugins::openshift_secondary_metadata_parser::plugin::tests::compare_quay_result_fixture_20200220104838 stdout ----
thread 'plugins::openshift_secondary_metadata_parser::plugin::tests::compare_quay_result_fixture_20200220104838' panicked at 'assertion failed: `(left == right)`

admiyo:
Matches the logic from
https://github.com/openshift/cincinnati-graph-data/blob/5fc8dd0825b42369de8070ecba2ae0c49d0a99d9/hack/graph-util.py#L187

Closes #236

### DIFF
--- a/cincinnati/src/plugins/internal/graph_builder/openshift_secondary_metadata_parser/plugin.rs
+++ b/cincinnati/src/plugins/internal/graph_builder/openshift_secondary_metadata_parser/plugin.rs
@@ -10,7 +10,6 @@ mod graph_data_model {
     use serde::Deserialize;
     use serde::Deserializer;
     use std::collections::HashMap;
-
     /// Represents the blocked edges files in the data repository.
     #[derive(Debug, Deserialize)]
     pub struct BlockedEdge {
@@ -386,6 +385,7 @@ impl OpenshiftSecondaryMetadataParserPlugin {
                         let mut channels_split = channels.split(',').collect::<Vec<_>>();
                         // this has to match the sorting at
                         // https://github.com/openshift/cincinnati-graph-data/blob/5fc8dd0825b42369de8070ecba2ae0c49d0a99d9/hack/graph-util.py#L187
+                        channels_split.sort_by(|a, b| a.cmp(b));
                         channels_split.sort_by(|a, b| {
                             let a_split: Vec<&str> = a.splitn(2, '-').collect();
                             let b_split: Vec<&str> = b.splitn(2, '-').collect();


cincinnati/src/plugins/internal/graph_builder/openshift_secondary_metadata_parser/plugin.rs
```diff
@@ -386,6 +385,7 @@ impl OpenshiftSecondaryMetadataParserPlugin {
                         let mut channels_split = channels.split(',').collect::<Vec<_>>();
                         // this has to match the sorting at
                         // https://github.com/openshift/cincinnati-graph-data/blob/5fc8dd0825b42369de8070ecba2ae0c49d0a99d9/hack/graph-util.py#L187
+                        channels_split.sort_by(|a, b| a.cmp(b));
```
steveej, line 388: I think it's possible to merge the two `sort_by` calls here using [Ordering::then](https://doc.rust-lang.org/std/cmp/enum.Ordering.html#method.then).

I imagine something like this (untested code):

```rust
channels_split.sort_by(|a, b| {
    let a_split: Vec<&str> = a.splitn(2, '-').collect();
    let b_split: Vec<&str> = b.splitn(2, '-').collect();

    # sort first by channel: prerelease, stable, fast, candidate, etc
    let cmp_channel = a_split[0].cmp(b_split[0]); 

    # sort by version: 4.1, 4.2, etc
    let cmp_version = a_split[1].cmp(b_split[1]);

    cmp_channel.then(cmp_version)
}
```

steveej (approved):
Thanks for this contribution!
/lgtm

I have a nitpick about merging the two sort calls. I'll leave the decision to merge as is to you.
/hold

steveej:
@admiyo In case you're not aware of this.

When the PR is approved and the _lgtm_ label is attached, you can decide to merge this PR by commenting on this PR with one line being `/hold cancel` in case you don't want to follow up on my suggestion.

steveej:
I'm taking the liberty to merge this as is now. My suggestion still stands for a follow-up refactor, but I don't think it's required as the plugin is meant for the graph-builder, which is not in the hot request-path.

/hold cancel

PR #243, opened 2020-03-04
dist/openshift: migrate graph-builder to toml config

steveej:
This will allow passing in different plugin configuration, which we may
require for e2e tests at one point.

- [x] Depends on #242.

### DIFF
--- a/hack/e2e.sh
+++ b/hack/e2e.sh
@@ -55,7 +55,18 @@ oc new-app -f template/cincinnati.yaml \
   ;
 
 # Wait for dc to rollout
-oc wait --for=condition=available --timeout=5m deploymentconfig/cincinnati
+oc wait --for=condition=available --timeout=5m deploymentconfig/cincinnati || {
+    status=$?
+    set +e -x
+
+    # Print various information about the deployment
+    oc get events
+    oc describe deploymentconfig/cincinnati
+    oc get configmap/cincinnati-configs -o yaml
+    oc logs --all-containers=true --timestamps=true --selector='app=cincinnati'
+
+    exit $status
+}
 
 # Expose services
 oc expose service cincinnati-policy-engine --port=policy-engine
--- a/hack/e2e.sh
+++ b/hack/e2e.sh
@@ -33,11 +33,25 @@ oc secrets link default ci-pull-secret --for=pull
 oc new-app -f template/cincinnati.yaml \
   -p IMAGE="${IMAGE_FORMAT%/*}/stable" \
   -p IMAGE_TAG=deploy \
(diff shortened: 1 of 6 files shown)


hack/e2e.sh
```diff
@@ -33,11 +33,30 @@ oc secrets link default ci-pull-secret --for=pull
 oc new-app -f template/cincinnati.yaml \
   -p IMAGE="${IMAGE_FORMAT%/*}/stable" \
   -p IMAGE_TAG=deploy \
-  -p GB_CINCINNATI_REPO="redhat/openshift-cincinnati-test-public-manual" \
   -p GB_CPU_REQUEST=50m \
   -p PE_CPU_REQUEST=50m \
   -p RUST_BACKTRACE="1" \
-  -p GB_REGISTRY_CREDENTIALS_PATH="" \
+  -p GB_PLUGIN_SETTINGS='
```
vrutkovs, line 39: nitpick: params are growing large, lets put these in a file and use `oc new-app --param-file`?

steveej: It's a bit hard to justify, given that we're talking about code in *hack*. We could move it to dist and take it more serious though.

vrutkovs:
Cluster failed to start

/retest

hack/e2e.sh
```diff
@@ -33,15 +33,39 @@ oc secrets link default ci-pull-secret --for=pull
...
+      name = "node-remove"
+
+      [[plugin_settings]]
+      name = "edge-add-remove"
+  ' \
   ;
 
 # Wait for dc to rollout
-oc wait --for=condition=available --timeout=5m deploymentconfig/cincinnati
+oc wait --for=condition=available --timeout=5m deploymentconfig/cincinnati || {
```
vrutkovs, line 58: Excellent, thanks for doing that! I think we also want `oc describe deploymentconfig/cincinnati` and `oc get events` here.

The only issue is that the objects in the template may eventually diverge and we won't find that out until tests start to fail

steveej: Do you care about the order of these?

hack/e2e.sh
steveej: Do you consider this a blocker?

vrutkovs: No, lets rework this later

hack/e2e.sh
vrutkovs: The order doesn't matter here really

vrutkovs:
some odd CI flake

/retest

vrutkovs (approved):
/lgtm

Thanks!

PR #244, opened 2020-03-05
e2e: allow redefining pull secret location

vrutkovs:
Customize a few parts specific to CI so that the script could be used for local testing

### DIFF
--- a/dist/Dockerfile.e2e/Dockerfile
+++ b/dist/Dockerfile.e2e/Dockerfile
@@ -24,6 +24,6 @@ ENV PATH="${PATH}:${HOME}/bin"
 
 COPY --from=builder /opt/cincinnati/bin/cincinnati-e2e-test /usr/bin
 COPY --from=builder hack/e2e.sh hack/
-COPY --from=builder dist/openshift/cincinnati.yaml template/
+COPY --from=builder dist/openshift/cincinnati.yaml dist/openshift/
 
 ENTRYPOINT ["hack/e2e.sh"]
--- a/hack/e2e.sh
+++ b/hack/e2e.sh
@@ -10,8 +10,6 @@
 #   * CINCINNATI_IMAGE (optional) - image with graph-builder and policy-engine
 #   * env var IMAGE_FORMAT (e.g `registry.svc.ci.openshift.org/ci-op-ish8m5dt/stable:${component}`)
 
-set -euo pipefail
-
 # Use CI image format by default unless CINCINNATI_IMAGE is set
 if [[ ! -z "${CINCINNATI_IMAGE}" ]]; then
   IMAGE=$(echo "${CINCINNATI_IMAGE}" | cut -d ':' -f1)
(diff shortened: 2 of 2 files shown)


vrutkovs:
sidevar OOM

/retest

dist/Dockerfile.e2e/Dockerfile
```diff
@@ -24,6 +24,6 @@ ENV PATH="${PATH}:${HOME}/bin"
 
 COPY --from=builder /opt/cincinnati/bin/cincinnati-e2e-test /usr/bin
 COPY --from=builder hack/e2e.sh hack/
-COPY --from=builder dist/openshift/cincinnati.yaml template/
+COPY --from=builder dist/openshift/cincinnati.yaml dist/openshift/
```
steveej, line 27: This change confuses me. Is there a change on CI to make this work as well?

steveej (commented):
I can see this being helpful, thanks!

When you write "local testIng", you mean against a cluster not run by CI, right? 

Just as a sidenote and nothing to block on for now, it would be nice if we had a way to test this without a full openshift cluster, e.g. if we could deploy the pod using podman.

dist/Dockerfile.e2e/Dockerfile
vrutkovs: This line needs to be coupled with https://github.com/openshift/cincinnati/pull/244/files#diff-9dac1c15b19c2b8ae927c1c5b610307aR48

steveej: Ah understood, thanks!

hack/e2e.sh
```diff
@@ -7,10 +7,25 @@
 
 # Prerequirements:
 #   * pull secret (with registry.svc.ci.openshift.org) part in `/tmp/cluster/pull-secret`
+#   * CINCINNATI_IMAGE (optional) - image with graph-builder and policy-engine
 #   * env var IMAGE_FORMAT (e.g `registry.svc.ci.openshift.org/ci-op-ish8m5dt/stable:${component}`)
```
steveej, line 11: Are we using this anywhere? If not I suggest removing it now which will make the script a bit simpler.

vrutkovs: `IMAGE_FORMAT` is passed by our CI, it is used to compose 
a pullspec to the deployment image. In dev builds users would override that with `CINCINNATI_IMAGE` - these sadly cannot be reconciled as the var names are hardcoded in cluster-launch templates on CI

hack/e2e.sh
```diff
@@ -32,7 +35,7 @@ oc project cincinnati-e2e
 oc create secret generic cincinnati-credentials --from-literal=""
 
 # Use this pull secret to fetch images from CI
-oc create secret generic ci-pull-secret --from-file=.dockercfg=/tmp/cluster/pull-secret --type=kubernetes.io/dockercfg
+oc create secret generic ci-pull-secret --from-file=.dockercfg=${PULL_SECRET} --type=kubernetes.io/dockercfg
```
LalatenduMohanty, line 37: :+1:

hack/e2e.sh
```diff
@@ -27,6 +25,7 @@ echo "IMAGE_TAG=${IMAGE_TAG}"
 # Use defined PULL_SECRET or fall back to CI location
 PULL_SECRET=${PULL_SECRET:-/tmp/cluster/pull-secret}
 
+set -euo pipefail
```
LalatenduMohanty, line 28: :+1:

steveej (approved):
/lgtm
/hold cancel

PR #248, opened 2020-03-10
e2e: reconfigure monitoring operator to scrape Cincinnati

vrutkovs:
Enable user workloads feature in monitoring operator, install Cincinnati ServiceMonitors and collect app data. This would allow us to define SLO in upstream repo and collect metrics during e2e tests / load testing - and fail the build if the performance is below a certain threshold.

This also adds several tests which ensure app metrics are sane after e2e test

### DIFF
--- a/graph-builder/tests/slo/mod.rs
+++ b/graph-builder/tests/slo/mod.rs
@@ -1,0 +1,93 @@
+use failure::{format_err, Fallible};
+use reqwest::header::{HeaderValue, AUTHORIZATION};
+use reqwest::Response;
+use serde::Deserialize;
+use std::env;
+use test_case::test_case;
+use tokio::runtime::Runtime;
+use url::Url;
+
+#[derive(Deserialize)]
+struct PromMetric {
+    value: (f64, String),
+}
+
+#[derive(Deserialize)]
+struct PromData {
+    result: Vec<PromMetric>,
+}
+
+#[derive(Deserialize)]
+struct PromResponse {
+    status: String,
+    data: PromData,
+}
+
+async fn prom_http_request(url: String, token: String) -> Fallible<Response> {
+    let header_value = format!("Bearer {}", token);
+    let authorization_header = HeaderValue::from_str(&header_value).unwrap();
+    reqwest::ClientBuilder::new()
+        .danger_accept_invalid_certs(true)
+        .build()
+        .unwrap()
+        .get(&url)
+        .header(AUTHORIZATION, authorization_header)
+        .send()
+        .await
+        .map_err(Into::into)
+}
+
+fn query_prom(runtime: &mut Runtime, url: String, token: String) -> Fallible<String> {
+    let res = runtime.block_on(prom_http_request(url, token)).unwrap();
+    let text = runtime.block_on(res.text()).unwrap();
+    println!("Prom response: {}", text);
+
+    let r: PromResponse = match serde_json::from_str(&text) {
+        Ok(r) => r,
+        Err(e) => panic!("Failed to serialize json: {}", e),
+    };
+    match r.status.as_str() {
+        "success" => match r.data.result.len() {
+            1 => Ok(r.data.result[0].value.1.clone()),
+            n => Err(format_err!("unexpected number of results: {}", n)),
+        },
+        status => Err(format_err!("incorrect query status: {}", status)),
+    }
+}
+
+// Service reports at all times
+#[test_case(r#"min_over_time(up{job="cincinnati-policy-engine"}[1h])"#, "1")]
+#[test_case(r#"min_over_time(up{job="cincinnati-graph-builder"}[1h])"#, "1")]
+// No upstream errors
+#[test_case("max_over_time(cincinnati_pe_http_upstream_errors_total[1h])", "0")]
+#[test_case("max_over_time(cincinnati_gb_graph_upstream_errors_total[1h])", "0")]
+// Use clamp_min to bring up the minimal serve duration to 0.1 seconds
+// If the quantile would produce a bigger result this test would fail
+#[test_case(
+    "clamp_min(histogram_quantile(0.90, sum(cincinnati_pe_v1_graph_serve_duration_seconds_bucket) by (le)), 0.1)",
+    "0.1"
+)]
+// At least one scrape has been performed
+#[test_case("clamp_max(cincinnati_gb_graph_upstream_scrapes_total, 1)", "1")]
+// No scrape errors
+#[test_case("cincinnati_gb_graph_upstream_errors_total", "0")]
+fn zz_check_slo(query: &'static str, expected: &'static str) {
+    let prom_url = match env::var("PROM_ENDPOINT") {
+        Ok(env) => format!("https://{}/api/v1/query", env),
+        _ => panic!("PROM_ENDPOINT unset"),
+    };
+
+    let prom_token = match env::var("PROM_TOKEN") {
+        Ok(env) => env,
+        _ => panic!("PROM_TOKEN unset"),
+    };
+
+    let mut runtime = commons::testing::init_runtime().unwrap();
+
+    let mut query_url = Url::parse(&prom_url).unwrap();
+    query_url.query_pairs_mut().append_pair("query", query);
+
+    println!("Querying {}", query_url.to_string());
+    let actual = query_prom(&mut runtime, query_url.to_string(), prom_token).unwrap();
+    pretty_assertions::assert_eq!(actual, expected)
+}
--- a/hack/e2e.sh
+++ b/hack/e2e.sh
@@ -112,8 +112,8 @@ done
 # Wait for cincinnati metrics to be recorded
 # Find out the token Prometheus uses from its serviceaccount secrets
 # and use it to query for GB build info
-PROM_ENDPOINT=$(oc -n openshift-monitoring get route thanos-querier -o jsonpath="{.spec.host}")
-PROM_TOKEN=$(oc -n openshift-monitoring get secret \
+export PROM_ENDPOINT=$(oc -n openshift-monitoring get route thanos-querier -o jsonpath="{.spec.host}")
+export PROM_TOKEN=$(oc -n openshift-monitoring get secret \
   $(oc -n openshift-monitoring get serviceaccount prometheus-k8s \
     -o jsonpath='{range .secrets[*]}{.name}{"\n"}{end}' | grep prometheus-k8s-token) \
   -o go-template='{{.data.token | base64decode}}')
@@ -126,5 +126,5 @@ for i in $(seq 1 10); do
   sleep ${DELAY}
 done
 
-# Run e2e tests
-/usr/bin/cincinnati-e2e-test
+# Run e2e tests in sequential mode to have SLO tests in the end
+env RUST_TEST_TASKS=1 /usr/bin/cincinnati-e2e-test
--- a/hack/e2e.sh
+++ b/hack/e2e.sh
@@ -44,6 +44,16 @@ done
 # Allow default serviceaccount to use CI pull secret
 oc secrets link default ci-pull-secret --for=pull
 
+# Reconfigure monitoring operator to support user workloads
+# https://docs.openshift.com/container-platform/4.3/monitoring/monitoring-your-own-services.html
+ oc -n openshift-monitoring create configmap cluster-monitoring-config --from-literal='config.yaml={"techPreviewUserWorkload": {"enabled": true}}'
+
+# Import observability template
+# ServiceMonitors are imported before app deployment to give Prometheus time to catch up with
+# metrics
+# `oc new-app` would stumble on unknown monitoring.coreos.com/v1 objects, so process and create instead
+oc process -f dist/openshift/observability.yaml -p NAMESPACE="cincinnati-e2e" | oc apply -f -
+
 # Apply oc template
 oc new-app -f dist/openshift/cincinnati.yaml \
   -p IMAGE="${IMAGE}" \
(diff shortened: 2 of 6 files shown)


hack/e2e.sh
```diff
@@ -102,5 +102,19 @@ while [ $ATTEMPTS -ge 0 ]; do
   fi
 done
 
+# Reconfigure monitoring operator to support user workloads
+# https://docs.openshift.com/container-platform/4.3/monitoring/monitoring-your-own-services.html
+cat > /tmp/config.yaml << EOF
+techPreviewUserWorkload:
+  enabled: true
+EOF
+oc -n openshift-monitoring create configmap cluster-monitoring-config --from-file=/tmp/config.yaml
```
steveej:
```suggestion
oc -n openshift-monitoring create configmap cluster-monitoring-config \
  --from-file=<(cat <<'EOF'
  techPreviewUserWorkload:
    enabled: true
EOF
)
```

hack/e2e.sh
```diff
@@ -102,5 +102,19 @@ while [ $ATTEMPTS -ge 0 ]; do
   fi
 done
 
+# Reconfigure monitoring operator to support user workloads
+# https://docs.openshift.com/container-platform/4.3/monitoring/monitoring-your-own-services.html
+cat > /tmp/config.yaml << EOF
+techPreviewUserWorkload:
+  enabled: true
+EOF
```
steveej:
```suggestion
```

steveej (commented):
This looks interesting!

Preliminary review so just a few suggestions.

hack/e2e.sh
vrutkovs: I don't like the fact that we create a file in `/tmp`, but I think I don't like bash version more :) Its really tough to debug these kind of issues.

Do you mind if I keep the tmp file version?

LalatenduMohanty:
> e2e runs can analyzed using https://promecieus.svc.ci.openshift.org

Did not understand this part. Do you mean create a web link to analyze the metric from an e2e run?

LalatenduMohanty:
@vrutkovs I guess you have plan to add tests to check the metric generated in the e2e tests as per SLO in near future?

vrutkovs:
>Did not understand this part. Do you mean create a web link to analyze the metric from an e2e run?

Yes, this tool can be used to analyze Prometheus data from a test run

>I guess you have plan to add tests to check the metric generated in the e2e tests as per SLO in near future?

Out of scope for this PR

hack/e2e.sh
```diff
@@ -44,6 +44,20 @@ done
 # Allow default serviceaccount to use CI pull secret
 oc secrets link default ci-pull-secret --for=pull
 
+# Reconfigure monitoring operator to support user workloads
+# https://docs.openshift.com/container-platform/4.3/monitoring/monitoring-your-own-services.html
+cat > /tmp/config.yaml << EOF
+techPreviewUserWorkload:
+  enabled: true
+EOF
+oc -n openshift-monitoring create configmap cluster-monitoring-config --from-file=/tmp/config.yaml
```
wking: To avoid both the `/tmp` file and [Bash gymnastics][1], how about:

```sh
 oc -n openshift-monitoring create configmap cluster-monitoring-config '--from-literal=config.yaml={"techPreviewUserWorkload": {"enabled": true}}'
```

?

[1]: https://github.com/openshift/cincinnati/pull/248#discussion_r390982511

hack/e2e.sh
```diff
@@ -102,5 +116,24 @@ while [ $ATTEMPTS -ge 0 ]; do
...
+PROM_TOKEN=$(oc -n openshift-monitoring get secret \
+  $(oc -n openshift-monitoring get serviceaccount prometheus-k8s \
+    -o jsonpath='{range .secrets[*]}{.name}{"\n"}{end}' | grep prometheus-k8s-token) \
+  -o jsonpath='{.data.token}' \
+  -o go-template='{{.data.token | base64decode}}')
+
+ATTEMPTS=10
+DELAY=30
+while [ $ATTEMPTS -ge 0 ]; do
+  curl -kLs -H "Authorization: Bearer ${PROM_TOKEN}" https://${PROM_ENDPOINT}/api/v1/query\?query\=cincinnati_gb_build_info | grep "metric" && break
```
wking: nit: `"https://${PROM_ENDPOINT}/api/v1/query?query=cincinnati_gb_build_info"` to avoid the backslashes and protect against the unlikely case that `PROM_ENDPOINT` has some shell-sensitive characters.

Also no need for the quotes in `"metric"`.

hack/e2e.sh
```diff
@@ -102,5 +116,24 @@ while [ $ATTEMPTS -ge 0 ]; do
...
 done
 
+# Wait for cincinnati metrics to be recorded
+# Find out the token Prometheus uses from its serviceaccount secrets
+# and use it to query for GB build info
+PROM_ENDPOINT=$(oc -n openshift-monitoring get route thanos-querier -o jsonpath="{.spec.host}")
+PROM_TOKEN=$(oc -n openshift-monitoring get secret \
+  $(oc -n openshift-monitoring get serviceaccount prometheus-k8s \
+    -o jsonpath='{range .secrets[*]}{.name}{"\n"}{end}' | grep prometheus-k8s-token) \
+  -o jsonpath='{.data.token}' \
```
wking: Isn't this clobbered by the subsequent `go-template`?  Can we drop this line?

hack/e2e.sh
```diff
@@ -102,5 +116,24 @@ while [ $ATTEMPTS -ge 0 ]; do
...
+  -o go-template='{{.data.token | base64decode}}')
+
+ATTEMPTS=10
+DELAY=30
+while [ $ATTEMPTS -ge 0 ]; do
```
wking: possibly simpler than the `while` would be:

```sh
for i in $(seq 10); do
```

which would let you drop the later `ATTEMPTS=$((ATTEMPTS-1))`.

hack/e2e.sh
vrutkovs: >Also no need for the quotes in "metric".

I added this intentionally, so that it would be highlighted in most editor and the command would make a bit more sense :)

hack/e2e.sh
```diff
@@ -89,17 +99,28 @@ PE_URL=$(oc get route cincinnati-policy-engine -o jsonpath='{.spec.host}')
...
-
-while [ $ATTEMPTS -ge 0 ]; do
+for i in $(seq 10); do
   CODE=$(curl -s -o /dev/null -w "%{http_code}" --header 'Accept:application/json' "${GRAPH_URL}?channel=a")
   if [ "${CODE}" == "200" ]; then
     break
-  else
-    sleep ${DELAY}
-    ATTEMPTS=$((ATTEMPTS-1))
-  fi
```
wking, line 103: [e2e][1]:

```
hack/e2e.sh: line 109: syntax error near unexpected token `done'
```

You need to keep this closing `fi`.

[1]: https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_cincinnati/248/pull-ci-openshift-cincinnati-master-e2e/363/build-log.txt

hack/e2e.sh
vrutkovs: Trevor has suggest the most elegant way

LalatenduMohanty:
> Yes, this tool can be used to analyze Prometheus data from a test run

Ok, this is new to me. Can you please give me the steps how to do that?

LalatenduMohanty (changes requested):
Did not realize this PR has two commits but it should be just one commit. I do not see any benefit of keeping https://github.com/openshift/cincinnati/pull/248/commits/e56381d492be357f5b793a37d78c43ca010be675 as a separate commit.

vrutkovs:
>Did not realize this PR has two commits but it should be just one commit.

We might want to keep the observability template but avoid applying it in e2e

LalatenduMohanty (approved):
/hold for @wking  take look and approve
/lgtm

vrutkovs:
/hold cancel

Needs another round of reviews - added user metrics verification

hack/e2e.sh
```diff
@@ -44,6 +44,16 @@ done
 # Allow default serviceaccount to use CI pull secret
 oc secrets link default ci-pull-secret --for=pull
 
+# Reconfigure monitoring operator to support user workloads
+# https://docs.openshift.com/container-platform/4.3/monitoring/monitoring-your-own-services.html
+ oc -n openshift-monitoring create configmap cluster-monitoring-config '--from-literal=config.yaml={"techPreviewUserWorkload": {"enabled": true}}'
```
steveej:
```suggestion
 oc -n openshift-monitoring create configmap cluster-monitoring-config --from-literal='config.yaml={"techPreviewUserWorkload": {"enabled": true}}'
```

hack/e2e.sh
```diff
@@ -89,18 +99,30 @@ PE_URL=$(oc get route cincinnati-policy-engine -o jsonpath='{.spec.host}')
 export GRAPH_URL="http://${PE_URL}/api/upgrades_info/v1/graph"
 
 # Wait for route to become available
-ATTEMPTS=10
 DELAY=10
-
-while [ $ATTEMPTS -ge 0 ]; do
+for i in $(seq 10); do
```
steveej: I'd prefer being explicit here:

```suggestion
for i in $(seq 1 10); do
```

hack/e2e.sh
```diff
@@ -89,18 +99,30 @@ PE_URL=$(oc get route cincinnati-policy-engine -o jsonpath='{.spec.host}')
...
+# Find out the token Prometheus uses from its serviceaccount secrets
+# and use it to query for GB build info
+export PROM_ENDPOINT=$(oc -n openshift-monitoring get route thanos-querier -o jsonpath="{.spec.host}")
+export PROM_TOKEN=$(oc -n openshift-monitoring get secret \
+  $(oc -n openshift-monitoring get serviceaccount prometheus-k8s \
+    -o jsonpath='{range .secrets[*]}{.name}{"\n"}{end}' | grep prometheus-k8s-token) \
+  -o go-template='{{.data.token | base64decode}}')
+
+DELAY=30
+for i in $(seq 10); do
```
steveej:
```suggestion
for i in $(seq 1 10); do
```

hack/e2e.sh
```diff
@@ -89,18 +99,30 @@ PE_URL=$(oc get route cincinnati-policy-engine -o jsonpath='{.spec.host}')
...
+  curl -kLs -H "Authorization: Bearer ${PROM_TOKEN}" "https://${PROM_ENDPOINT}/api/v1/query?query=cincinnati_gb_build_info" | grep "metric" && break
+
+  sleep ${DELAY}
 done
 
-# Run e2e tests
-/usr/bin/cincinnati-e2e-test
+# Run e2e tests in sequential mode to have SLO tests in the end
+env RUST_TEST_TASKS=1 /usr/bin/cincinnati-e2e-test
```
steveej, line 130: Do you actually need them to run at the end, or just not in parallel to any others?

steveej (commented):
I think that we eventually want the slo and e2e tests outside of the graph-builder crate, because they're not specific to that. Do you feel like looking into if we can have them in the top-level workspace crate?

I have not reviewed the Rust portion of the test yet.

graph-builder/tests/slo/mod.rs
```diff
@@ -0,0 +1,93 @@
...
+}
+
+#[derive(Deserialize)]
+struct PromResponse {
+    status: String,
+    data: PromData,
+}
+
+fn query_prom(url: String, token: String) -> Fallible<String> {
+    let mut runtime = commons::testing::init_runtime().unwrap();
```
steveej: I suggest moving the runtime variable to the test itself, and passing in a `&mut` of it to this function, or even better, make this an `async fn` and return a future.

steveej (commented):
I started building a rudimentary prometheus query client a while ago at https://github.com/steveeJ-forks/openshift-cincinnati/blob/043f07156a7a417c43977da553a9d6e79161569b/prometheus-query/src/v1/mod.rs. Would you be interested in having that in a repo and sharing it?

hack/e2e.sh
vrutkovs: We do - its pointless to collect metrics if there are more tests planned.

Hopefully this is a temporary solution - initially I tried making different binaries for that, but Rust mixes up binary names so Dockerfile.e2e becomes a mess.

vrutkovs:
>step cargo-test failed: test "cargo-test" failed: could not create watcher for pod: Get https://172.30.0.1:443/api/v1/namespaces/ci-op-wm24qdch/pods?fieldSelector=metadata.name%3Dcargo-test&watch=true: unexpected EOF

/test cargo-test

graph-builder/tests/slo/mod.rs
```diff
@@ -0,0 +1,93 @@
...
+        "success" => match r.data.result.len() {
+            1 => Ok(r.data.result[0].value.1.clone()),
+            n => Err(format_err!("unexpected number of results: {}", n)),
+        },
+        status => Err(format_err!("incorrect query status: {}", status)),
+    }
+}
+
+// Service reports at all times
+#[test_case("min_over_time(up{job=\"cincinnati-policy-engine\"}[1h])", "1")]
```
steveej:
```suggestion
#[test_case(r#"min_over_time(up{job="cincinnati-policy-engine"}[1h])"#, "1")]
```

hack/e2e.sh
steveej: Please explain how we know that the SLO tests are at the end of the sequence now as it's not obvious to me.

vrutkovs: Rust sorts tests alphabetically and starts them simultaneously. So in order to run SLO tests last in this PR:
* make sure the package is named "slo" so that it would run after "e2e"
* use RUST_TEST_TASKS=1 to run tests sequentially - meaning e2e tests would start one by one and then start SLO verification tests

steveej: I see, thanks! Let's hope it keeps the alphabetical ordering until we figure out a more explicit way for the ordering.

hack/e2e.sh
```diff
@@ -89,18 +99,30 @@ PE_URL=$(oc get route cincinnati-policy-engine -o jsonpath='{.spec.host}')
...
+# and use it to query for GB build info
+export PROM_ENDPOINT=$(oc -n openshift-monitoring get route thanos-querier -o jsonpath="{.spec.host}")
+export PROM_TOKEN=$(oc -n openshift-monitoring get secret \
+  $(oc -n openshift-monitoring get serviceaccount prometheus-k8s \
+    -o jsonpath='{range .secrets[*]}{.name}{"\n"}{end}' | grep prometheus-k8s-token) \
+  -o go-template='{{.data.token | base64decode}}')
+
+DELAY=30
+for i in $(seq 1 10); do
+  curl -kLs -H "Authorization: Bearer ${PROM_TOKEN}" "https://${PROM_ENDPOINT}/api/v1/query?query=cincinnati_gb_build_info" | grep "metric" && break
```
steveej: Since `set -euo pipefail` is set above the test will fail with this call. Is this intended?

graph-builder/tests/slo/mod.rs
```diff
@@ -0,0 +1,93 @@
...
+// No upstream errors
+#[test_case("max_over_time(cincinnati_pe_http_upstream_errors_total[1h])", "0")]
+#[test_case("max_over_time(cincinnati_gb_graph_upstream_errors_total[1h])", "0")]
+// Use clamp_min to bring up the minimal serve duration to 0.1 seconds
+// If the quantile would produce a bigger result this test would fail
+#[test_case(
+    "clamp_min(histogram_quantile(0.90, sum(cincinnati_pe_v1_graph_serve_duration_seconds_bucket) by (le)), 0.1)",
+    "0.1"
+)]
+// At least one scrap has been performed
```
steveej:
```suggestion
// At least one scrape has been performed
```

steveej (commented):
Some more comments.

Now I also had a look at the `ServiceMonitor` which I don't know much about. For the e2e tests, where would we receive the alerts? And is there any way we can share the rule and alert definitions with other deployments?

vrutkovs:
>For the e2e tests, where would we receive the alerts?

In stage/prod we have alertmanager configured for that - and alerts to trigger whenever metrics go crazy. In e2e AlertManager is not configured at all, so Alerts are useless. Its not worth adding those here anyway, as triggering alerts is core Prometheus functionality and we should not be testing that

hack/e2e.sh
vrutkovs: Oh, good catch. Pushed a fix for this

steveej:
/lgtm

> Its not worth adding those here anyway, as triggering alerts is core Prometheus functionality and we should not be testing that

I see, It sounds like we want to remove the alert definitions from this PR then, right?

vrutkovs:
>I see, It sounds like we want to remove the alert definitions from this PR then, right?

I added those to have an upstream version of stage/prod SLO alerts.

On the second thought these are still useful - we could verify that no Cincinnati alerts are firing, as prometheus exposes a metric for firing alerts - in the follow up PRs

vrutkovs:
/hold

I'll squash the commits first

steveej (approved):
/lgtm

PR #249, opened 2020-03-17
Move e2e tests out of graph-builder, check SLO after load-testing

vrutkovs:
Reworks e2e test to be separate from graph-builder unit tests:

* Create a new e2e crate
* These tests are not executed during `cargo-test`, instead the binaries are copied to `e2e` image
* e2e test configures cluster Prometheus to support user workloads and starts e2e tests using a prebuilt binary
* once these have passed load-testing using vegeta is started
* after 30 secs (time for Prometheus to fetch metrics from Cincinnati) another test is started to verify SLO are not exceeded

TODO:
* [x] Wait for 5 sec before collecting SLO metrics
  After e2e tests are done Prometheus may have not yet collected their metrics, which causes e2e to flake
* [x] Use https://github.com/steveeJ-forks/openshift-cincinnati/blob/043f07156a7a417c43977da553a9d6e79161569b/prometheus-query/src/v1/mod.rs
to query Prometheus
* [x] Split into 2 binaries:
  * e2e tests
  * SLO verification
  so that we could put load-testing in between and use SLO verification binary for stage / prod tests (is it useful though?)
* [x] Update `graph-builder/Cargo.toml` and remove e2e packages
* [ ] Ensure `prometheus_query` tests are executed


vrutkovs:
commits are still a mess thus putting it on hold, but its ready for review

/hold
/cc @steveeJ @wking

vrutkovs:
`kube_pod_container_status_restarts_total{namespace="cincinnati-e2e"}` shows cincinnati container restarted twice due to OOMKills

wking:
No comment from me on the Rust (where I'm happy to take your word for it ;), but the rest looks good to me.

e2e/e2e.rs
```diff
@@ -0,0 +1,2 @@
+#[cfg(test)]
+mod tests;
```
steveej, line 2: The tests directory doesn't have a `mod.rs`. My unverified assumption has been that that's needed, so just want to hint at this to ensure the tests are actually run.

prometheus-query/src/v1/queries/instant.rs
```diff
@@ -0,0 +1,122 @@
...
+            .and_then(|mut response| response.json().map_err(Into::into))
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    #[cfg(feature = "test-net-private")]
+    #[ignore]
+    #[test]
+    fn query_infogw() -> Fallible<()> {
```
steveej: Are you interested in fixing and activating this, or some other, test?

vrutkovs: Oh, right, certainly - added a TODO

e2e/e2e.rs
vrutkovs: https://github.com/openshift/cincinnati/pull/249/commits/7dadb9960ae88c23249c521f7e2e223ed5fdf6c0 reworks this to make e2e tests binary have `e2e-...` name (previously it had `mod-...`).

[e2e test results](https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_cincinnati/249/pull-ci-openshift-cincinnati-master-e2e/385/artifacts/e2e/container-logs/test.log):
```
 tests
test e2e_channel_success::channel_b_amd64 ... ok
test e2e_channel_success::channel_a_amd64 ... ok
test e2e_channel_success::channel_test_amd64 ... ok
```

prometheus-query/src/v1/queries/instant.rs
steveej: A TODO on the PR level means that fixing and activating these tests are in scope of this PR, right?

vrutkovs: So tests compiles, but it requires PROMETHEUS_TOKEN to function. 

Not sure how to proceed - I guess it should be moved to e2e suite and use in-cluster prometheus instead of production URL?

steveej: > I guess it should be moved to e2e suite and use in-cluster prometheus instead of production URL?

What controls the life-cycle of the in-cluster prometheus? Will it only include data from the current test-run -- which would be ideal? If so, that sounds like an ideal solution to me.

vrutkovs: >What controls the life-cycle of the in-cluster prometheus?

It would live as long as the cluster itself

>Will it only include data from the current test-run

Yep

I guess the safest option would be:
a) move to e2e
b) re-use `PROM_ENDPOINT`/`PROM_TOKEN` env vars for this test
c) query something simple like `up{job="apiserver"}` there

vrutkovs:
e2e flakes due to graph-builder miscommunication
>error fetching graph: failed to fetch upstream graph: error sending request for url (http://localhost:8080/v1/graph): connection error: Connection reset by peer (os error 104)

/retest

prometheus-query/src/v1/queries/instant.rs
vrutkovs: This would require Prometheus to return expected timestamps and labels (like node names), which are being generated every e2e run. So the test currently compiles, but I'm not running it in e2e suite

steveej: Could we come up with a query which has an at least partially static result, so we can only verify that part?

vrutkovs: We'd still have to rip out timestamps, so it would take more work to enable it

e2e/tests/prometheus_query.rs
```diff
@@ -0,0 +1,66 @@
...
+
+    let client = Client::builder()
+        .api_base(Some(api_base))
+        .access_token(Some(token))
+        .build()?;
+
+    let query = r#"up{job="apiserver"}"#;
+
+    let expected_json = r#"
+        {"status":"success","data":{"resultType":"vector","result":[{"metric":{"version":"4.0.0-0.3"},"value":[1552056334,"0.04081632653061224"]},{"metric":{"version":"4.0.0-0.alpha-2019-03-05-054505"},"value":[1552056334,"0.056451612903225805"]},{"metric":{"version":"4.0.0-0.6"},"value":[1552056334,"0.07692307692307693"]},{"metric":{"version":"4.0.0-0.7"},"value":[1552056334,"0.04519774011299435"]},{"metric":{"version":"4.0.0-0.5"},"value":[1552056334,"0.0625"]}]}}
```
steveej, line 21: I understand this is a fixture which we can't reproduce at runtime because the versions and timestamps will vary.

e2e/tests/prometheus_query.rs
```diff
@@ -0,0 +1,66 @@
...
+            std::cmp::Ordering::Greater
+        } else {
+            std::cmp::Ordering::Equal
+        }
+    }
+
+    expected_result.sort_by(sort_by_version);
+    result.sort_by(sort_by_version);
+
+    assert_eq!(expected_result, result);
```
steveej: I suggest we skip this assertion, which leaves us with an integration tests checking for a successful query and deserialization, and failing if we somehow break that.

vrutkovs: Done. In any case this test is creating a different test binary - `prometheus-query-...` and it doesn't get copied to `e2e` container, so its effectively disabled

steveej: Can we copy it or move it somewhere else so it's executed? The situation I want to prevent is where this codepath breaks untested, subsequently sabotages other tests, and we don't have a clue what's going on.

wking:
Needs a rebase for `Cargo.lock` conflicts.

cincinnati/Cargo.toml
```diff
@@ -44,7 +44,6 @@ twoway = "^0.2"
 walkdir = "2.3.1"
 pretty_assertions = "0.6.1"
 test-case = "1.0.0"
-prettydiff = {}
```
steveej: We need this there. See https://github.com/openshift/cincinnati/pull/232/commits/b27bc65effffb85c2951f6196ac7d7fd3db57999 for an actual fix.

vrutkovs: Oh, okay, that confused my local cargo build

e2e/tests/prometheus_query.rs
```diff
@@ -0,0 +1,69 @@
+use chrono;
+use failure::{Fallible, ResultExt};
+use prometheus_query::v1::queries::{QueryData, QueryResult, VectorResult};
+use prometheus_query::v1::Client;
+#[ignore]
```
steveej: This test looks like we can unignore it now!

e2e/tests/prometheus_query.rs
```diff
@@ -0,0 +1,69 @@
...
+        };
+
+    let mut runtime = commons::testing::init_runtime().unwrap();
+
+    let mut result: Vec<VectorResult> = match runtime.block_on(client.query(
+        query.to_string(),
+        Some(chrono::DateTime::from_utc(
+            chrono::NaiveDateTime::from_timestamp(1_552_056_334, 675 * 100),
+            chrono::Utc,
+        )),
```
steveej: AFAIU passing `Some(..)` here doesn't make sense now.

```suggestion
        None,
```

e2e/tests/prometheus_query.rs
```diff
@@ -0,0 +1,61 @@
+use failure::{Fallible, ResultExt};
+use prometheus_query::v1::queries::{QueryData, QueryResult, VectorResult};
+use prometheus_query::v1::Client;
+#[test]
```
steveej, line 6: missing newline above here

e2e/tests/prometheus_query.rs
```diff
@@ -0,0 +1,61 @@
...
+            std::cmp::Ordering::Greater
+        } else {
+            std::cmp::Ordering::Equal
+        }
+    }
+
+    expected_result.sort_by(sort_by_version);
+    result.sort_by(sort_by_version);
+
+    // TODO(vrutkovs): remove timestamps from expected_result/result before uncommenting assert
```
steveej, line 60: I think this is fine to not do within this PR. The actual query part of the test ensures that the result is successful which already is a good indicator whether the requests are working or not.

e2e/tests/prometheus_query.rs
steveej, line 6: I'd also add a new feature for the prom-query e2e tests, so we have easy control over which of them we want to build and run locally.

```suggestion

#[cfg(feature = "test-e2e-prom-query")]
#[test]
```

e2e/tests/prometheus_query.rs
vrutkovs: Yeah, I'll defer that

e2e/tests/prometheus_query.rs
vrutkovs: Done

hack/e2e.sh
```diff
@@ -112,19 +112,37 @@ done
...
+export RUST_BACKTRACE="1"
+
+# Run e2e tests
+/usr/bin/cincinnati-e2e-test
+
+# Ensure prometheus_query tests are executed
+/usr/bin/cincinnati-prometheus_query-test
+
+# Run load-testing script
+bash /usr/local/bin/load-testing.sh
```
steveej: * Why the extra `bash` here?
* Can the load testing fail somehow?

steveej (commented):
Thanks for this! I took a complete review pass.

e2e/e2e.rs
```diff
@@ -0,0 +1,3 @@
...
+pub extern crate chrono;
```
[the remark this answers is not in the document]
vrutkovs: Ah, yes, we shouldn't have sticked to an old input
Apparently its not necessary, removed. Once `prometheus_query` would be fixed it would be enabled there

hack/e2e.sh
```diff
@@ -112,19 +112,37 @@ done
...
+
 export PROM_TOKEN=$(oc -n openshift-monitoring get secret \
   $(oc -n openshift-monitoring get serviceaccount prometheus-k8s \
     -o jsonpath='{range .secrets[*]}{.name}{"\n"}{end}' | grep prometheus-k8s-token) \
   -o go-template='{{.data.token | base64decode}}')
 
 DELAY=30
 for i in $(seq 1 10); do
-  PROM_OUTPUT=$(curl -kLs -H "Authorization: Bearer ${PROM_TOKEN}" "https://${PROM_ENDPOINT}/api/v1/query?query=cincinnati_gb_build_info" || continue)
+  PROM_OUTPUT=$(curl -kLs -H "Authorization: Bearer ${PROM_TOKEN}" "${PROM_ENDPOINT}/api/v1/query?query=cincinnati_gb_build_info" || continue)
```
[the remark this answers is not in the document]
vrutkovs: Ah, `|| continue` must be outside of `$()`:

```
>   set -x -euo pipefail
>   for i in `seq 0 10`; do
>     echo $i
>     var=$(cat asdf) || continue
>     exit 1
>   done
> )
++ seq 0 10
+ for i in `seq 0 10`
+ echo 0
0
++ cat asdf
cat: asdf: No such file or directory
+ var=
+ continue
+ for i in `seq 0 10`
+ echo 1
1
++ cat asdf
cat: asdf: No such file or directory
+ var=
+ continue
+ for i in `seq 0 10`
+ echo 2
2
...
```

hack/e2e.sh
```diff
@@ -112,19 +112,37 @@ done
...
   grep "metric" <<< "${PROM_OUTPUT}" && break
```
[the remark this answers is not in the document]
vrutkovs: Fixed:
```
(
  set -x -euo pipefail
  for i in `seq 0 10`; do
    foo="bar$i"
    echo $foo
    grep "bar4" <<< $foo || continue && break
    exit 1
  done
)

++ seq 0 10
+ for i in `seq 0 10`
+ foo=bar0
+ echo bar0
bar0
+ grep --color=auto bar4
+ continue
+ for i in `seq 0 10`
+ foo=bar1
+ echo bar1
bar1
+ grep --color=auto bar4
+ continue
+ for i in `seq 0 10`
+ foo=bar2
+ echo bar2
bar2
+ grep --color=auto bar4
+ continue
+ for i in `seq 0 10`
+ foo=bar3
+ echo bar3
bar3
+ grep --color=auto bar4
+ continue
+ for i in `seq 0 10`
+ foo=bar4
+ echo bar4
bar4
+ grep --color=auto bar4
bar4
+ break
```

hack/e2e.sh
vrutkovs: * Removed extra bash as `load-testing.sh` is not executable
* It may fail only if `vegeta` doesn't get installed. In other cases `vegeta report` would return 0, even if there is no data set

prometheus-query/Cargo.toml
```diff
@@ -0,0 +1,27 @@
...
+log = "^0.4.3"
+chrono = "^0.4.0"
+
+[dev-dependencies]
+env_logger = "^0.6.0"
+tokio = "^0.1.14"
+
+[features]
+test-net = []
+test-net-private = []
```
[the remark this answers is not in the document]
vrutkovs: Right, removed those - tests are now in `e2e` crate

prometheus-query/src/lib.rs
```diff
@@ -0,0 +1,18 @@
+//! Asynchronous library for the Prometheus HTTP API
+//!
+//! See https://github.com/prometheus/prometheus/blob/9de0ab3c8a32f8e09ab68f793dab0c76ec3e93d0/docs/querying/api.md#http-api
+
+#[macro_use]
+extern crate failure;
```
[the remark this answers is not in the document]
vrutkovs: Odd, my RLS complains about unused imports if I do that.

I'd rather defer that now

hack/load-testing.sh
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/env bash
+#
+# This script is used to load test Cincinnati instance.
+# It uses vegeta - `go get -u github.com/tsenart/vegeta`
+
+# PE_URL=$(oc -n cincinnati-e2e get route cincinnati-policy-engine -o jsonpath='{.spec.host}')
+# export GRAPH_URL="http://${PE_URL}/api/upgrades_info/v1/graph"
+
+
+mkdir /tmp/results
```
steveej: Should we use a directory created by `mktemp -d` instead? Otherwise subsequent runs could clobber the previous ones.

prometheus-query/src/v1/mod.rs
```diff
@@ -0,0 +1,110 @@
...
+            None => bail!("api_base not set"),
+        };
+        let client = Client {
+            api_base,
+            danger_accept_invalid_certs: self.danger_accept_invalid_certs,
+            hclient,
+            token: self.token,
+        };
+
+        trace!("built client '{:?}'", &client);
```
steveej: This will expose the given token which we probably don't want.

dist/cargo_test.sh
```diff
@@ -7,12 +7,14 @@ cargo_test_flags["cincinnati"]="--features test-net"
...
 cargo_test_flags["graph-builder"]="--features test-net"
 cargo_test_flags["policy-engine"]=""
+cargo_test_flags["prometheus-query"]=""
 cargo_test_flags["quay"]="--features test-net"
 
 if [[ -n "${CINCINNATI_TEST_CREDENTIALS_PATH}" && -n "${CINCINNATI_TEST_QUAY_API_TOKEN_PATH}" ]]; then
     echo Secrets available, enabling private tests...
     cargo_test_flags["cincinnati"]+=",test-net-private"
     cargo_test_flags["graph-builder"]+=",test-net-private"
+    cargo_test_flags["prometheus-query"]="--features test-net-private"
```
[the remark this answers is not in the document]
steveej: The `test-net-private` feature is gone as of https://github.com/openshift/cincinnati/pull/249/commits/21a18bc329384ad8be2906f8506dc06ae2a89ac2 so it can be removed here.

```suggestion
```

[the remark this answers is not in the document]
vrutkovs: Removed

dist/cargo_test.sh
```diff
@@ -7,12 +7,14 @@ cargo_test_flags["cincinnati"]="--features test-net"
 cargo_test_flags["commons"]=""
 cargo_test_flags["graph-builder"]="--features test-net"
 cargo_test_flags["policy-engine"]=""
+cargo_test_flags["prometheus-query"]=""
```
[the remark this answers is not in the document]
vrutkovs: Removed it - the only remaining prometheus_query test in `prometheus_query/src/v1/queries/mod.rs` doesn't need any special flags

prometheus-query/src/v1/mod.rs
vrutkovs: Its acceptable for CI usecase - cluster gets destroyed before artifacts are uploaded to GCS

steveej: The crate and the builder aren't CI specific. The caller can do the trace of the built client if we really need it, but the builder shouldn't do any log output itself.

steveej (changes requested):

steveej (approved):
Thanks for the hard work on this one!
This should be ready to land after squashing the commits.

vrutkovs:
/hold cancel

Refactored commits

steveej (approved):
/lgtm

steveej (approved):
/lgtm

PR #250, opened 2020-03-19
openshift_secondary_metadata_parser: add 20200319.204124 test fixtures

steveej:
This adds new test fixtures to investigate a potential bug discovered while working at #232.


steveej:
@vrutkovs the SLO tests in the e2e job fail due to authentication issues. I don't know if this could be a flake or not. PTAL at https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/pr-logs/pull/openshift_cincinnati/250/pull-ci-openshift-cincinnati-master-e2e/391

vrutkovs:
Flake:
```
deploymentconfig.apps.openshift.io/cincinnati condition met
route.route.openshift.io/cincinnati-policy-engine exposed
The connection to the server api.ci-op-jjij2m9s-90ac5.origin-ci-int-aws.dev.rhcloud.com:6443 was refused - did you specify the right host or port?

running 10 tests
```
Token was unset here

/retest

steveej:
/hold

The CI is currently broken due to an issue in the test script, so this is very likely a false negative.

cincinnati/src/plugins/internal/graph_builder/openshift_secondary_metadata_parser/plugin.rs
```diff
@@ -265,46 +265,88 @@ impl OpenshiftSecondaryMetadataParserPlugin {
...
-                    let arch =
-                        // Special case for a few cases where the "s390x" arch was encoded with '-' instead of '+'
-                        if to.pre == vec![semver::Identifier::AlphaNumeric("s390x".to_string())] {
-                            "s390x"
+                // Evaluate the architectures to block
+                let target_versions = {
+                    let mut to = blocked_edge.to.clone();
+
+                    if !to.build.is_empty() {
+                        // Don't change architecture if given
```
vrutkovs: "Don't update architecture if its explicitely defined"?

cincinnati/src/plugins/internal/graph_builder/openshift_secondary_metadata_parser/plugin.rs
```diff
@@ -265,46 +265,88 @@ impl OpenshiftSecondaryMetadataParserPlugin {
...
+                                            self.settings.key_prefix, "previous.remove_regex"
+                                        ),
+                                        blocked_edge.from.to_string(),
+                                    );
+                                }
+                                Err(e) => warn!("{}", e),
+                            };
+                        }
+                        None => {
+                            info!("Release with version {} not found in graph", to);
```
vrutkovs, line 346: I don't think we expect that to happen (`blocked_edges` should not mention invalid nodes). Let throw a warning here?

cincinnati/src/plugins/internal/node_remove.rs
```diff
@@ -85,19 +85,23 @@ mod tests {
             let metadata: TestMetadata = vec![
                 (
                     0,
-                    hashmap! { format!("{}.{}", key_prefix, key_suffix) => String::from("true") },
+                    hashmap! { format!("{}.{}", key_prefix, key_suffix) => String::from("true") }
```
vrutkovs: Shouldn't we use `btreemap!` here?

vrutkovs (changes requested):

cincinnati/src/plugins/internal/graph_builder/openshift_secondary_metadata_parser/plugin.rs
steveej: This is expected because some blocked edges are given without an architecture, hence we look for the given version for all architectures which may not exist. I'm fine changing this to be warning anyway.

cincinnati/src/plugins/internal/node_remove.rs
steveej: I honestly didn't look for its existence. `anymap!` would be best :-D

steveej:
I'm currently refactoring this to make the pretty comparison reusable.

/hold

LalatenduMohanty:
@steveeJ What is the performance impact of this change?

cincinnati/src/plugins/internal/node_remove.rs
steveej: I decided to remove the macro invocation and the maplit dependency altogether.

steveej:
> What is the performance impact of this change?

That's an interesting question which I've not pursued yet. Maybe the best option is to switch the map/set implementations only for tests, as that's where we rely on sorting. I'll try that out and push if I think it's viable.

cincinnati/src/plugins/internal/graph_builder/openshift_secondary_metadata_parser/plugin.rs
```diff
@@ -509,21 +510,16 @@ mod tests {
...
-        pretty_assertions::assert_eq!(graph_result_sorted, graph_expected_sorted);
+        if let Err(e) = compare_graphs_verbose(
+            graph_expected,
+            graph_result,
+            &[
+                "io.openshift.upgrades.graph.previous.remove",
+                "io.openshift.upgrades.graph.previous.remove_regex",
+            ],
+        ) {
+            panic!("{}", e);
```
vrutkovs, line 563: Not sure why we need to panic here - can't this func just return `Err`?

steveej: This function is the test itself. Usually we use `unwrap()` also causes a panic, but it doesn't print the error message nicely, e.g. it doesn't even respect newlines but prints `\n` instead. `panic`ing here explicitly which respects the formatting, even color and such.

steveej:
I consider this state complete now. Now it's up to the reviewers ;-)

/hold cancel

cincinnati/src/plugins/internal/graph_builder/openshift_secondary_metadata_parser/plugin.rs
vrutkovs: Oh, I see, okay

PR #252, opened 2020-03-20
cincinnati: trace adding EdgeIndexes

vrutkovs:
Ref: https://issues.redhat.com/browse/OTA-156

This also fixes a small typo in the comment.

### DIFF
--- a/cincinnati/src/lib.rs
+++ b/cincinnati/src/lib.rs
@@ -22,7 +22,7 @@ extern crate serde_derive;
 pub mod plugins;
 
 use daggy::petgraph::visit::{IntoNodeReferences, NodeRef};
-use daggy::{Dag, Walker};
+use daggy::{Dag, EdgeIndex, Walker};
 use failure::{Error, Fallible};
 use serde::de::{self, Deserialize, Deserializer, MapAccess, Visitor};
 use serde::ser::{Serialize, SerializeStruct, Serializer};
@@ -188,17 +188,18 @@ impl Graph {
     /// Add a transition (edge) from `source` to `target`.
     ///
     /// Fails with the `WouldCycle` error if the new edge would lead to a cycle.
-    pub fn add_edge(&mut self, from: &ReleaseId, to: &ReleaseId) -> Result<(), Error> {
+    pub fn add_edge(&mut self, from: &ReleaseId, to: &ReleaseId) -> Result<EdgeIndex, Error> {
+        let from_release: String = self.find_by_releaseid(from)?.version().to_string();
+        let to_release: String = self.find_by_releaseid(to)?.version().to_string();
         if self.dag.find_edge(from.0, to.0).is_some() {
             return Err(Error::from(errors::EdgeAlreadyExists {
-                from: self.find_by_releaseid(from)?.version().to_string(),
-                to: self.find_by_releaseid(to)?.version().to_string(),
+                from: from_release,
+                to: to_release,
             }));
         }
 
         self.dag
             .add_edge(from.0, to.0, Empty {})
-            .map(|_| ())
             .map_err(Into::into)
     }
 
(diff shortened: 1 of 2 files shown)


cincinnati/src/lib.rs
```diff
@@ -187,18 +188,27 @@ impl Graph {
...
+                from: from_release,
+                to: to_release,
             }));
         }
 
         self.dag
             .add_edge(from.0, to.0, Empty {})
-            .map(|_| ())
+            .map(|edge| {
+                trace!(
```
steveej: I'm not a fan of adding log output to the core library. What do you think of returning the edge index from `add_edge` instead, and giving the caller the chance to log it?

vrutkovs: hmm, right, that would be cleaner

vrutkovs: Updated, PTAL

cincinnati/src/lib.rs
```diff
@@ -22,8 +22,9 @@ extern crate serde_derive;
 pub mod plugins;
 
 use daggy::petgraph::visit::{IntoNodeReferences, NodeRef};
-use daggy::{Dag, Walker};
+use daggy::{Dag, EdgeIndex, Walker};
 use failure::{Error, Fallible};
+pub use log::trace;
```
steveej: leftover

cincinnati/src/lib.rs
```diff
@@ -187,26 +188,31 @@ impl Graph {
...
-            .map(|_| ())
             .map_err(Into::into)
     }
 
     /// Add edges for all given key/value pairs of releases.
-    pub fn add_edges(&mut self, indices: HashMap<ReleaseId, ReleaseId>) -> Result<(), Error> {
+    pub fn add_edges(
+        &mut self,
+        indices: HashMap<ReleaseId, ReleaseId>,
+    ) -> Result<Vec<EdgeIndex>, Error> {
```
steveej: I'm not sure if this is useful. Iterating a HashMap isn't ordered, and so it's impossible for the caller to know which key/value pair resulted in which edge index. You could return the releaseids with the edgeindex if you think this is useful. E.g.

```suggestion
    ) -> Result<Vec<(EdgeIndex, ReleaseId, ReleaseId)>, Error> {
```

cincinnati/src/lib.rs
```diff
@@ -187,26 +188,31 @@ impl Graph {
...
         indices
             .iter()
-            .try_fold((), |_, (from, to)| self.add_edge(&from, &to))
+            .map(|(from, to)| self.add_edge(&from, &to))
```
steveej: Given the above comment this could be 

```suggestion
            .map(|(from, to)| (self.add_edge(&from, &to), from, to))
```

I think that the ordering matters here due to the evaluation precedence. Borrows of `to` and `from` are passed to `self.add_edge`, and then their ownership is passed along in the tuple. Besides this thought I didn't test this suggestion, but you'll have to reconcile with the compiler anyway :laughing:

vrutkovs: Its not gonna work:
```
a value of type `std::result::Result<std::vec::Vec<(petgraph::graph_impl::EdgeIndex, ReleaseId, ReleaseId)>, failure::Error>` cannot be built from an iterator over elements of type `(std::result::Result<petgraph::graph_impl::EdgeIndex, failure::Error>, ReleaseId, ReleaseId)`

value of type `std::result::Result<std::vec::Vec<(petgraph::graph_impl::EdgeIndex, ReleaseId, ReleaseId)>, failure::Error>` cannot be built from `std::iter::Iterator<Item=(std::result::Result<petgraph::graph_impl::EdgeIndex, failure::Error>, ReleaseId, ReleaseId)>
```

ReleaseId's need to be injected in Result<EdgeIndex> somehow

cincinnati/src/lib.rs
vrutkovs: iiuc this is not even used anywhere

steveej: If that's the case we can just leave this function alone for now.

cincinnati/src/lib.rs
steveej: Well, it's going to work with the map statement and the changed result

steveej: To elaborate: `collect()` without a given type will try to built a collection which it finds in whatever it's assigned to, in this case that's the return value for the function. So in this case the `map` statement has to return items which match the return type of the function.

wking:
/lgtm

Let's get this into stage so we can see if it helps debug.  We can always polish more later before promoting to prod.

PR #253, opened 2020-03-20
WIP cargo_test: rename artifacts dir

vrutkovs:
ci-operator has been updated and now uses ARTIFACTS instead of ARTIFACTS_DIR


vrutkovs:
/close

Not required

PR #251, opened 2020-03-20
Dockerfile.build: bump rust to 1.42.0

steveej:
Since our CI configuration is decoupled from this repository, it's not
possible to atomically merge a change which switches the rust version.
By leaving the oldest version installed, we can cleanly make the switch
without introducing a predictable failure state on CI.

---

Coupled with https://github.com/openshift/release/pull/7811

PR #255, opened 2020-03-20
dist/cargo_test: properly propagate test errors; plugins/edge_add_remove: sort edge-indices on removal

steveej:
This fixes various bugs. PTAL at the commit messages.

### DIFF
--- a/Justfile
+++ b/Justfile
@@ -43,7 +43,7 @@ run-ci-tests:
 	set -e
 	hack/run-all-tests.sh
 
-path_prefix = "api/upgrades_info/"
+path_prefix := "api/upgrades_info/"
 
 
 get-and-display-graph:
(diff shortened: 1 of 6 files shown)


wking:
/hold

Need to review after [the push][1].

[1]: https://github.com/openshift/cincinnati/pull/255#event-3150781526

wking:
Ah, just rebasing around #251 and #252

steveej:
> The following test **failed**, say `/retest` to rerun all failed tests:
> ci/prow/cargo-test 	[71a4fe6](https://github.com/openshift/cincinnati/commit/71a4fe664929aec17c343b6485497d92913255e8) 	[link](https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/pr-logs/pull/openshift_cincinnati/255/pull-ci-openshift-cincinnati-master-cargo-test/1090) 	`/test cargo-test`

With the test errors being effective, we're now seeing

```
thread 'slo::zz_check_slo::_cincinnati_gb_graph_upstream_errors_total_0_' panicked at 'PROM_ENDPOINT unset', graph-builder/tests/slo/mod.rs:77:14
```

which will be properly fixed in #249. In the meantime, I've added a feature gate to the SLO tests in 
44d3c4f .

steveej:
> I don't understand this syntax. Is this "don't worry about the SLO tests", somehow?

it prevents the [slo tests](https://github.com/openshift/cincinnati/tree/master/graph-builder/tests/slo) from being built and executed, except when `--feature test-e2e` is passed to cargo. We only pass this in the [e2e test](https://github.com/openshift/cincinnati/blob/44d3c4fbe083f8f49813d0f92dedba090a0d5df5/dist/Dockerfile.e2e/Dockerfile#L7), which is where we want to run the SLO tests.

The upstream documentation on the syntax is [here](https://doc.rust-lang.org/reference/conditional-compilation.html#the-cfg-attribute).

wking:
If CI passes, I'm happy:

/lgtm

Although...

> We only pass this in the e2e test...

Isn't e2e a required test for this PR?

Screenshot from 2020-03-20 14-18-02

Or is that a different e2e test from the one that gets `--feature test-e2e`?

steveej:
> Isn't e2e a required test for this PR?

Yes, but that one is overridable. Maybe the other tests are as well, but I'm not comfortable doing that, while I consider the e2e test semi-blocking.

wking:
/overide e2e

Only failure was [SLO][1], discussed above.

[1]: https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/pr-logs/pull/openshift_cincinnati/255/pull-ci-openshift-cincinnati-master-e2e/411

wking:
/override ci/prow/e2e

Maybe Prow expects the full path.

PR #256, opened 2020-03-20
cincinnati/src: Drop AbstractRelease

wking:
We've had `AbstractRelease` since at least 05c049fe13 (2018-07-13).  The motivation wasn't spelled out in Git history, and [this comment][1] doesn't add much.  Because abstract releases were pruned in the graph builder, there's no loss of functionality by moving to using only concrete releases (now just `Release`, since we no longer need to distinguish).  The "how to add an edge if we haven't ingested one endpoint yet?" issue is addressed by removing all edge management from the graph engine and instead handling it all in the `edge-add-remove` plugin (which operates on the metadata assembed by previous layers, including the metadata extracted from the release images themselves and secondary metadata extracted from Quay labels, GitHub repositories, etc.).

[1]: https://github.com/openshift/cincinnati/pull/188#discussion_r375799508

### DIFF
--- a/cincinnati/src/lib.rs
+++ b/cincinnati/src/lib.rs
@@ -41,48 +41,12 @@ pub struct Graph {
     dag: Dag<Release, Empty>,
 }
 
-/// Wrapper enum for the concrete and abstract release types.
-#[derive(Debug, Deserialize, Serialize, PartialEq, Clone)]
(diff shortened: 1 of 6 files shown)


LalatenduMohanty:
What is an abstract release?

wking:
The effort continues in #474.

PR #257, opened 2020-03-23
tests/e2e: gate SLO tests with `test-e2e-slo` feature

steveej:
CI currently isn't set up to run these tests properly and they fail by
default. An additional feature gate works around that.

### DIFF
--- a/graph-builder/tests/mod.rs
+++ b/graph-builder/tests/mod.rs
@@ -1,4 +1,4 @@
 #[cfg(feature = "test-e2e")]
 mod e2e;
-#[cfg(feature = "test-e2e")]
+#[cfg(feature = "test-e2e-slo")]
 mod slo;


LalatenduMohanty (approved):
/hold for @vrutkovs to approve

We can fix the tests related to SLO and when it is stable , include it in CI

/lgtm

vrutkovs:
This would effectively disable those, but I'm not sure what's the issue with those - they run (perhaps flake though).

Also #249 would rework SLO tests

steveej:
When #255 was merged, they did not run successfully as seen in https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/pr-logs/pull/openshift_cincinnati/255/pull-ci-openshift-cincinnati-master-e2e/411

steveej:
> We can fix the tests related to SLO and when it is stable , include it in CI

I agree with that. As discussed out of band we'll go forward with this.

/hold cancel

PR #258, opened 2020-03-23
Justfile: make running the graph-builder and e2e tests simpler

steveej:


Justfile
```diff
@@ -45,18 +45,46 @@ run-ci-tests:
 
 path_prefix := "api/upgrades_info/"
 
+display-graph:
+	jq -cM . | {{invocation_directory()}}/hack/graph.sh | dot -Tsvg > graph.svg; xdg-open graph.svg
```
LalatenduMohanty, line 59: We need to add the graph.svg to `.gitignore`.

LalatenduMohanty (changes requested):
IMO this should be a single commit.

Justfile
LalatenduMohanty: For some reason this is not working in my tmux session. But thats not a blocker for the PR.

steveej: Please paste the command and its output here. It's supposed to work for you as well.

steveej:
> IMO this should be a single commit.

I don't think so as they're unrelated changes. But if you feel strongly about it I'd rather squash than having this PR linger around :wink:

LalatenduMohanty:
> I don't think so as they're unrelated changes. But if you feel strongly about it I'd rather squash than having this PR linger around wink

I do not have a strong opinion.

Justfile
LalatenduMohanty: I guess I am using `$ just display-graph` in a wrong way. Can you add an example in the dev docs?

LalatenduMohanty (changes requested):

Justfile
LalatenduMohanty: Running `$ just get-and-display-graph` works fine. But as expected `$ just display-graph` would hang as it need `json` input. If this is right, we should make `$ just display-graph` a private recipe.

LalatenduMohanty: @steveeJ  mentioned that the right usage is `just get-graph-pe-production "stable-4.2" "amd64" | just display-graph`. As discussed we should add  docs for ` display-graph` in the Justfile for `just --list` to show. I will also suggest that we add `just --list` to the dev docs.

steveej: I'm not yet ready to document all of this. I would like to, but I'm not convinced this is a stable interface. Right now we're observing a divergence, or at best a duplication, between the Justfile and the scripts in _dist/_ and _hack/_. I would like to unify this as much as possible soon along with updating the documentation.

LalatenduMohanty: I think the divergence exists because we have not accepted the Justfile yet as the default way for local development. By making it easier to use the `Justfile` will be a positive step in that direction. I would suggest lets add the doc required for this PR in `$ just --list` and then take up the other things required to unify things.

steveej: Okay. So you're suggesting to simply add docs to the recipes in this PR?

LalatenduMohanty (approved):
/lgtm

PR #261, opened 2020-03-30
e2e/load testing: use production data

steveej:


steveej:
These changes won't work on their own because the e2e test still expects the data from the test repository.

steveej:
This was merged as part of #232.

PR #263, opened 2020-03-31
Adjust max workers in load testing scenarios and improve PE <-> GB connections

vrutkovs:
Set `max-workers` == `workers` so that PE wouldn't get a large stream of additional connections.

This also adds the following improvements for PE/GB:

* Use gzip compression on GB side and enable using compressed connection on PE side
* Add timeouts to PE -> GB requests
* Enable keep-alive on GB and PE side

TODO:

* [x] Add a config option to `cincinnati-graph-fetch` plugin to configure -> GB client timeout 

Ref: https://issues.redhat.com/browse/OTA-105 to improve load-testing results

### DIFF
--- a/cincinnati/src/plugins/internal/cincinnati_graph_fetch.rs
+++ b/cincinnati/src/plugins/internal/cincinnati_graph_fetch.rs
@@ -19,12 +19,18 @@ use std::time::Duration;
 /// Default URL to upstream graph provider.
 pub static DEFAULT_UPSTREAM_URL: &str = "http://localhost:8080/v1/graph";
 
+/// Default graph-builder connection timeout in seconds.
+pub static DEFAULT_TIMEOUT_SECS: u64 = 30;
+
 /// Plugin settings.
 #[derive(Clone, CustomDebug, Deserialize, SmartDefault)]
 #[serde(default)]
 struct CincinnatiGraphFetchSettings {
     #[default(DEFAULT_UPSTREAM_URL.to_string())]
     upstream: String,
+
+    #[default(DEFAULT_TIMEOUT_SECS)]
+    timeout: u64,
 }
 
 /// Graph fetcher for Cincinnati `/v1/graph` endpoints.
@@ -48,7 +54,7 @@ pub struct CincinnatiGraphFetchPlugin {
 impl PluginSettings for CincinnatiGraphFetchSettings {
     fn build_plugin(&self, registry: Option<&prometheus::Registry>) -> Fallible<BoxedPlugin> {
         let cfg = self.clone();
-        let plugin = CincinnatiGraphFetchPlugin::try_new(cfg.upstream, registry)?;
+        let plugin = CincinnatiGraphFetchPlugin::try_new(cfg.upstream, cfg.timeout, registry)?;
         Ok(new_plugin!(InternalPluginWrapper(plugin)))
     }
 }
@@ -68,6 +74,7 @@ impl CincinnatiGraphFetchPlugin {
 
     fn try_new(
         upstream: String,
+        timeout: u64,
         prometheus_registry: Option<&prometheus::Registry>,
     ) -> Fallible<Self> {
         let http_upstream_reqs = Counter::new(
@@ -87,7 +94,7 @@ impl CincinnatiGraphFetchPlugin {
 
         let client = reqwest::ClientBuilder::new()
             .gzip(true)
-            .timeout(Duration::from_secs(30))
+            .timeout(Duration::from_secs(timeout))
             .build()
             .context("Building reqwest client")?;
 
@@ -173,7 +180,9 @@ mod tests {
                     .with_body($mock_body.to_string())
                     .create();
 
-                let plugin = CincinnatiGraphFetchPlugin::try_new(mockito::server_url(), None)?;
+                let timeout: u64 = 30;
+                let plugin =
+                    CincinnatiGraphFetchPlugin::try_new(mockito::server_url(), timeout, None)?;
                 let http_upstream_reqs = plugin.http_upstream_reqs.clone();
                 let http_upstream_errors_total = plugin.http_upstream_errors_total.clone();
 
@@ -238,7 +247,7 @@ mod tests {
                     .with_body($mock_body.to_string())
                     .create();
 
-                let plugin = CincinnatiGraphFetchPlugin::try_new($upstream.to_string(), None)?;
+                let plugin = CincinnatiGraphFetchPlugin::try_new($upstream.to_string(), 30, None)?;
                 let http_upstream_reqs = plugin.http_upstream_reqs.clone();
                 let http_upstream_errors_total = plugin.http_upstream_errors_total.clone();
 
@@ -301,7 +310,10 @@ mod tests {
             metrics_prefix.clone(),
         ))?));
 
-        let _ = CincinnatiGraphFetchPlugin::try_new(mockito::server_url(), Some(registry))?;
+        let timeout: u64 = 30;
+
+        let _ =
+            CincinnatiGraphFetchPlugin::try_new(mockito::server_url(), timeout, Some(registry))?;
 
         let metrics_call = metrics::serve::<metrics::RegistryWrapper>(actix_web::web::Data::new(
             RegistryWrapper(registry),
--- a/cincinnati/src/plugins/internal/cincinnati_graph_fetch.rs
+++ b/cincinnati/src/plugins/internal/cincinnati_graph_fetch.rs
@@ -14,6 +14,7 @@ use failure::{Fallible, ResultExt};
 use prometheus::Counter;
 use reqwest;
 use reqwest::header::{HeaderValue, ACCEPT};
+use std::time::Duration;
 
 /// Default URL to upstream graph provider.
 pub static DEFAULT_UPSTREAM_URL: &str = "http://localhost:8080/v1/graph";
@@ -85,6 +86,8 @@ impl CincinnatiGraphFetchPlugin {
         };
 
         let client = reqwest::ClientBuilder::new()
+            .gzip(true)
+            .timeout(Duration::from_secs(30))
             .build()
             .context("Building reqwest client")?;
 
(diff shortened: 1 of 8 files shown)


e2e/tests/slo.rs
```diff
@@ -9,11 +9,11 @@ use test_case::test_case;
 // No upstream errors
 #[test_case("max_over_time(cincinnati_pe_http_upstream_errors_total[1h])", "0")]
 #[test_case("max_over_time(cincinnati_gb_graph_upstream_errors_total[1h])", "0")]
-// Use clamp_min to bring up the minimal serve duration to 0.1 seconds
+// Use clamp_min to bring up the minimal serve duration to 500ms seconds
```
steveej:
```suggestion
// Use clamp_min to bring up the minimal serve duration to 500ms
```

graph-builder/src/main.rs
```diff
@@ -98,12 +98,14 @@ fn main() -> Result<(), Error> {
...
     HttpServer::new(move || {
         App::new()
+            .wrap(middleware::Compress::default())
             .app_data(actix_web::web::Data::new(main_state.clone()))
             .service(
                 actix_web::web::resource(&format!("{}/v1/graph", app_prefix.clone()))
                     .route(actix_web::web::get().to(graph::index)),
             )
     })
+    .keep_alive(10)
```
steveej, line 108: What does this do?

policy-engine.log
```diff
@@ -0,0 +1 @@
+   Compiling cincinnati v0.1.0 (/var/home/vrutkovs/src/github.com/openshift/cincinnati/cincinnati)
```
steveej: this file is probably added by accident

policy-engine/src/main.rs
```diff
@@ -103,6 +104,7 @@ fn main() -> Result<(), Error> {
                     .route(actix_web::web::get().to(openapi::index)),
             )
     })
+    .keep_alive(10)
```
steveej, line 107: What does this do?

cincinnati/src/plugins/internal/cincinnati_graph_fetch.rs
```diff
@@ -85,6 +86,8 @@ impl CincinnatiGraphFetchPlugin {
         };
 
         let client = reqwest::ClientBuilder::new()
+            .gzip(true)
+            .timeout(Duration::from_secs(30))
```
steveej: This should be configurable on the plugin level

graph-builder/src/main.rs
vrutkovs: This instructs graph-builder not close the client connections immediately but reuse it for other requests from PE.

policy-engine/src/main.rs
vrutkovs: Same but for policy-engine: if the client is hitting PE with requests it should reuse connections instead of closing / creating new ones. The latter is an expensive operation and might help PE to cope with malicious client (== load testing scripts)

cincinnati/src/plugins/internal/cincinnati_graph_fetch.rs
vrutkovs: Good idea

policy-engine.log
vrutkovs: Ooops, removed

cincinnati/src/plugins/internal/cincinnati_graph_fetch.rs
vrutkovs: Implemented this in f0411e877f65b9c3eec473d714c8ec5d2c3cd9df although I'm not sure if that is sufficient

cincinnati/src/plugins/internal/cincinnati_graph_fetch.rs
```diff
@@ -263,6 +278,7 @@ mod tests {
     fetch_upstream_failure_test!(
         name: fetch_fail_invalid_url,
         upstream: "invalid.url",
+        gb_timeout: 30,
```
steveej: looks like all occurrences of the `fetch_upstream_failure_test!` macro use the same timeout value. According to the DRY principle it can therefore be hardcoded in the macro.

cincinnati/src/plugins/internal/cincinnati_graph_fetch.rs
```diff
@@ -33,6 +39,9 @@ pub struct CincinnatiGraphFetchPlugin {
     /// The upstream from which to fetch the graph
     pub upstream: String,
 
+    /// graph-builder connection timeout in seconds.
+    pub gb_timeout: u64,
```
steveej: We try to use rich types everywhere, so this should be a `Duration`

cincinnati/src/plugins/internal/cincinnati_graph_fetch.rs
```diff
@@ -19,12 +19,18 @@ use std::time::Duration;
 /// Default URL to upstream graph provider.
 pub static DEFAULT_UPSTREAM_URL: &str = "http://localhost:8080/v1/graph";
 
+/// Default graph-builder connection timeout in seconds.
+pub static DEFAULT_GB_TIMEOUT: u64 = 30;
```
steveej:
```suggestion
pub static DEFAULT_GB_TIMEOUT_SECS: u64 = 30;
```

cincinnati/src/plugins/internal/cincinnati_graph_fetch.rs
vrutkovs: hmm, in fact we don't need it here - timeout is set once during `Client` build phase in `try_new`

cincinnati/src/plugins/internal/cincinnati_graph_fetch.rs
vrutkovs: Randomized values for `gb_timeout` there. These are not being verified yet, but it may trigger an error so its better than hardcoding.

Does that sound acceptable?

cincinnati/src/plugins/internal/cincinnati_graph_fetch.rs
```diff
@@ -14,16 +14,23 @@ use failure::{Fallible, ResultExt};
...
+
 /// Plugin settings.
 #[derive(Clone, CustomDebug, Deserialize, SmartDefault)]
 #[serde(default)]
 struct CincinnatiGraphFetchSettings {
     #[default(DEFAULT_UPSTREAM_URL.to_string())]
     upstream: String,
+
+    #[default(DEFAULT_GB_TIMEOUT_SECS)]
+    gb_timeout: u64,
```
steveej: Sorry for the late nitpick, but the `gb_` prefix is not appropriate. The plugin fetches a graph, it could do so from any graph endpoint. Let's just call it timeout?

```suggestion
    timeout: u64,
```

cincinnati/src/plugins/internal/cincinnati_graph_fetch.rs
```diff
@@ -260,6 +274,7 @@ mod tests {
     fetch_upstream_failure_test!(
         name: fetch_fail_invalid_url,
         upstream: "invalid.url",
+        gb_timeout: 10,
```
steveej: I'm still not convinced even though the numbers are randomized now ;-) If we don't use them, let's skip them for now please. I would prefer to actually test the timeout of course!

graph-builder/src/main.rs
steveej: I trust you to have included this in the commit message ;-)

policy-engine/src/main.rs
steveej: I trust you to have included this in the commit message ;-)

cincinnati/src/plugins/internal/cincinnati_graph_fetch.rs
vrutkovs: Yeah, sure

cincinnati/src/plugins/internal/cincinnati_graph_fetch.rs
steveej: That should so!

cincinnati/src/plugins/internal/cincinnati_graph_fetch.rs
steveej: Even better ;-)

cincinnati/src/plugins/internal/cincinnati_graph_fetch.rs
steveej: I don't think so as I find it too misleading.

vrutkovs: Hardcoded this in the macros so that won't lead us into thinking this is actually properly tested

steveej: I think we still have a misunderstanding happening here. What I mean by hardcoding in the macro (note the singular) is to use a fixed number in the macro itself, instead of exposing a macro argument for it. From me macro's point of view this is still dynamically passed in.

steveej (approved):
/lgtm

One last moment for you to think about implementing a test for the timeout case, which I encourage you to do. Especially interesting is the error which the caller in the policy-engine is faced.
/hold

vrutkovs:
I suppose more Cincinnati areas would needs timeout tests, so lets fix it as a whole - filed  OTA-175 to track timeout tests

/retest

steveej:
> I suppose more Cincinnati areas would needs timeout tests, so lets fix it as a whole

Even though I'm fine postponing, this explanation doesn't make much sense to me :wink:

steveej:
Just to clarify: I wanted you to give a chance to think about testing again now. Please feel free to remove the hold.

policy-engine/src/main.rs
wking: > ... if the client is hitting PE with requests it should reuse connections instead of closing / creating new ones.

But we expect the cluster-version operator to poll the policy engine every 2 to 4 minutes.  Do we expect the CVO to hold those open for that long?  Creating new connections doesn't seem like it would be _that_ expensive :p.

vrutkovs: On PE's side this measure is against malicious clients, which create a lot of connections and close them instantly.

CVO requests probably won't be affected by this - our average payload is ~15KB, so it should be almost instant on any connection.

PR #264, opened 2020-03-31
dist/openshift/cincinnati: Drop graph-builder pause to 5 minutes

wking:
The default seems to have lived in the 30s range since at least 517af6ee8f (#69).  But we can allow some latency here.  On the cluster-version operator (CVO) side:

* `cvo.New` [takes an argument][1] for the poll interval.
* That is fed by [`resyncPeriod(o.ResyncInterval)()`][2].
* `ResyncInterval` is [set from `minResyncPeriod`][3].
* `minResyncPeriod` is [2 minutes][4].
* `resyncPeriod` adds some jitter (returning [between 1x and 2x the input value][5]).

So should be uniformly distributed between 2 and 4 minutes.  That puts a floor on latency in getting this information out to clusters.

It's also not a problem to add latency above the CVO-enforced floor.  A delay in minutes in getting a new release out into a channel is not significant, because the biggest publication rush would be releasing a CVE fix into the stable channel.  And that's already delayed by needing to cook in the candidate and fast channels, and eventually having the push to stable being a phased rollout.

A delay in pulling an edge that has proven itself unstable is more serious, but we already have latencies in the minutes before alerts fire, and Telemetry/Insights that report issues to Red Hat add additional latencies in the minutes (Telemetry) to hours (Insights) range.  Delaying the edge pull by a few more minutes is not going to greatly increase overall cluster-breaks -> edge-pulled latency here.

[1]: https://github.com/openshift/cluster-version-operator/blob/12623332615a0657a5468fae27cff1998a70bfef/pkg/cvo/cvo.go#L165
[2]: https://github.com/openshift/cluster-version-operator/blob/12623332615a0657a5468fae27cff1998a70bfef/pkg/start/start.go#L341
[3]: https://github.com/openshift/cluster-version-operator/blob/12623332615a0657a5468fae27cff1998a70bfef/pkg/start/start.go#L93
[4]: https://github.com/openshift/cluster-version-operator/blob/12623332615a0657a5468fae27cff1998a70bfef/pkg/start/start.go#L43
[5]: https://github.com/openshift/cluster-version-operator/blob/12623332615a0657a5468fae27cff1998a70bfef/pkg/start/start.go#L250-L251

### DIFF
--- a/dist/openshift/cincinnati.yaml
+++ b/dist/openshift/cincinnati.yaml
@@ -266,7 +266,7 @@ parameters:
     displayName: "Policy-engine CPU request"
     description: "Requested amount of CPU (millicores) allowed for policy-engine (default: 350m)"
   - name: GB_PAUSE_SECS
-    value: "30"
+    value: "300"
     displayName: Seconds to pause between scrapes
   - name: GB_PORT
     value: "8080"
--- a/docs/user/graph-builder-configuration.md
+++ b/docs/user/graph-builder-configuration.md
@@ -20,6 +20,6 @@ TOML configuration currently supports the following sections and options:
    - `registry` (section): configuration for Docker-v2 registry provider.
      - `credentials_path` (string): path to file containing registry credentials, in "dockercfg" format. Default: unset.
      - `manifestref_key` (string): metadata key where to record the manifest-reference. Default: "io.openshift.upgrades.graph.release.manifestref".
-     - `pause_secs` (unsigned integer): pause between repository scrapes, in seconds. Default: 30.
+     - `pause_secs` (unsigned integer): pause between repository scrapes, in seconds. Default: 300.
      - `repository` (string): target image in the registry. Default: "openshift".
      - `url` (string): URL for the registry. Default: "http://localhost:5000". 
--- a/graph-builder/src/config/settings.rs
+++ b/graph-builder/src/config/settings.rs
@@ -32,7 +32,7 @@ pub struct AppSettings {
     pub path_prefix: String,
 
     /// Pause (in seconds) between registry scrapes.
-    #[default(time::Duration::from_secs(30))]
+    #[default(time::Duration::from_secs(300))]
     pub pause_secs: time::Duration,
 
     /// Listening port for the main service.


wking:
The above discussion hopefully shows that there is little downside to growing the latency up to 10 minutes (and with this commit, the delay is 10 minutes plus the time it takes the plugins to execute).  The upside is that it decreases our load on our external dependencies (quay.io, GitHub), reducing issues like:

```
[2020-03-31T21:35:16Z DEBUG graph_builder::graph] graph update triggered
[2020-03-31T21:35:20Z ERROR graph_builder::graph] Checking for new commit
[2020-03-31T21:35:20Z ERROR graph_builder::graph] Parsing {"message":"API rate limit exceeded for user ID ....","documentation_url":"https://developer.github.com/v3/#rate-limiting"} to Vec<Branch>
[2020-03-31T21:35:20Z ERROR graph_builder::graph] invalid type: map, expected a sequence at line 1 column 0
[2020-03-31T21:35:50Z DEBUG graph_builder::graph] graph update triggered
```

It's also reduces our CPU and network load, since the graph-builder pod spends more time resting in between recalculate-the-graph spikes.  Just the load consideration is enough for me to want to tune this latency as high as we can while keeping it well below other edge-pulling latency components.

wking:
[e2e][1] died on throttling, trying to set up the cluster, before we got around to running our tests.

/retest

[1]: https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/pr-logs/pull/openshift_cincinnati/264/pull-ci-openshift-cincinnati-master-e2e/505

graph-builder/src/config/file.rs
```diff
@@ -143,7 +143,7 @@ mod tests {
 
                 [upstream]
                 method = "registry"
-                pause_secs = 35
+                pause_secs = 600
```
steveej: There's no point in keeping this in sync with the default, it's just a parser test.

steveej (commented):
I don't have any objections to this but I'd like to gather consensus to see if others think it's fine to increase the default latency this much.

One nitpick in the comments, feel free to ignore it.

graph-builder/src/config/file.rs
wking: Dropped with f7c476f -> 000eacc.

steveej:
It seems that @wking has investigated the chain of events thoroughly to put this delay into perspective. However, I would like to discuss what happens if for some reason the scraping fails a few times in a row. With a delay as low as 30s this is not a concern, but with 10m it is. My hunch here is that we want a smarter algorithm rather than a higher default delay. Some sort of adaptive back-off algorithm with a defined minimum and maximum value would be ideal.

vrutkovs:
+1 to increase this. I'd drop this to 5 minutes though:

* we ensure an alert is thrown on 3 consecutive upstream scrape errors (==15 mins)

* we can triage the issue and fix it / restart the pod within 15 mins

so that we fit into max 30 mins of graph staleness. With 10 minutes delay we'd get an alert with no time to fix it without violating SLO

vrutkovs:
>We might not even need this if we use github-mirror 

This is about quay image scraper, not policy-engine's filter

wking:
> > We might not even need this if we use github-mirror
>
> This is about quay image scraper, not policy-engine's filter

My understanding is that currently both the quay.io poll and the GitHub poll happen once each successful graph-engine cycle.  Until we can decouple those polls ([OTA-165][1]; sorry external folks), I think we have this single knob to control the frequency of both.

> However, I would like to discuss what happens if for some reason the scraping fails a few times in a row.

I'm fine with delaying several hours in the rare event that Quay and/or GitHub have an extended outage.  But sure, we can revisit this once we have a better handle on the SLO.  Is there a value greater than the current 30s that we feel we can agree on in the meantime?  1m?  5m?

[1]: https://issues.redhat.com/browse/OTA-165

vrutkovs:
>My understanding is that currently both the quay.io poll and the GitHub poll happen once each successful graph-engine cycle.

Correct, github fetch however takes ~0.2 secs and being cached, while quay image 
checks has to happen every scrape (to ensure tags didn't change meanwhile).

I think 5m would be the compromise here, 1m won't show significant improvement yet

LalatenduMohanty:
Considering quay is the culprit where the scrape gets stuck, I am still think the right thing to do is to improve the code so that it never gets stuck irrespective of what is the frequency. 

I am not sure if reducing frequency will solve the issue. So I am not keen on reducing the frequency as it will impact our ability to respond to secondary metadata changes. I will try to get feedback on this from stakeholders. If you folks strongly feel we should reduce the frequency, I can accept anything less than 5 mins. I would suggest 2 mins, but 10 mins too long for me (till we get feedback from others that they don't care)

/hold cancel

dist/openshift/cincinnati.yaml
```diff
@@ -266,7 +266,7 @@ parameters:
     displayName: "Policy-engine CPU request"
     description: "Requested amount of CPU (millicores) allowed for policy-engine (default: 350m)"
   - name: GB_PAUSE_SECS
-    value: "30"
+    value: "600"
```
LalatenduMohanty: 10 minutes is a very big change from 30 seconds. I am fine with anything less than 5 mins as mentioned in https://github.com/openshift/cincinnati/pull/264#issuecomment-608335262

LalatenduMohanty (changes requested):

LalatenduMohanty:
> My hunch here is that we want a smarter algorithm rather than a higher default delay. Some sort of adaptive back-off algorithm with a defined minimum and maximum value would be ideal.

Sounds better than one hard coded value for frequency.

dist/openshift/cincinnati.yaml
wking: Shifted to 5m with 000eacc -> 3449492.

wking:
I have no problems with smarter code, but bumping the pause is a patch we have today, and smarter code doesn't exist yet ;).

wking:
> Considering quay is the culprit where the scrape gets stuck...

That is an orthogonal issue.  The proximal issue here is that we were consuming too many GitHub API tokens.  And the strategic issue is that a 30s pause is needlessly pounding away on our external dependencies in an attempt to achieve graph-update latencies orders of magnitude lower than anyone actually cares about ;).

vrutkovs:
/lgtm
/hold

for @steveeJ @LalatenduMohanty to confirm we're good with 5 minutes

wking:
I thought @LalatenduMohanty had confirmed 5m [here][1] and [here][2], although reading that again he is saying "anything less than 5 mins" so < not <=.  Not sure how formal he was being.  Pretend I chose 4.999999999999999999999998 minutes ;).  I'm also fine waiting for him to weigh in again on his own behalf ;).

[1]: https://github.com/openshift/cincinnati/pull/264#issuecomment-608335262
[2]: https://github.com/openshift/cincinnati/pull/264#discussion_r402880161

LalatenduMohanty:
Lets go with 5 mins , till we have a better solution. Also I am going to ask openshift release-admins/architects about what is the latency acceptable to them 
/hold cancel

PR #266, opened 2020-04-03
Introduce opentracing

vrutkovs:
This commit instruments graph-builder and policy-engine web servers and
plugins to send traces to Jaeger instance. Traces are useful to dig
into details of a particular requests and plugin performance.

TODO:
* [x] Rework plugin handling code to set span name to the active plugin
  https://github.com/vrutkovs/cincinnati/commit/b6da9786a581a3e161ad91a39ad2999386f97d5f throws stack overflow errors. Plugins now declare names in plugin code

* [x] Hide behind the feature flag / config option
  This would require Cincinnati rebuild or reconfiguration, so tracing can't be performed on a live instance probably.
  Requires `tracing_endpoint` option for GB/PE to enable provider. Use "127.0.0.1:6831" to test locally

pe-request

scrape

Ref: https://issues.redhat.com/browse/OTA-161

### DIFF
--- a/Justfile
+++ b/Justfile
@@ -8,6 +8,7 @@ pause_secs := "9999999"
 registry := "https://quay.io"
 repository := "openshift-release-dev/ocp-release"
 credentials_file := "${HOME}/.docker/config.json"
+default_tracing_endpoint := "localhost:6831"
 
 metadata_reference_e2e:
 	printf 'reference_revision = "%s"' "$(cat {{testdata_dir}}/{{metadata_revision_file}})"
@@ -144,6 +145,7 @@ run-graph-builder:
 		address = "127.0.0.1"
 		port = 8080
 		path_prefix = "{{path_prefix}}"
+		tracing_endpoint = "{{default_tracing_endpoint}}"
 
 		[status]
 		address = "127.0.0.1"
@@ -184,7 +186,7 @@ run-graph-builder-e2e:
 run-policy-engine:
 	#!/usr/bin/env bash
 	export RUST_BACKTRACE=1 RUST_LOG="policy_engine=trace,cincinnati=trace,actix=trace,actix_web=trace"
-	cargo run --package policy-engine -- -vvvv --service.address 0.0.0.0 --service.path_prefix {{path_prefix}} --upstream.cincinnati.url 'http://127.0.0.1:8080/{{path_prefix}}v1/graph' --service.mandatory_client_parameters='channel'
+	cargo run --package policy-engine -- -vvvv --service.address 0.0.0.0 --service.path_prefix {{path_prefix}} --upstream.cincinnati.url 'http://127.0.0.1:8080/{{path_prefix}}v1/graph' --service.mandatory_client_parameters='channel' --service.tracing_endpoint "{{default_tracing_endpoint}}"
 
 kill-daemons:
 	pkill graph-builder
--- a/cincinnati/src/plugins/internal/cincinnati_graph_fetch.rs
+++ b/cincinnati/src/plugins/internal/cincinnati_graph_fetch.rs
@@ -10,10 +10,13 @@ use self::cincinnati::plugins::prelude_plugin_impl::*;
 use self::cincinnati::CONTENT_TYPE;
 
 use commons::prelude_errors::*;
+use commons::tracing::{get_tracer, set_context};
+use opentelemetry::api::{Span, Tracer};
+
 use commons::GraphError;
 use prometheus::Counter;
 use reqwest;
-use reqwest::header::{HeaderValue, ACCEPT};
+use reqwest::header::{HeaderMap, HeaderValue, ACCEPT};
 use std::time::Duration;
 
 /// Default URL to upstream graph provider.
@@ -109,13 +112,23 @@ impl CincinnatiGraphFetchPlugin {
 
 impl CincinnatiGraphFetchPlugin {
     async fn do_run_internal(self: &Self, io: InternalIO) -> Fallible<InternalIO> {
+        // extract current trace ID from headers
+        // this is required to make graph-builder trace a child of police-engine request
+        let mut headers = HeaderMap::new();
+        headers.insert(ACCEPT, HeaderValue::from_static(CONTENT_TYPE));
+        {
+            let span = get_tracer().get_active_span();
+            set_context(span.get_context(), &mut headers)
+                .context("failed to set the tracing context")?;
+        }
+
         trace!("getting graph from upstream at {}", self.upstream);
         self.http_upstream_reqs.inc();
 
         let res = self
             .client
             .get(&self.upstream)
-            .header(ACCEPT, HeaderValue::from_static(CONTENT_TYPE))
+            .headers(headers)
             .send()
             .map_err(|e| GraphError::FailedUpstreamFetch(e.to_string()))
             .await?;
@@ -160,7 +173,6 @@ mod tests {
     use super::*;
     use cincinnati::testing::generate_custom_graph;
     use commons::metrics::{self, RegistryWrapper};
-    use commons::prelude_errors::*;
     use commons::testing::{self, init_runtime};
     use prometheus::Registry;
 
--- a/cincinnati/src/plugins/mod.rs
+++ b/cincinnati/src/plugins/mod.rs
@@ -15,9 +15,12 @@ use self::cincinnati::plugins::interface::{PluginError, PluginExchange};
 
 use async_trait::async_trait;
 pub use commons::prelude_errors::*;
+use commons::tracing::get_tracer;
 use std::collections::HashMap;
 use std::convert::{TryFrom, TryInto};
 use std::fmt::Debug;
+
+use opentelemetry::api::{trace::futures::Instrument, Tracer};
 
 pub mod prelude {
     use crate as cincinnati;
@@ -381,11 +384,14 @@ where
 {
     let mut io = initial_io;
 
+    let _ = get_tracer().start("plugins", None);
+
     for next_plugin in plugins {
         let plugin_name = next_plugin.get_name();
         log::trace!("Running next plugin '{}'", plugin_name);
 
-        io = next_plugin.run(io).await?;
+        let plugin_span = get_tracer().start(plugin_name, None);
+        io = next_plugin.run(io).instrument(plugin_span).await?;
     }
 
     io.try_into()
@@ -454,7 +460,6 @@ where
 #[cfg(test)]
 mod tests {
     use super::*;
-    use crate::plugins::Plugin;
     use crate::testing::generate_graph;
     use futures::lock::Mutex as FuturesMutex;
     use lazy_static::lazy_static;
--- a/commons/src/tracing.rs
+++ b/commons/src/tracing.rs
@@ -1,0 +1,109 @@
+//! Tracing service.
+
+use opentelemetry::api::{
+    Carrier, HttpTextFormat, Key, Provider, Span, SpanContext, TraceContextPropagator,
+};
+use opentelemetry::{global, sdk};
+use opentelemetry_jaeger::{Exporter, Process};
+use std::collections::HashMap;
+
+use actix_web::dev::ServiceRequest;
+use actix_web::http;
+use reqwest::header::{HeaderMap, HeaderName, HeaderValue};
+
+use crate::prelude_errors::*;
+
+/// init_tracer sets up Jaeger tracer
+pub fn init_tracer(name: &'static str, maybe_agent_endpoint: Option<String>) -> Fallible<()> {
+    // Skip provider config if agent endpoint is not set
+    let agent_endpoint = match maybe_agent_endpoint {
+        None => return Ok(()),
+        Some(s) => s,
+    };
+
+    let exporter = Exporter::builder()
+        .with_agent_endpoint(agent_endpoint)
+        .with_process(Process {
+            service_name: name.to_string(),
+            tags: vec![Key::new("exporter").string("jaeger")],
+        })
+        .init()?;
+
+    let provider = sdk::Provider::builder()
+        .with_simple_exporter(exporter)
+        .with_config(sdk::Config {
+            default_sampler: Box::new(sdk::Sampler::Always),
+            ..Default::default()
+        })
+        .build();
+    global::set_provider(provider);
+
+    Ok(())
+}
+
+/// get_tracer returns an instance of global tracer
+pub fn get_tracer() -> global::BoxedTracer {
+    global::trace_provider().get_tracer("")
+}
+
+struct HttpHeaderMapCarrier<'a>(&'a http::HeaderMap);
+impl<'a> Carrier for HttpHeaderMapCarrier<'a> {
+    fn get(&self, key: &'static str) -> Option<&str> {
+        self.0.get(key).and_then(|value| value.to_str().ok())
+    }
+
+    fn set(&mut self, _key: &'static str, _value: String) {
+        unimplemented!()
+    }
+}
+
+struct ClientHeaderMapCarrier(HashMap<String, String>);
+impl Carrier for ClientHeaderMapCarrier {
+    fn get(&self, key: &'static str) -> Option<&str> {
+        self.0.get(key).map(|s| s.as_str())
+    }
+
+    fn set(&mut self, key: &'static str, value: String) {
+        self.0.insert(key.to_string(), value);
+    }
+}
+
+/// Return the parent context for the request if specific headers found.
+pub fn get_context(req: &ServiceRequest) -> SpanContext {
+    let propagator = TraceContextPropagator::new();
+    propagator.extract(&HttpHeaderMapCarrier(&req.headers()))
+}
+
+/// Inject context data into headers
+pub fn set_context(context: SpanContext, headers: &mut HeaderMap) -> crate::errors::Fallible<()> {
+    use std::str::FromStr;
+
+    let mut carrier = {
+        let headers_converted = headers.iter().try_fold(
+            HashMap::<String, String>::with_capacity(headers.len()),
+            |mut sum, (name, value)| -> crate::errors::Fallible<_> {
+                sum.insert(name.as_str().to_string(), value.to_str()?.to_string());
+                Ok(sum)
+            },
+        )?;
+
+        ClientHeaderMapCarrier(headers_converted)
+    };
+
+    let propagator = TraceContextPropagator::new();
+    propagator.inject(context, &mut carrier);
+
+    for (name, value) in carrier.0 {
+        headers.insert(HeaderName::from_str(&name)?, HeaderValue::from_str(&value)?);
+    }
+
+    Ok(())
+}
+
+/// Add span attributes from servicerequest
+pub fn set_span_tags(req: &ServiceRequest, span: &dyn Span) {
+    span.set_attribute(Key::new("path").string(req.path()));
+    req.headers().iter().for_each(|(k, v)| {
+        span.set_attribute(Key::new(format!("header.{}", k)).bytes(v.as_bytes().to_vec()))
+    });
+}
--- a/graph-builder/src/graph.rs
+++ b/graph-builder/src/graph.rs
@@ -18,8 +18,10 @@ use actix_web::{HttpRequest, HttpResponse};
 use cincinnati::plugins::prelude::*;
 use cincinnati::CONTENT_TYPE;
 use commons::metrics::HasRegistry;
+use commons::tracing::get_tracer;
 use commons::{Fallible, GraphError};
 use lazy_static;
+use opentelemetry::api::Tracer;
 pub use parking_lot::RwLock;
 use prometheus::{self, histogram_opts, labels, opts, Counter, Gauge, Histogram, IntGauge};
 use serde_json;
@@ -97,6 +99,8 @@ pub async fn index(
     req: HttpRequest,
     app_data: actix_web::web::Data<State>,
 ) -> Result<HttpResponse, GraphError> {
+    let _ = get_tracer().start("index", None);
+
     V1_GRAPH_INCOMING_REQS.inc();
 
     // Check that the client can accept JSON media type.
--- a/graph-builder/src/main.rs
+++ b/graph-builder/src/main.rs
@@ -12,11 +12,14 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use actix_service::Service;
 use actix_web::{middleware, App, HttpServer};
 use commons::metrics::{self, HasRegistry};
 use commons::prelude_errors::*;
+use commons::tracing::{get_context, get_tracer, init_tracer, set_span_tags};
 use graph_builder::{self, config, graph, status};
 use log::debug;
+use opentelemetry::api::{trace::futures::Instrument, Tracer};
 use parking_lot::RwLock;
 use std::collections::HashSet;
 use std::sync::Arc;
@@ -34,6 +37,9 @@ fn main() -> Result<(), Error> {
 
     let registry: prometheus::Registry =
         metrics::new_registry(Some(config::METRICS_PREFIX.to_string()))?;
+
+    // Enable tracing
+    init_tracer("graph-builder", settings.tracing_endpoint.clone())?;
 
     let plugins = settings.validate_and_build_plugins(Some(&registry))?;
 
@@ -99,6 +105,12 @@ fn main() -> Result<(), Error> {
     HttpServer::new(move || {
         App::new()
             .wrap(middleware::Compress::default())
+            .wrap_fn(|req, srv| {
+                let parent_context = get_context(&req);
+                let span = get_tracer().start("request", Some(parent_context));
+                set_span_tags(&req, &span);
+                srv.call(req).instrument(span)
+            })
             .app_data(actix_web::web::Data::new(main_state.clone()))
             .service(
                 actix_web::web::resource(&format!("{}/v1/graph", app_prefix.clone()))
--- a/policy-engine/src/graph.rs
+++ b/policy-engine/src/graph.rs
@@ -5,7 +5,9 @@ use actix_web::web::Query;
 use actix_web::{HttpRequest, HttpResponse};
 use cincinnati::plugins::BoxedPlugin;
 use cincinnati::CONTENT_TYPE;
+use commons::tracing::get_tracer;
 use commons::{self, Fallible, GraphError};
+use opentelemetry::api::{trace::futures::Instrument, Tracer};
 use prometheus::{histogram_opts, Counter, Histogram, Registry};
 use serde_json;
 use std::collections::HashMap;
@@ -38,6 +40,8 @@ pub(crate) async fn index(
     req: HttpRequest,
     app_data: actix_web::web::Data<AppState>,
 ) -> Result<HttpResponse, GraphError> {
+    let span = get_tracer().start("index", None);
+
     V1_GRAPH_INCOMING_REQS.inc();
 
     // Check that the client can accept JSON media type.
@@ -54,6 +58,7 @@ pub(crate) async fn index(
     let timer = V1_GRAPH_SERVE_HIST.start_timer();
 
     let response = process_plugins(app_data.plugins.iter(), plugin_params)
+        .instrument(span)
         .await
         .map_err(|e| {
             error!(
--- a/policy-engine/src/main.rs
+++ b/policy-engine/src/main.rs
@@ -23,10 +23,13 @@ mod config;
 mod graph;
 mod openapi;
 
+use actix_service::Service;
 use actix_web::{middleware, App, HttpServer};
 use cincinnati::plugins::BoxedPlugin;
 use commons::metrics::{self, RegistryWrapper};
 use commons::prelude_errors::*;
+use commons::tracing::{get_tracer, init_tracer, set_span_tags};
+use opentelemetry::api::{trace::futures::Instrument, Tracer};
 use prometheus::{labels, opts, Counter, Registry};
 use std::collections::HashSet;
 
@@ -81,6 +84,9 @@ fn main() -> Result<(), Error> {
     .bind((settings.status_address, settings.status_port))?
     .run();
 
+    // Enable tracing
+    init_tracer("policy-engine", settings.tracing_endpoint.clone())?;
+
     // Main service.
     let plugins = settings.validate_and_build_plugins(Some(registry))?;
     let state = AppState {
@@ -92,6 +98,11 @@ fn main() -> Result<(), Error> {
     HttpServer::new(move || {
         let app_prefix = state.path_prefix.clone();
         App::new()
+            .wrap_fn(|req, srv| {
+                let span = get_tracer().start("request", None);
+                set_span_tags(&req, &span);
+                srv.call(req).instrument(span)
+            })
             .app_data(actix_web::web::Data::<AppState>::new(state.clone()))
             .service(
                 actix_web::web::resource(&format!("{}/v1/graph", app_prefix))
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -12,7 +12,7 @@ dependencies = [
  "bytes 0.5.3 (registry+https://github.com/rust-lang/crates.io-index)",
  "crossbeam-channel 0.4.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "derive_more 0.99.2 (registry+https://github.com/rust-lang/crates.io-index)",
- "futures 0.3.1 (registry+https://github.com/rust-lang/crates.io-index)",
+ "futures 0.3.4 (registry+https://github.com/rust-lang/crates.io-index)",
  "lazy_static 1.4.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "log 0.4.8 (registry+https://github.com/rust-lang/crates.io-index)",
  "parking_lot 0.10.0 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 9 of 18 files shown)


cincinnati/src/lib.rs
```diff
@@ -21,6 +21,9 @@ extern crate serde_derive;
 #[macro_use]
 pub mod plugins;
 
+#[macro_use]
+extern crate trackable;
```
vrutkovs: This throws a warning here, since crate macros are being used later but not directly here. Is there a better way to import it?

steveej: You should be able to place `use trackable::macro_name` in the scope where you need the macro.

steveej (commented):
The code seems quite repetitive and invasive to me. I think that this functionality should be gated with a feature flag e.g. `tracing`, and I'd like to experiment with macros to try reduce the boilerplate. My concern is code readability and maintainability. For example, I think we can get around changing the plugin traits here.

vrutkovs:
Updated a todo, so holding this
/hold

I'm not entirely sure how to properly declutter it - seems macros could help (extracting trace ID from the headers is a good first candidate), perhaps moving it to a separate crate? My biggest concern is additional parameters to plugins though - not sure how to approach that problem - smth like `Traceable` trait?

steveej:
@vrutkovs I'm aware that the TODO's are a big change here and we discussed out-of-band that you feel stuck with this. If you would welcome it, I can work out a solution in my fork based on your work so far.

vrutkovs:
>If you would welcome it, I can work out a solution in my fork based on your work so far.

Sure thing, note that I got spare cycles to work on it myself - just not sure where to start

commons/src/tracing.rs
```diff
@@ -0,0 +1,85 @@
...
+
+    fn set(&mut self, key: &'static str, value: String) {
+        self.0.insert(
+            HeaderName::from_bytes(key.as_bytes()).unwrap(),
+            HeaderValue::from_str(&value).unwrap(),
+        );
+    }
+}
+
+/// get_context returns parent context for the request if specific headers found
```
steveej: nit: 

```suggestion
/// Return the parent context for the request if specific headers found.
```

commons/src/tracing.rs
```diff
@@ -0,0 +1,85 @@
...
+pub fn get_context(req: &ServiceRequest) -> SpanContext {
+    let propagator = TraceContextPropagator::new();
+    propagator.extract(&HttpHeaderMapCarrier(&req.headers()))
+}
+
+/// set_context inject context data into headers
```
steveej:
```suggestion
/// Inject a given context data into given headers.
```

commons/src/tracing.rs
```diff
@@ -0,0 +1,85 @@
...
+pub fn set_context(context: SpanContext, headers: &mut HeaderMap) {
+    let propagator = TraceContextPropagator::new();
+    propagator.inject(context, &mut ClientHeaderMapCarrier(headers));
+}
+
+/// set_span_tags adds span attributes from servicerequest
```
steveej:
```suggestion
/// Add span attributes from a given `ServiceRequest`.
```

graph-builder/src/graph.rs
```diff
@@ -28,6 +27,9 @@ use std::collections::HashSet;
 use std::sync::Arc;
 use std::thread;
 
+use commons::tracing::get_tracer;
+use opentelemetry::api::Tracer;
```
steveej: I think these both can go up with the other uses.

graph-builder/src/graph.rs
```diff
@@ -98,6 +100,8 @@ pub async fn index(
     req: HttpRequest,
     app_data: actix_web::web::Data<State>,
 ) -> Result<HttpResponse, GraphError> {
+    let _span = get_tracer().start("index", None);
```
steveej: Would

```suggestion
    let _ = get_tracer().start("index", None);
```

work as well?

graph-builder/src/main.rs
```diff
@@ -35,6 +39,9 @@ fn main() -> Result<(), Error> {
     let registry: prometheus::Registry =
         metrics::new_registry(Some(config::METRICS_PREFIX.to_string()))?;
 
+    // Enable tracing
+    init_tracer("graph-builder")?;
```
steveej: Should this be configurable?

policy-engine/src/graph.rs
```diff
@@ -11,6 +11,9 @@ use prometheus::{histogram_opts, Counter, Histogram, Registry};
 use serde_json;
 use std::collections::HashMap;
 
+use commons::tracing::get_tracer;
+use opentelemetry::api::{trace::futures::Instrument, Tracer};
```
steveej: These can go up

policy-engine/src/graph.rs
```diff
@@ -39,6 +42,8 @@ pub(crate) async fn index(
     req: HttpRequest,
     app_data: actix_web::web::Data<AppState>,
 ) -> Result<HttpResponse, GraphError> {
+    let _span = get_tracer().start("index", None);
```
steveej: also `_` maybe here? See above

policy-engine/src/main.rs
```diff
@@ -25,13 +25,17 @@ mod config;
...
+use actix_service::Service;
 use actix_web::{middleware, App, HttpServer};
 use cincinnati::plugins::BoxedPlugin;
 use commons::metrics::{self, RegistryWrapper};
+use commons::tracing::{get_tracer, init_tracer, set_span_tags};
 use failure::Error;
 use prometheus::{labels, opts, Counter, Registry};
 use std::collections::HashSet;
 
+use opentelemetry::api::{trace::futures::Instrument, Tracer};
```
steveej: Can go up

policy-engine/src/main.rs
```diff
@@ -83,6 +87,9 @@ fn main() -> Result<(), Error> {
     .bind((settings.status_address, settings.status_port))?
     .run();
 
+    // Enable tracing
+    init_tracer("policy-engine")?;
```
steveej: Make this configurable?

policy-engine/src/openapi.rs
```diff
@@ -1,6 +1,7 @@
 use crate::AppState;
 use actix_web::HttpResponse;
 use openapiv3::{OpenAPI, ReferenceOr};
+
```
steveej:
```suggestion
```

commons/src/tracing.rs
```diff
@@ -0,0 +1,32 @@
...
+        })
+        .build();
+    global::set_provider(provider);
+
+    Ok(())
+}
+
+/// get_tracer returns an instance of global tracer
+pub fn get_tracer() -> global::BoxedTracer {
+    global::trace_provider().get_tracer("")
```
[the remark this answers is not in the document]
vrutkovs: >it wraps just a simple, single line of code, and it leaks the external library's type

it does wrap a single line of code, but it gets used often. Should I create a macros instead?

As for the type I'm not too much concerned really - we still have to import external type's traits and can't fully abstract from a particular library anyway

>By using the same name (the empty string) for all callers, this will return the same tracer for all calls

It returns the default tracer, which is the one created with "init_tracer". It would be distinct in graph-builder and policy-engine.

commons/src/tracing.rs
vrutkovs: Ah, I assumed Rust requires comments to start with fn name - just like golang does :)

graph-builder/src/main.rs
vrutkovs: Its done once per binary, I think its safe to hardcode the name here

policy-engine/src/graph.rs
vrutkovs: We need a name here so that async call on line 58 could be instrumented

steveej: Oh, in that case we can remove the `_` prefix since it's actually used.

```suggestion
    let span = get_tracer().start("index", None);
```

policy-engine/src/main.rs
```diff
@@ -94,6 +100,11 @@ fn main() -> Result<(), Error> {
     HttpServer::new(move || {
         let app_prefix = state.path_prefix.clone();
         App::new()
+            .wrap_fn(|req, srv| {
+                let _span = get_tracer().start("request", None);
```
steveej:
```suggestion
                let span = get_tracer().start("request", None);
```

cincinnati/src/plugins/mod.rs
```diff
@@ -358,7 +361,8 @@ where
     let mut io = initial_io;
 
     for next_plugin in plugins {
-        io = next_plugin.run(io).await?;
+        let _span = get_tracer().start("plugin", None);
```
steveej:
```suggestion
        let span = get_tracer().start("plugin", None);
```

cincinnati/src/plugins/mod.rs
```diff
@@ -398,8 +405,12 @@ where
         let tx = tx.clone();
 
         std::thread::spawn(move || {
-            let io_future =
-                async { tokio::time::timeout(timeout, process(plugins, initial_io)).await };
+            let _span = get_tracer().start("plugins", None);
```
steveej:
```suggestion
            let span = get_tracer().start("plugins", None);
```

graph-builder/src/main.rs
```diff
@@ -99,6 +105,12 @@ fn main() -> Result<(), Error> {
     HttpServer::new(move || {
         App::new()
             .wrap(middleware::Compress::default())
+            .wrap_fn(|req, srv| {
+                let parent_context = get_context(&req);
+                let _span = get_tracer().start("request", Some(parent_context));
```
steveej:
```suggestion
                let span = get_tracer().start("request", Some(parent_context));
```

graph-builder/src/main.rs
steveej: I meant the initialization of the tracer in general. Related to this is this question: how is the endpoint configured to which the traces are sent?

commons/src/tracing.rs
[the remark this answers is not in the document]
steveej: > it does wrap a single line of code, but it gets used often. Should I create a macros instead?

I see, I think it's fine. Optimization advocates would maybe suggest to put a `#[inline(always)]` on top of this function, but I trust the compiler to make this decision.



> As for the type I'm not too much concerned really - we still have to import external type's traits and can't fully abstract from a particular library anyway

I think we can, but I understand you're implying that we don't want to.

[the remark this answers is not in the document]
vrutkovs: I don't think abstracting too much is a good idea. We already get all the features we want now from currently selected libraries, so I don't anticipate a switch to a different set of libs anytime soon

graph-builder/src/main.rs
vrutkovs: Current implementation is for developers only, thus hardcoded endpoint in the init_tracer. I'm planning to update this later on, once Jaeger gets deployed on app-sre side and we sort out the connection details.

Not sure if I should expose configuration early - it might have to be reworked later anyway

cincinnati/src/plugins/mod.rs
```diff
@@ -358,7 +361,8 @@ where
     let mut io = initial_io;
 
     for next_plugin in plugins {
-        io = next_plugin.run(io).await?;
+        let span = get_tracer().start("plugin", None);
+        io = next_plugin.run(io).instrument(_span).await?;
```
steveej: Sorry, I had assumed you'd replace the other occurrences :smile: 

```suggestion
        io = next_plugin.run(io).instrument(span).await?;
```

cincinnati/src/plugins/mod.rs
```diff
@@ -387,7 +391,10 @@ where
     let mut runtime = tokio::runtime::Runtime::new()?;
 
     let timeout = match timeout {
-        None => return runtime.block_on(process(plugins, initial_io)),
+        None => {
+            let _span = get_tracer().start("plugins", None);
```
steveej: Sorry, I had assumed you'd replace the other occurrences :smile: 

```suggestion
            let span = get_tracer().start("plugins", None);
```

cincinnati/src/plugins/mod.rs
```diff
@@ -387,7 +391,10 @@ where
...
+            return runtime.block_on(process(plugins, initial_io).instrument(_span));
```
steveej:
```suggestion
            return runtime.block_on(process(plugins, initial_io).instrument(span));
```

cincinnati/src/plugins/mod.rs
```diff
@@ -398,8 +405,12 @@ where
         let tx = tx.clone();
 
         std::thread::spawn(move || {
-            let io_future =
-                async { tokio::time::timeout(timeout, process(plugins, initial_io)).await };
+            let span = get_tracer().start("plugins", None);
+            let io_future = async {
+                tokio::time::timeout(timeout, process(plugins, initial_io))
+                    .instrument(_span)
```
steveej: Sorry, I had assumed you'd replace the other occurrences :smile: 

```suggestion
                    .instrument(span)
```

graph-builder/src/main.rs
```diff
@@ -99,6 +105,12 @@ fn main() -> Result<(), Error> {
     HttpServer::new(move || {
         App::new()
             .wrap(middleware::Compress::default())
+            .wrap_fn(|req, srv| {
+                let parent_context = get_context(&req);
+                let span = get_tracer().start("request", Some(parent_context));
+                set_span_tags(&req, &_span);
```
steveej:
```suggestion
                set_span_tags(&req, &span);
```

graph-builder/src/main.rs
```diff
@@ -99,6 +105,12 @@ fn main() -> Result<(), Error> {
...
+                srv.call(req).instrument(_span)
```
steveej:
```suggestion
                srv.call(req).instrument(span)
```

graph-builder/src/main.rs
steveej: What happens if the endpoint where traces are sent isn't available? Or is it pull-based like Prometheus?

policy-engine/src/main.rs
```diff
@@ -94,6 +100,11 @@ fn main() -> Result<(), Error> {
     HttpServer::new(move || {
         let app_prefix = state.path_prefix.clone();
         App::new()
+            .wrap_fn(|req, srv| {
+                let span = get_tracer().start("request", None);
+                set_span_tags(&req, &_span);
+                srv.call(req).instrument(_span)
```
steveej:
```suggestion
                set_span_tags(&req, &span);
                srv.call(req).instrument(span)
```

graph-builder/src/main.rs
vrutkovs: Its fire-and-forget. If the collector doesn't respond then no error thrown and its completely invisible to the app.

Not sure if I like that when its deployed on prod, but probably it can be tweaked later

vrutkovs:
/retest

cluster could not be installed

graph-builder/src/main.rs
steveej: Please let's take the time and tweak it before we land it. Is there a way it can verify that the controller is reachable?
If so, I think that `init_tracer` should throw an error if it's not.

Either way we should only call `init_tracer` if tracing has been configured. I think it makes sense to configure the controller socket as well. WDYT?

policy-engine/src/main.rs
steveej: Referring to the equivalent thread in the graph-builder: https://github.com/openshift/cincinnati/pull/266/files#r417288000

graph-builder/src/main.rs
vrutkovs: > Either way we should only call init_tracer if tracing has been configured

So you're proposing a flag to enable tracing and require Jaeger? 
That doesn't align well with the idea of open tracing - you should be able to deploy Jaeger and get trace data without Cincinnati reconfiguration (and pod redeploy).

I don't mind exposing a better config option for tracing later, once our production would have Jaeger deployed and default setting won't suit us (or customers).

steveej: >  That doesn't align well with the idea of open tracing - you should be able to deploy Jaeger and get trace data without Cincinnati reconfiguration (and pod redeploy).

It does come with some unnecessary load of sending the packets and having the kernel discard them since there's no one listening. I wasn't aware that it's meant to be a plug & play system and that it's designed this way. I'm not really fond of this design, but I can accept it.

cincinnati/src/plugins/internal/arch_filter.rs
```diff
@@ -87,6 +91,10 @@ lazy_static! {
 #[async_trait]
 impl InternalPlugin for ArchFilterPlugin {
     async fn run_internal(self: &Self, internal_io: InternalIO) -> Fallible<InternalIO> {
+        get_tracer()
+            .get_active_span()
+            .update_name("arch-filter".to_string());
```
steveej: Use the predefined constant instead:
```suggestion
            .update_name(Self::PLUGIN_NAME.to_string());
```

cincinnati/src/plugins/internal/channel_filter.rs
```diff
@@ -54,6 +57,9 @@ lazy_static! {
 #[async_trait]
 impl InternalPlugin for ChannelFilterPlugin {
     async fn run_internal(self: &Self, internal_io: InternalIO) -> Fallible<InternalIO> {
+        get_tracer()
+            .get_active_span()
+            .update_name("channel-filter".to_string());
```
steveej: Use the predefined constant instead:
```suggestion
            .update_name(Self::PLUGIN_NAME.to_string());
```

cincinnati/src/plugins/internal/cincinnati_graph_fetch.rs
```diff
@@ -143,6 +155,9 @@ impl CincinnatiGraphFetchPlugin {
 #[async_trait]
 impl InternalPlugin for CincinnatiGraphFetchPlugin {
     async fn run_internal(self: &Self, io: InternalIO) -> Fallible<InternalIO> {
+        get_tracer()
+            .get_active_span()
+            .update_name("graph_fetch".to_string());
```
steveej: Use the predefined constant instead:
```suggestion
            .update_name(Self::PLUGIN_NAME.to_string());
```

cincinnati/src/plugins/internal/graph_builder/github_openshift_secondary_metadata_scraper/plugin.rs
```diff
@@ -445,6 +448,10 @@ impl PluginSettings for GithubOpenshiftSecondaryMetadataScraperSettings {
 #[async_trait]
 impl InternalPlugin for GithubOpenshiftSecondaryMetadataScraperPlugin {
     async fn run_internal(self: &Self, mut io: InternalIO) -> Fallible<InternalIO> {
+        get_tracer()
+            .get_active_span()
+            .update_name("github-secondary-metadata".to_string());
```
steveej: Use the predefined constant instead:
```suggestion
            .update_name(Self::PLUGIN_NAME.to_string());
```

cincinnati/src/plugins/internal/edge_add_remove.rs
```diff
@@ -25,6 +27,10 @@ pub struct EdgeAddRemovePlugin {
 #[async_trait]
 impl InternalPlugin for EdgeAddRemovePlugin {
     async fn run_internal(self: &Self, io: InternalIO) -> Fallible<InternalIO> {
+        get_tracer()
+            .get_active_span()
+            .update_name("edge_add_remove".to_string());
```
steveej: Use the predefined constant instead:
```suggestion
            .update_name(Self::PLUGIN_NAME.to_string());
```

cincinnati/src/plugins/internal/graph_builder/openshift_secondary_metadata_parser/plugin.rs
```diff
@@ -472,6 +475,10 @@ impl OpenshiftSecondaryMetadataParserPlugin {
 #[async_trait]
 impl InternalPlugin for OpenshiftSecondaryMetadataParserPlugin {
     async fn run_internal(self: &Self, mut io: InternalIO) -> Fallible<InternalIO> {
+        get_tracer()
+            .get_active_span()
+            .update_name("parse-secondary-metadata".to_string());
```
steveej: Use the predefined constant instead:
```suggestion
            .update_name(Self::PLUGIN_NAME.to_string());
```

cincinnati/src/plugins/internal/graph_builder/release_scrape_dockerv2/plugin.rs
```diff
@@ -134,6 +137,10 @@ impl ReleaseScrapeDockerv2Plugin {
 #[async_trait]
 impl InternalPlugin for ReleaseScrapeDockerv2Plugin {
     async fn run_internal(self: &Self, io: InternalIO) -> Fallible<InternalIO> {
+        get_tracer()
+            .get_active_span()
+            .update_name("registry-scrape".to_string());
```
steveej: Use the predefined constant instead:
```suggestion
            .update_name(Self::PLUGIN_NAME.to_string());
```

cincinnati/src/plugins/internal/metadata_fetch_quay.rs
```diff
@@ -99,6 +102,10 @@ impl QuayMetadataFetchPlugin {
 #[async_trait]
 impl InternalPlugin for QuayMetadataFetchPlugin {
     async fn run_internal(self: &Self, io: InternalIO) -> Fallible<InternalIO> {
+        get_tracer()
+            .get_active_span()
+            .update_name("metadata-fetch-quay".to_string());
```
steveej: Use the predefined constant instead:
```suggestion
            .update_name(Self::PLUGIN_NAME.to_string());
```

cincinnati/src/plugins/internal/node_remove.rs
```diff
@@ -38,6 +40,10 @@ impl NodeRemovePlugin {
 #[async_trait]
 impl InternalPlugin for NodeRemovePlugin {
     async fn run_internal(self: &Self, io: InternalIO) -> Fallible<InternalIO> {
+        get_tracer()
+            .get_active_span()
+            .update_name("node-remove".to_string());
```
steveej: Use the predefined constant instead:
```suggestion
            .update_name(Self::PLUGIN_NAME.to_string());
```

cincinnati/src/plugins/mod.rs
```diff
@@ -387,7 +391,10 @@ where
     let mut runtime = tokio::runtime::Runtime::new()?;
 
     let timeout = match timeout {
-        None => return runtime.block_on(process(plugins, initial_io)),
+        None => {
+            let span = get_tracer().start("plugins", None);
```
steveej: How about we expose the plugin name so we can read it here, and start a span around the plugin? This would allow us to gather plugin statistics without requiring support of the plugin.

commons/src/tracing.rs
```diff
@@ -0,0 +1,85 @@
...
+use opentelemetry_jaeger::{Exporter, Process};
+
+use actix_web::dev::ServiceRequest;
+use actix_web::http;
+use reqwest::header::{HeaderMap, HeaderName, HeaderValue};
+
+/// init_tracer sets up Jaeger tracer
+pub fn init_tracer(name: &'static str) -> thrift::Result<()> {
+    let exporter = Exporter::builder()
+        .with_agent_endpoint("127.0.0.1:6831".parse().unwrap())
```
steveej: I do think we should expose this as a configuration option, e.g. as a `Option<String>` to make it really simple. if that is `None`, we just return `Ok(())` here immediately without setting up anything. Would that result in any error down the line?

vrutkovs: Added PE/GB settings to customize the endpoints, implemented in 8e8c1b676a603ad1833b1f1b8fc72c642f8f14a8

Not entirely sure if "instrument" calls won't crash if provider was not initialized - I'll give it a try

vrutkovs: Update: both GB and PE work as expected if `init_tracer` is omitted

Should we hide it behind the feature flag / config option? I'd prefer not to, so that Jeager could seamlessly receive traces without Cincy restart or reconfiguration

cincinnati/src/plugins/mod.rs
vrutkovs: So I tried implementing this (see https://github.com/vrutkovs/cincinnati/commit/b6da9786a581a3e161ad91a39ad2999386f97d5f and https://github.com/vrutkovs/cincinnati/commit/dd3429667fd457526410fc129d1d13fcd783dc3b) - and I get stack overflow errors. Presumably due to `next_plugin.get_name().await`. Lets iron out other detail and come back to that?

cincinnati/src/plugins/internal/arch_filter.rs
vrutkovs: I'll keep this one not merged yet, as it probably would be replaced by `interface.rs`/`plugins/mod.rs` modifications

vrutkovs:
Quay scraping didn't complete in 5 minutes

/retest

commons/src/tracing.rs
steveej: Is it likely that we need to plug-in Jaeger in that manner? I would assume we'd either know it's running and configure its address appropriately, or we know it's not running and skip the config option.

vrutkovs: `now` -> `not` in "we know it's now running"?

Okay, lets go with that. If the setting is set its configured, if not - the function exits

vrutkovs: Implemented in a4f4cd9

vrutkovs:
Rebased on latest master and reworked plugins to use `Self::PLUGIN_NAME`. Dynamic name discovery throws an odd error - we might want to fix that later.

Removing hold as all review comments have been addressed.

/hold cancel

cincinnati/src/plugins/internal/arch_filter.rs
```diff
@@ -12,6 +12,7 @@ use self::cincinnati::plugins::prelude::*;
 use self::cincinnati::plugins::prelude_plugin_impl::*;
 
 use commons::GraphError;
+
```
steveej: superfluous spacing

cincinnati/src/plugins/mod.rs
steveej: I think https://github.com/openshift/cincinnati/pull/266#issuecomment-641876748 concludes that this poses an error which you're not willing to work through in this PR which is fine by me.

commons/src/tracing.rs
```diff
@@ -0,0 +1,91 @@
...
+struct ClientHeaderMapCarrier<'a>(&'a mut HeaderMap);
+impl<'a> Carrier for ClientHeaderMapCarrier<'a> {
+    fn get(&self, key: &'static str) -> Option<&str> {
+        self.0.get(key).and_then(|value| value.to_str().ok())
+    }
+
+    fn set(&mut self, key: &'static str, value: String) {
+        self.0.insert(
+            HeaderName::from_bytes(key.as_bytes()).unwrap(),
+            HeaderValue::from_str(&value).unwrap(),
```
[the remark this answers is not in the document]
steveej: I tried to implement something like this in https://github.com/vrutkovs/cincinnati/pull/1/files.

steveej (commented):
Please see the comments and also https://github.com/vrutkovs/cincinnati/pull/1 which I filed to remove all the usages of `unwrap`.

cincinnati/src/plugins/mod.rs
vrutkovs: Unfortunately we'd have to have hardcode names in plugins. I'll take another stab at this later

steveej: As I said I'm fine postponing this. Alternatively I can give it a try if you point me to a branch which contains the erroneous code.

vrutkovs: Sure, https://github.com/vrutkovs/cincinnati/commit/jaeger-tracing-plugin-names has commits with my experiments

steveej: I made a working version of the plugin name exposure and split it out to #288. You should be able to rebase on top of that and use the plugin names to create named spans. PTAL and let me know if it works as expected.

vrutkovs:
`took 2.384937281s despite timeout of 2s` and quay issues
/retest

cincinnati/src/plugins/mod.rs
steveej: Could we move this to `fn process` further up at https://github.com/openshift/cincinnati/pull/266/files#diff-2c51cb19b5fdc6531cb418780c1691c9R386?

cincinnati/src/plugins/mod.rs
```diff
@@ -425,8 +432,12 @@ where
         let tx = tx.clone();
 
         std::thread::spawn(move || {
-            let io_future =
-                async { tokio::time::timeout(timeout, process(plugins, initial_io)).await };
+            let span = get_tracer().start("plugins", None);
```
steveej: Could we move this to `fn process` further up at https://github.com/openshift/cincinnati/pull/266/files#diff-2c51cb19b5fdc6531cb418780c1691c9R386?

Justfile
```diff
@@ -144,6 +144,7 @@ run-graph-builder:
 		address = "127.0.0.1"
 		port = 8080
 		path_prefix = "{{path_prefix}}"
+		tracing_endpoint = "localhost:6831"
```
steveej: you could add a global variable e.g. `default_tracing_endpoint` and then reuse that throughout the Justfile

Justfile
```diff
@@ -184,7 +185,7 @@ run-graph-builder-e2e:
 run-policy-engine:
 	#!/usr/bin/env bash
 	export RUST_BACKTRACE=1 RUST_LOG="policy_engine=trace,cincinnati=trace,actix=trace,actix_web=trace"
-	cargo run --package policy-engine -- -vvvv --service.address 0.0.0.0 --service.path_prefix {{path_prefix}} --upstream.cincinnati.url 'http://127.0.0.1:8080/{{path_prefix}}v1/graph' --service.mandatory_client_parameters='channel'
+	cargo run --package policy-engine -- -vvvv --service.address 0.0.0.0 --service.path_prefix {{path_prefix}} --upstream.cincinnati.url 'http://127.0.0.1:8080/{{path_prefix}}v1/graph' --service.mandatory_client_parameters='channel' --service.tracing_endpoint "localhost:6831"
```
steveej: you could add a global variable e.g. `default_tracing_endpoint` and then reuse that throughout the Justfile

commons/src/tracing.rs
```diff
@@ -0,0 +1,107 @@
...
+use opentelemetry::{global, sdk};
+use opentelemetry_jaeger::{Exporter, Process};
+use std::collections::HashMap;
+
+use actix_web::dev::ServiceRequest;
+use actix_web::http;
+use reqwest::header::{HeaderMap, HeaderName, HeaderValue};
+
+/// init_tracer sets up Jaeger tracer
+pub fn init_tracer(name: &'static str, maybe_agent_endpoint: Option<String>) -> thrift::Result<()> {
```
steveej: What's the point of using `thrift::Result` here?

steveej (commented):
Awesome, we're getting close to landing this!

commons/src/tracing.rs
vrutkovs: `opentelemetry_jaeger::Builder.init()` may thrown an error, compatible with `Result`, so `init_tracer` caller has to handle this error

steveej: I think it'd be cleaner if we'd convert it to our error type. The `?` operator is likely to enough to do so.

vrutkovs: Good idea, implemented in 348aa84

commons/src/tracing.rs
```diff
@@ -0,0 +1,113 @@
...
+        .with_agent_endpoint(agent_endpoint)
+        .with_process(Process {
+            service_name: name.to_string(),
+            tags: vec![Key::new("exporter").string("jaeger")],
+        })
+        .init()
+    {
+        Ok(e) => e,
+        Err(e) => return Err(e.into()),
+    };
```
steveej: Did the following not work here?

```suggestion
        .init()?:
```

vrutkovs: It did, but it was throwing `thrift::error`. It probably should stay as an implementation detail of `init_tracer`, thus this commit with a `match`

steveej: Weird, I just tried it locally on top of your branch with the following diff:

```diff
diff --git a/commons/src/tracing.rs b/commons/src/tracing.rs
index 5f0c4d9a..9b7c1ba7 100644
--- a/commons/src/tracing.rs
+++ b/commons/src/tracing.rs
@@ -21,17 +21,13 @@ pub fn init_tracer(name: &'static str, maybe_agent_endpoint: Option<String>) ->
         Some(s) => s,
     };

-    let exporter = match Exporter::builder()
+    let exporter = Exporter::builder()
         .with_agent_endpoint(agent_endpoint)
         .with_process(Process {
             service_name: name.to_string(),
             tags: vec![Key::new("exporter").string("jaeger")],
         })
-        .init()
-    {
-        Ok(e) => e,
-        Err(e) => return Err(e.into()),
-    };
+        .init()?;
```

steveej:
```suggestion
    let exporter = Exporter::builder()
        .with_agent_endpoint(agent_endpoint)
        .with_process(Process {
            service_name: name.to_string(),
            tags: vec![Key::new("exporter").string("jaeger")],
        })
        .init()?;
```

steveej: See my latest review, it works for me like that.

vrutkovs: Oh, got it, Rust can autoconvert this to `Fallible`, pushed bf73b0c576e4acd4ac2fbfc8af5ed9864d807e8e

steveej (commented):
Assuming you're going to squash the commits, please split off the _Cargo.*_ changes into its own commit.
Other than that, I think the code is good to go now :tada: Thanks for addressing the many comments!

steveej (approved):
Thank you!

/lgtm

:tada:

PR #267, opened 2020-04-03
dist/run_containerized_loadtesting: create a script to run load-testing in a container

vrutkovs:
This PR adds a script, which runs load-testing in a dedicated container. In order to run post-build actions on stage we can't reuse `hack/e2e.sh` and should mount targets and wrapper script into a vegeta container.

The followup PRs would add a detailed documentation.

Ref: https://issues.redhat.com/browse/OTA-105

### DIFF
--- a/dist/run_containerized_loadtesting.sh
+++ b/dist/run_containerized_loadtesting.sh
@@ -1,0 +1,16 @@
+#!/usr/bin/env bash
+
+# Pull vegeta container, mount load testing script and run it against GRAPH_URL
+# This is a wrapper around hack/load-testing.sh which we run after stage deployment is complete
+
+set -ex
+
+VEGETA_IMAGE="docker.io/peterevans/vegeta:6.8"
+
+docker run --rm \
+  --volume hack/load-testing.sh:/usr/local/bin/load-testing.sh \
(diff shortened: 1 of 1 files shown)


LalatenduMohanty (commented):
The script looks fine to me. However https://github.com/openshift/cincinnati/blob/master/hack/load-testing.sh is in hack directory but this is in `dist`.  How  do we decide between `hack` and `dist` directories?

LalatenduMohanty:
I do not see any documentation around load testing in either master  or in the PR. As part of the documentation we should have the load testing plan we have (started by Stefan).  The idea is that if anyone wants to understand what kind of load testing we do it is easier to just go through the doc rather then the scripts. Not sure if you folks already had discussions on this and agreed to not include the doc.

vrutkovs:
>How do we decide between hack and dist directories?

Current consensus is that `dist` files are used on production and customer-facing. `hack` scripts are CI scripts, including load-testing on stage.

>The idea is that if anyone wants to understand what kind of load testing we do it is easier to just go through the doc rather then the scripts

Good point, added as a TODO

LalatenduMohanty (approved):
/lgtm

The documentation would be a follow up PR. Cincinnati stage build was failing as the PR to use this file was already merged in the app-sre repository. Hence merging this.

PR #268, opened 2020-04-06
graph-builder: add scrape timeout flag and set to 300s by default

steveej:
The timeout case looks like this in the log:

```
(...)
[2020-04-06T15:27:38Z TRACE cincinnati::plugins::internal::graph_builder::release_scrape_dockerv2::registry] [4.2.7] Downloading layer sha256:de05399aafa027d3bddcff8d4aa011c57689fec00fde83fc744face6af919037
[2020-04-06T15:27:38Z TRACE cincinnati::plugins::internal::graph_builder::release_scrape_dockerv2::registry] [4.2.8] Downloading layer sha256:d08d07c970bc38298f26257215ba27c01be86760228264c0751ed7a3be4eaf5b
[2020-04-06T15:27:38Z ERROR graph_builder::graph] Exceeded timeout of 10s
(... possibly more output of the leaked threads)
```

### DIFF
--- a/Justfile
+++ b/Justfile
@@ -139,6 +139,7 @@ run-graph-builder:
 		verbosity = "vvv"
 
 		[service]
+		scrape_timeout_secs = 300
 		pause_secs = {{pause_secs}}
 		address = "127.0.0.1"
 		port = 8080
--- a/cincinnati/src/plugins/mod.rs
+++ b/cincinnati/src/plugins/mod.rs
@@ -364,6 +364,69 @@ where
     io.try_into()
 }
 
+/// Wrapper around `process` with an optional timeout.
+///
+/// It creates a new runtime per call which is moved to a new thread.
+/// This has the desired effect of timing out even if the runtime is blocked by a task.
+/// It has the sideeffect that these threads are unrecoverably leaked.
+///
+/// These two strategies are used in combination to implement the timeout:
+/// 1. Use the runtime's internal timeout implementation which works for proper async tasks.
+/// 2. Spawn a separate sleeper thread to enforce a deadline of 101% of the timeout
+///    in case the async timeout is not effective.
+pub fn process_blocking<T>(
+    plugins: T,
+    initial_io: PluginIO,
+    timeout: Option<std::time::Duration>,
+) -> Fallible<InternalIO>
+where
+    T: Iterator<Item = &'static BoxedPlugin>,
+    T: Sync + Send,
+    T: 'static,
+{
+    let mut runtime = tokio::runtime::Runtime::new()?;
+
+    let timeout = match timeout {
+        None => return runtime.block_on(process(plugins, initial_io)),
+        Some(timeout) => timeout,
+    };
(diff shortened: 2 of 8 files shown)


steveej:
> @steveeJ: The following test **failed**, say `/retest` to rerun all failed tests:

flake:

```
 --> RUN yum update -y &&     yum install -y openssl &&     yum clean all
Loaded plugins: fastestmirror, ovl
Determining fastest mirrors
Could not retrieve mirrorlist http://mirrorlist.centos.org/?release=7&arch=x86_64&repo=os&infra=container error was
14: curl#6 - "Could not resolve host: mirrorlist.centos.org; Unknown error" 
```

/retest

dist/openshift/cincinnati.yaml
```diff
@@ -265,6 +266,9 @@ parameters:
     value: "350m"
     displayName: "Policy-engine CPU request"
     description: "Requested amount of CPU (millicores) allowed for policy-engine (default: 350m)"
+  - name: GB_SCRAPE_TIMEOUT_SECS
+    value: "300"
+    displayName: Graphb-uilder scrape timeout in seconds
```
```suggestion
    displayName: Graph-builder scrape timeout in seconds
```

cincinnati/src/plugins/mod.rs
```diff
@@ -364,6 +364,67 @@ where
...
+        None => return runtime.block_on(process(plugins, initial_io)),
+        Some(timeout) => timeout,
+    };
+
+    let (tx, rx) = std::sync::mpsc::channel::<Fallible<InternalIO>>();
+
+    {
+        let tx = tx.clone();
+
+        std::thread::spawn(move || {
```
vrutkovs, line 400: IIUC this is necessary so that `release_scraper_dockerv2` could do `tokio::task::spawn_blocking(move || assemble_metadata(&blob, metadata_filename))
            .await`?
Is it necessary to spawn a blocking thread there?

Justfile
```diff
@@ -139,6 +139,7 @@ run-graph-builder:
 		verbosity = "vvv"
 
 		[service]
+		scrape_timeout_secs = 300
```
vrutkovs, line 142: Perhaps its worth excluding initial scrape and setting a 30 sec timeout for subsequent scrapes?

Readiness check would kill the container and roll back the deployment if the initial scrape doesn't pass in 3 + 10*3 + 3*3 = 42 seconds in any case

steveej: It's tempting but I don't think we really need to set this so tightly. This timeout is meant as an emergency break in case the scraping gets stuck. Subsequent to this PR we can work on individual request timeouts.

cincinnati/src/plugins/mod.rs
steveej: The code here tries to deal with the worst case which would be a plugin which completely blocks the processing. The split into two threads here is to have a reliable timeout source.

vrutkovs (approved):
/lgtm
/hold

Lets have stage load tests enabled first to ensure the performance won't regress

PR #271, opened 2020-04-07
dist/run_containerized_loadtesting: fix mounting vegeta targets and wrapper scripts

vrutkovs:
Docker assumes local paths to be volume names by default. Use absolute path
(using `$(pwd)`) to mount vegeta targets and wrapper script.

Fixes the following error on stage:
```
docker: Error response from daemon: create hack/vegeta.targets: "hack/vegeta.targets" includes invalid characters for a local volume name, only "[a-zA-Z0-9][a-zA-Z0-9_.-]" are allowed. If you intended to pass a host directory, use absolute path.
```


LalatenduMohanty (approved):
/lgtm

PR #272, opened 2020-04-07
hack/load-testing: run using sh

vrutkovs:
Image used in `dist/run_containerized_loadtesting.sh` doesn't have `bash`,
so we should be using plain shell instead, as no bash-specific tricks
are required to run the script wrapper.

Fixes `env: can't execute 'bash': No such file or directory` when running containerized load tests

### DIFF
--- a/hack/load-testing.sh
+++ b/hack/load-testing.sh
@@ -1,4 +1,4 @@
-#!/usr/bin/env bash
+#!/usr/bin/env sh
 #
 # This script is used to load test Cincinnati instance.
 # It uses vegeta - `go get -u github.com/tsenart/vegeta`


LalatenduMohanty (approved):
/lgtm

steveej:
```
failed to generate asset \"DNS Config\": getting public zone for \"origin-ci-int-aws.dev.rhcloud.com\": listing hosted zones: Throttling: Rate exceeded\n\tstatus code: 400, request id: c3fcb64b-9eb8-4edd-a356-012158b28ffa"
```

I guess an immediate retest isn't helpful on rate throttling.

PR #274, opened 2020-04-13
WIP: Use tag or sha256 digest for Cincinnati deployment

LalatenduMohanty:
If we just use ${IMAGE}, we can use both the digest and tag for the Cincinnati image.
```
quay.io/app-sre/cincinnati@sha256:a54195540164af7207db1ee9b52deeef2f6dac50284df55d14863f19d7afb0ae
```
Or
```
quay.io/app-sre/cincinnati:c1e2b4e
```

### DIFF
--- a/dist/openshift/cincinnati.yaml
+++ b/dist/openshift/cincinnati.yaml
@@ -30,7 +30,7 @@ objects:
             deploymentconfig: cincinnati
         spec:
           containers:
-            - image: ${IMAGE}:${IMAGE_TAG}
+            - image: ${IMAGE}
               imagePullPolicy: Always
               name: cincinnati-graph-builder
               env:
@@ -80,7 +80,7 @@ objects:
                 - name: configs
                   mountPath: /etc/configs
                   readOnly: true
-            - image: ${IMAGE}:${IMAGE_TAG}
+            - image: ${IMAGE}
               name: cincinnati-policy-engine
               imagePullPolicy: Always
               env:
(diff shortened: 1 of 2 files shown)


dist/openshift/cincinnati.yaml
```diff
@@ -227,13 +227,9 @@ objects:
...
 parameters:
   - name: IMAGE
-    value: "quay.io/app-sre/cincinnati"
-    displayName: cincinnati image
-    description: cincinnati docker image. Defaults to quay.io/app-sre/cincinnati
-  - name: IMAGE_TAG
-    value: "latest"
-    displayName: cincinnati version
-    description: cincinnati version which defaults to latest
+    value: "quay.io/app-sre/cincinnati@sha256:latest"
```
wking: should probably be `:latest` instead of `@sha256:latest`.

dist/openshift/cincinnati.yaml
```diff
@@ -227,13 +227,9 @@ objects:
...
-    value: "quay.io/app-sre/cincinnati"
-    displayName: cincinnati image
-    description: cincinnati docker image. Defaults to quay.io/app-sre/cincinnati
-  - name: IMAGE_TAG
-    value: "latest"
-    displayName: cincinnati version
-    description: cincinnati version which defaults to latest
+    value: "quay.io/app-sre/cincinnati:latest"
+    displayName: cincinnati image name with tag or digest
+    description: cincinnati docker image. Defaults to quay.io/app-sre/cincinnati@sha256:latest
```
steveej, line 232:
```suggestion
    description: cincinnati docker image. Defaults to quay.io/app-sre/cincinnati:latest
```

steveej (changes requested):
/lgtm

Since you marked it as a WIP
/hold

steveej:
> /lgtm
> 
> Since you marked it as a WIP
> /hold

I realize this message is confusing in the presence of a request for change in the same review.
@LalatenduMohanty did yo notice the suggestion I had in a comment? It's only a minor one and I think otherwise this looks good.

PR #275, opened 2020-04-20
Run 2 Cincinnati pods during e2e

vrutkovs:
This PR updates the deployment template:

* support specifying replicas number (default - 1)
* add anti-affinity rule so that Cincinnati pods would be placed on different nodes. This ensures cincinnati pods are not distrupted on node reboots / cluster updates
* add PodDistruptionBudget to ensure at least one Cincinnati pod is running during pod eviction.

Note that pods are not synced in any way, so amount of Github / Quay requests would increase proportionally to replicas. As a result `e2e` script now waits for 10 mins (instead of 5) to make sure initial scrape succeeds on both pods.

2 replica deployment seems to improve response latency - 34.96% of requests handled in 50ms in master now vs. 58.04% with this patch

Ref: OTA-148

### DIFF
--- a/hack/e2e.sh
+++ b/hack/e2e.sh
@@ -88,6 +88,7 @@ oc new-app -f dist/openshift/cincinnati.yaml \
 EOF
 )" \
   -p ENVIRONMENT_SECRETS="{}" \
+  -p REPLICAS="2" \
   ;
 
 # Wait for dc to rollout
--- a/hack/e2e.sh
+++ b/hack/e2e.sh
@@ -92,7 +92,7 @@ EOF
   ;
 
 # Wait for dc to rollout
-oc wait --for=condition=available --timeout=5m deploymentconfig/cincinnati || {
+oc wait --for=condition=available --timeout=10m deploymentconfig/cincinnati || {
     status=$?
     set +e -x
 
--- a/dist/openshift/cincinnati.yaml
+++ b/dist/openshift/cincinnati.yaml
@@ -11,7 +11,7 @@ objects:
         app: cincinnati
       name: cincinnati
     spec:
-      replicas: 1
+      replicas: ${{REPLICAS}}
       selector:
         app: cincinnati
         deploymentconfig: cincinnati
@@ -341,3 +341,5 @@ parameters:
     value: "/etc/configs/gb.toml"
   - name: ENVIRONMENT_SECRETS
     value: '{ "CINCINNATI_GITHUB_SCRAPER_OAUTH_TOKEN_PATH": "/etc/secrets/github_token.key" }'
+  - name: REPLICAS
+    value: "1"
--- a/dist/openshift/cincinnati.yaml
+++ b/dist/openshift/cincinnati.yaml
@@ -29,6 +29,18 @@ objects:
             app: cincinnati
             deploymentconfig: cincinnati
         spec:
+          affinity:
+            podAntiAffinity:
+              preferredDuringSchedulingIgnoredDuringExecution:
+                - weight: 100
+                  podAffinityTerm:
+                    labelSelector:
+                      matchExpressions:
+                        - key: app
+                          operator: In
+                          values:
+                            - cincinnati
+                    topologyKey: kubernetes.io/hostname
           containers:
             - image: ${IMAGE}:${IMAGE_TAG}
               imagePullPolicy: Always
--- a/dist/openshift/cincinnati.yaml
+++ b/dist/openshift/cincinnati.yaml
@@ -201,6 +201,15 @@ objects:
           targetPort: ${{PE_STATUS_PORT}}
       selector:
         deploymentconfig: cincinnati
+  - apiVersion: policy/v1beta1
+    kind: PodDisruptionBudget
+    metadata:
+      name: cincinnati-pdb
+      namespace: cincinnati
+    spec:
+      minAvailable: 1
+    selector:
+      app: cincinnati
   - apiVersion: v1
     kind: ConfigMap
     metadata:


vrutkovs:
/retest

Quay scraping got stuck?

vrutkovs:
`Throttling: Rate exceeded`

/retest

hack/e2e.sh
```diff
@@ -88,10 +88,11 @@ oc new-app -f dist/openshift/cincinnati.yaml \
 EOF
 )" \
   -p ENVIRONMENT_SECRETS="{}" \
+  -p REPLICAS="2" \
```
steveej, line 91: What's the point of having two replicas in the e2e test?

hack/e2e.sh
```diff
@@ -88,10 +88,11 @@ oc new-app -f dist/openshift/cincinnati.yaml \
...
   ;
 
 # Wait for dc to rollout
-oc wait --for=condition=available --timeout=5m deploymentconfig/cincinnati || {
+oc wait --for=condition=available --timeout=10m deploymentconfig/cincinnati || {
```
steveej, line 95: If the replicas are spawned in parallel I don't expect the initialization time to double. The two pods should have all the resources they need since they're placed on different nodes by the anti-affinity. The only limitation would be the registry (Quay in this case) and the bandwidth towards that.

hack/e2e.sh
vrutkovs: Eventually stage / prod would scale pods (so that we could have zero-downtime rollout), so e2e tests ensure this scenario is tested

hack/e2e.sh
vrutkovs: >The only limitation would be the registry (Quay in this case) and the bandwidth towards that.

Yes, it appears Quay is rate-limiting requests from the cluster so initial scrape may not succeed in 5 mins on both pods

LalatenduMohanty (approved):
/hold  for @steveeJ to verify if he is fine with the answers to his comments
/lgtm

hack/e2e.sh
steveej: I don't expect us to require two replicas in production anytime soon. Can we leave this at 1 until we use a local-to-CI registry for tests? It seems like an unnecessary burden on Quay.

hack/e2e.sh
steveej: See https://github.com/openshift/cincinnati/pull/275#discussion_r420090801

hack/e2e.sh
vrutkovs: Being resilent to OSD upgrades and other disruptions is a requirement from app-sre. Its not being enforced yet, but it would be eventually.

Having >1 pod, anti-affinity rules and PDB would eventually become obligatory

steveej: > Having >1 pod, anti-affinity rules and PDB would eventually become obligatory

Why aren't the latter two enough to achieve what we want? I.e, why do we need to set >1 pod? My understanding is that the motivation is to prevent from downtime in case of a node-failure.

steveej: Also, as I mentioned, I'm not against testing this in general, but I'd like to postpone until we don't hit Quay directly with all pods.

steveej:
I'm not in favor of making 100% uptime a requirement in the future, because I don't see the necessity in the case of the Cincinnati service. Therefore I don't think we need to run two concurrent scrapes in every e2e test, putting unnecessary load on Quay, and should postpone the 2-replica-e2e test until we don't hit Quay anymore.

hack/e2e.sh
vrutkovs: >Why aren't the latter two enough to achieve what we want?

During node upgrade it would be taken down and the pod restarted on a different node. In order to have zero-downtime during that we need a backing pod. Anti-affinity would ensure it run on a different node (so that these are not taken down at the same time). PDB would ensure any other pod disrupting operations would keep at least one pod up (e.g. both nodes update at the same time). So anti-affinity and PDB won't do much when the app has one node

vrutkovs: > I'd like to postpone until we don't hit Quay directly with all pods.

We're not violating any contract when we hit Quay more often. Avoiding additional requests to Quay is certainly nice, but its not required and is outside of the scope for this PR (I'll come up with another one to improve that)

vrutkovs:
>I'm not in favor of making 100% uptime a requirement in the future

Our requirement is 95% uptime

> putting unnecessary load on Quay

Is it a significant load? We do ~200 requests every scrape (every hour) on each pod, thus making it 0.05rps. Multiplying this by two won't increase the load on Quay

LalatenduMohanty:
> I'm not in favor of making 100% uptime a requirement in the future, because I don't see the necessity in the case of the Cincinnati service.

I think your reasoning is that an OpenShift cluster's availability is not impacted when Cincinnati is not running.  But it definitly impacts our ability to responding to a new version (which need to be introduced in the graph) which has some CVE fix.  Also when we will suggest customers run Cincinnati in airgap environments , I am sure they would expect it to be highly available as it is the intuitive thing to assume for a service. It also gives us the benefit of taking one node down during maintenance (for app-sre) 

> Therefore I don't think we need to run two concurrent scrapes in every e2e test, putting unnecessary load on Quay, and should postpone the 2-replica-e2e test until we don't hit Quay anymore.

Quay should be resilient enough  to handle this workload as it is a public instance. So we should not be worried about the effect on quay.

LalatenduMohanty:
Also if this helps APP-SRE to streamline their operation/maintenance for all the services (along with Cincinnati) we should do it.

LalatenduMohanty:
@steveeJ If you are ok I am willing to remove the hold and let it merge. Let me know.

hack/e2e.sh
LalatenduMohanty: > Avoiding additional requests to Quay is certainly nice, but its not required and is outside of the scope for this PR (I'll come up with another one to improve that)

@vrutkovs Please file a Jira to track the above comment. We can use the sprint planning to schedule this in our sprint.

LalatenduMohanty:
@steveeJ Thanks, I am going ahead with the PR as I see the world where having Cincinnati better availability is good thing. 

/hold cancel

hack/e2e.sh
vrutkovs: That's OTA-219, me and Trevor discussed it there and most likely we won't go that route - that'd require additional external and stateful layer with little to no benefits

PR #280, opened 2020-05-14
WIP Timeout initial scrape and retry without a pause until it succeeds

vrutkovs:
Timeout initial scrape and retry it without a timeout. This would ensure initial scrape doesn't hang up. Image metadata is cached in graph-builder, so the subsequent retries would finish sooner.

TODO:
* [ ] Cache manifests by digest to avoid re-pulling those
* [ ] Ensure partial graph is not served if initial scrape fails

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -453,6 +453,11 @@ version = "0.11.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
+name = "base64"
+version = "0.12.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+
+[[package]]
 name = "bitflags"
 version = "1.2.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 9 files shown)


steveej (commented):
I'd like to quote a comment I wrote on https://github.com/openshift/cincinnati/pull/264#issuecomment-607307896:

> However, I would like to discuss what happens if for some reason the scraping fails a few times in a row. With a delay as low as 30s this is not a concern, but with 10m it is. My hunch here is that we want a smarter algorithm rather than a higher default delay. Some sort of adaptive back-off algorithm with a defined minimum and maximum value would be ideal.

vrutkovs:
>I would like to discuss what happens if for some reason the scraping fails a few times in a row.

That's a different problem - initial scrape is special and should be handled same as other scrapes. Backoff retries should be implemented in a different PR

steveej:
> That's a different problem - initial scrape is special and should be handled same as other scrapes. Backoff retries should be implemented in a different PR

From the perspective of the services that we're scraping it's the same problem. Therefore I don't think it's a good idea to hardcode hammering them. Imagine if all consumers of a service would behave in this way, that would practically create a DDoS scenario. 
We could implement the [circuit breaker pattern](https://en.wikipedia.org/wiki/Circuit_breaker_design_pattern) quite easily. If we encounter an error we set an error state in the main loop. Subsequently we pass a specific parameter to the plugin chain, which causes them to do a "check connection" command, and doesn't try a full-fledged scrape. The main loop delay could still vary dynamically using some form of backoff algorithm, e.g. [exponential backoff](https://en.wikipedia.org/wiki/Exponential_backoff).

steveej:
@vrutkovs do you still want to pursue this or something similar to it? If so I suggest to create a Jira ticket for it so we can pick it up in the next grooming session.

PR #282, opened 2020-05-14
dist/openshift/cincinnati: bump GB pod memory limit

vrutkovs:
Temporarily increase memory limit for graph-builder container. Too many releases cause the memory to grow rapidly, which causes graph-builder being killed


wking:
> Too many releases...

We haven't cut _that_ many releases.  Is Cincinnati trying to hold lots of fat layers in memory or something?  Ideally we'd be able to stream the layers through without holding even one layer tarball in memory at once, because the metadata files we're hunting for are both really tiny and often included in the image's top-most layer.

PR #285, opened 2020-06-03
Replace whitelist/blacklist terminology by allowlist/denylist

thiagoalessio:
I've seen [rails](https://github.com/rails/rails/pull/33681) and [elixir](https://github.com/elixir-lang/elixir/pull/10078/rails#33681) doing this, and would like to suggest the same for all our openshift repos.


steveej:
Thanks!

/lgtm

cincinnati/src/plugins/internal/graph_builder/github_openshift_secondary_metadata_scraper/plugin.rs
```diff
@@ -88,7 +88,7 @@ pub struct GithubOpenshiftSecondaryMetadataScraperSettings {
     /// Vector of regular expressions used as a positive output filter.
     /// An empty vector is regarded as a configuration error.
     #[default(DEFAULT_OUTPUT_WHITELIST.iter().map(|s| (*s).to_string()).collect())]
-    output_whitelist: Vec<String>,
+    output_allowlist: Vec<String>,
```
lucab, line 91: FYI this is a breaking change in the configuration format (as shown in the test change below).
I don't know whether anybody is relying on this at the moment though.

PR #289, opened 2020-06-24
plugins/release_scrape_dockerv2: include port in release source

steveej:
This fixes ReleaseScrapeDockerv2Plugin implementation to include the
port in the source url for each release.
In order to achieve this a few actions were taken:

* Pass the a `Registry` instance all the way down to `lookup_or_fetch`
  to have the methods available instead of just the Strings.

* Extend the tests to check for the port exposure

* Cache: store `Metadata` instead of `Release`, since the former does
  not store the source. For situations where the cache is used throughout
  calls given different registries (as is the case in some tests) it seems
  more elegant to assemble a new `Release` with the correct source,
  instead of overwriting the source of a previously cached release.

In addition this prepares for removing the scheme handling from the
registry URL, because this is not valid according to Docker and podman.


LalatenduMohanty (approved):
/lgtm

PR #290, opened 2020-06-25
Follow HTTP redirects

yprokule:
Without it fails like:
```bash
UPSTREAM=https://openshift-release.svc.ci.openshift.org/graph CHANNEL=stable-4.4 bash -x available-updates.sh 4.4.9
+ UPSTREAM=https://openshift-release.svc.ci.openshift.org/graph
+ CHANNEL=stable-4.4
+ ARCH=amd64
+ VERSION=4.4.9
+ test -z 4.4.9 -o 1 -ne 1
+ URI='https://openshift-release.svc.ci.openshift.org/graph?channel=stable-4.4&arch=amd64'
++ curl -sH Accept:application/json 'https://openshift-release.svc.ci.openshift.org/graph?channel=stable-4.4&arch=amd64'
+ DATA='<html>
<head><title>302 Found</title></head>
<body>
<center><h1>302 Found</h1></center>
<hr><center>nginx/1.17.10</center>
</body>
'/html>
+ test -z '<html>
<head><title>302 Found</title></head>
<body>
<center><h1>302 Found</h1></center>
<hr><center>nginx/1.17.10</center>
</body>
'/html>
+ echo '<html>
<head><title>302 Found</title></head>
<body>
<center><h1>302 Found</h1></center>
<hr><center>nginx/1.17.10</center>
</body>
'/html>
+ jq -r '
  (.nodes | with_entries(.key |= tostring)) as $nodes_by_index |
  [
    .edges[] |
    select($nodes_by_index[(.[0] | tostring)].version == "4.4.9")[1] |
    tostring |
    $nodes_by_index[.]
  ] | sort_by(.version)[] |
  .version + "\t" + .payload + "\t" + (.metadata.url // "-")
'
parse error: Invalid numeric literal at line 1, column 7
```

PR #291, opened 2020-06-25
Fix options order

yprokule:
Fix introduced in https://github.com/openshift/cincinnati/pull/290 doesn't work cause `-L` option interrupts `-H`option

### DIFF
--- a/hack/available-updates.sh
+++ b/hack/available-updates.sh
@@ -12,7 +12,7 @@ then
 fi
 
 URI="${UPSTREAM}?channel=${CHANNEL}&arch=${ARCH}"
-DATA="$(curl -sHL Accept:application/json "${URI}")"
+DATA="$(curl --silent --location --header 'Accept:application/json' "${URI}")"
 if test -z "${DATA}"
 then
 	 echo "Failed to fetch data from ${URI}"


steveej (commented):
Ha, I didn't catch that last time. What do you think about converting the short options to long options? Seem less error prone.

yprokule:
> Ha, I didn't catch that last time. What do you think about converting the short options to long options? Seem less error prone.

makes sense, updated

steveej (approved):
Thanks again!

/lgtm

steveej:
@vrutkovs please take a look at [this error in the e2e tests](https://deck-ci.apps.ci.l2s4.p1.openshiftapps.com/view/gcs/origin-ci-test/pr-logs/pull/openshift_cincinnati/291/pull-ci-openshift-cincinnati-master-e2e/1277904981348847616#1:build-log.txt%3A383). I don't understand the cause of it.

vrutkovs:
```
 thread 'check_slo::_cincinnati_gb_graph_upstream_errors_total_0_' panicked at 'assertion failed: `(left == right)`
  left: `"1"`,
 right: `"0"`', e2e/tests/slo.rs:55:5 
```
quay scraping errors, its transient

PR #297, opened 2020-08-20
Output digraph content from Cincinnati json

jupierce:
Drop in replacement for graph.sh, but defaults to excluding hotfixes releases.

### DIFF
--- a/hack/graph.py
+++ b/hack/graph.py
@@ -1,0 +1,62 @@
+#!/usr/bin/env python3
+
+import argparse
+import sys
+import json
+from typing import Dict, List
+
+
+def run():
+    parser = argparse.ArgumentParser(description=f'Output digraph data for Cincinnati json',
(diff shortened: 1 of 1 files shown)


steveej:
> /retest

I'm afraid there's currently a bug on the test platform which prevents the e2e tests from passing. But scripts like this I don't see an issue if we override the e2e tests once we decide to merge it.

On the content of the script, would it be interesting for you to support this format on the server-side so the client could just specify a specific value in the `Accept` header to receive it?

jupierce:
Hey @steveeJ - Unfortunately I don't know much about the server side. I was pointed to https://github.com/openshift/cincinnati/blob/master/hack/graph.sh as the source of graphs like: https://cincinnati-web-graph.op.redhat.com/export?channel-name=fast&channel-version=4.5 . The impetus for the PR at the time being the desire to hide hotfixes in this visualization. 
At this point, I don't know if anyone sees that as a priority, but I do think the python implementation offers a better basis on which to augment the visualization in the future.

steveej:
As this will go into _hack_ I don't think the maintenance burden is on the team but it's related enough to live in this repo.

/lgtm

PR #300, opened 2020-09-07
dist: shift rust version to include 1.46

steveej:


dist/Dockerfile.build/Dockerfile
```diff
@@ -11,8 +11,8 @@ ENV HOME="/root"
 ENV PATH="${HOME}/.cargo/bin:${PATH}"
 
 # build: Rust stable toolchain + 2 previous versions
-RUN curl https://sh.rustup.rs -sSf | sh -s -- --default-toolchain 1.45.0 -y && \
-  rustup install 1.42.0 && \
+RUN curl https://sh.rustup.rs -sSf | sh -s -- --default-toolchain 1.46.0 -y && \
+  rustup install 1.46.0 && \
   rustup install 1.43.0 && \
   rustup install 1.44.0 && \
   rustup default 1.45.0
```
vrutkovs:
```suggestion
  rustup default 1.46.0
```

steveej: ~~line 15 already contains 1.46.0. I used the minimal diff approach to bump here.~~

I think we should handle the default differently altogether.

steveej: I removed the `rustup default` command as it's redundant with the `--default-toolchain` argument given above.

steveej:
Rust 1.40 and 1.41 are no longer in the image.

/override  ci/prow/rustfmt-1.40.0
/override  ci/prow/rustfmt-1.41.0

dist/Dockerfile.build/Dockerfile
```diff
@@ -11,11 +11,10 @@ ENV HOME="/root"
 ENV PATH="${HOME}/.cargo/bin:${PATH}"
 
 # build: Rust stable toolchain + 2 previous versions
-RUN curl https://sh.rustup.rs -sSf | sh -s -- --default-toolchain 1.45.0 -y && \
-  rustup install 1.42.0 && \
+RUN curl https://sh.rustup.rs -sSf | sh -s -- --default-toolchain 1.46.0 -y && \
```
vrutkovs: That installs 4 Rust versions - 1.43, 1.44, 1.45 and 1.46:
```
  1.46.0 installed - rustc 1.46.0 (04488afe3 2020-08-24)
...
  1.43.0-x86_64-unknown-linux-gnu installed - rustc 1.43.0 (4fb7144ed 2020-04-20)
...
  1.44.0-x86_64-unknown-linux-gnu installed - rustc 1.44.0 (49cae5576 2020-06-01)
...
  1.45.0-x86_64-unknown-linux-gnu installed - rustc 1.45.0 (5c1f21c3b 2020-07-13)

```
from https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_cincinnati/300/pull-ci-openshift-cincinnati-master-images/1303000667190726656/artifacts/build-logs/buildroot.log

steveej: I'm aware of that. We previously did this to not break CI: as we shift the 3-installed-versions window asynchronous to shifting the 3-tested-versions window. We could stop doing that until we grow back our 3-tested-versions window after [we're cutting it to 1](https://github.com/openshift/release/pull/11523). Shall we?

vrutkovs: Let stick to one version then and add another one (same as in CPaaS builds) then

steveej:
Rust 1.40 and 1.41 are no longer in the image.

/override ci/prow/rustfmt-1.40.0
/override ci/prow/rustfmt-1.41.0

vrutkovs:
Seem kcov no longer supports `--verify` option - https://github.com/openshift/cincinnati/pull/301

steveej:
@vrutkovs 

> Seem kcov no longer supports `--verify` option - #301

It still supports that option. The issue was that the test binary location changed. I added a commit to adapt the script. PTAL

vrutkovs:
Looks good, I believe it also needs https://github.com/vrutkovs/cincinnati/commit/98e30dde9854754ab7a7c387eb7f6225d675d166 to fix `images` test.

UPD: https://github.com/vrutkovs/cincinnati/commit/8e89729fca04a8431c22cb93a6f9ab4ebf2d2652 is a better solution: build all tests, extract filenames from Rust build info, copy them to `/opt/cincinnati/bin` and remove temporary prefix

vrutkovs (changes requested):
https://github.com/vrutkovs/cincinnati/commit/8e89729fca04a8431c22cb93a6f9ab4ebf2d2652 to fix image build, otherwise it won't be promoted

vrutkovs:
>thread 'main' panicked at 'already borrowed: BorrowMutError', /rustc/5c1f21c3b82297671ad3ae1e8c942d2ca92e84f2/src/libcore/cell.rs:877:9

steveej:
> thread 'main' panicked at 'already borrowed: BorrowMutError', /rustc/5c1f21c3b82297671ad3ae1e8c942d2ca92e84f2/src/libcore/cell.rs:877:9

I think that's a red herring and only a side-effect of a bug in the cherry-picked code for which I just pushed a fix. We'll know soon :wink:

vrutkovs (approved):
/lgtm

PR #301, opened 2020-09-07
cargo_test: remove '--verify' kcov arg

vrutkovs:
New kcov versions don't seem to support it anymore, breaking 
`cargo-test`


vrutkovs:
Doesn't seem to help

PR #302, opened 2020-09-10
Dockerfile.e2e: remove duplicate yq install

vrutkovs:
e2e image uses Dockerfile.build as a base, which already has jq installed.
Previously duplicate jq install was added to work around base image
caching - and now its no longer necessary.

### DIFF
--- a/dist/Dockerfile.e2e/Dockerfile
+++ b/dist/Dockerfile.e2e/Dockerfile
@@ -2,7 +2,6 @@ FROM quay.io/app-sre/cincinnati:builder AS rust_builder
 # build e2e tests
 COPY . .
 
-RUN yum -y install jq
 RUN hack/build_e2e.sh
 
 FROM registry.svc.ci.openshift.org/openshift/release:golang-1.13 AS golang_builder


steveeJ:
When I tried to add it only to the base image I found that it didn't work. I may be that the base image is only propagated after the PR is merged, and so it wasn't possible to introduce the change in the same PR with the change that depended on it. It may also be that we don't understand what the actual base image of the e2e image is.

steveej (commented):
When I tried to add it only to the base image I found that it didn't work. I may be that the base image is only propagated after the PR is merged, and so it wasn't possible to introduce the change in the same PR with the change that depended on it. It may also be that we don't understand what the actual base image of the e2e image is.

vrutkovs:
>I may be that the base image is only propagated after the PR is merged,

Correct

steveej:
The e2e test fails with:

```
   * could not run steps: step e2e failed: could not wait for template instance to be ready: failed to create objects: Pod "e2e" is invalid: spec.containers[4].name: Duplicate value: "artifacts" 
```

Is this related to https://github.com/openshift/release/pull/11174?

steveej:
> Yep. We might want to override it, as `jq` is being called during image build anyway

I understand that this PR is not the cause of the failure.
However I don't understand what would resolve the failure and I'd prefer to do that first and then rerun the test here.

steveej (approved):
/lgtm

PR #304, opened 2020-09-15
Change visibility of plugin classes

vrutkovs:
This exposes some modules, functions and class variables so that cincinnati-graph-data CI could re-use Cincinnati models.

* `graph_data_model` is made public so that CI tool could import `BlockedEdge` and `Channel` classes
* `deserialize_directory_files` function is public so that CI tool could convert YAMLs into `BlockedEdge`/`Channel` instances
* `BLOCKED_EDGES_DIR`/`CHANNELS_DIR` const made public to avoid hardcoding those in CI tool
* `ReleaseScrapeDockerv2Settings` made public so that CI tool could reuse registry configuration and verify all blocked-edge/channel release are uploaded to Quay

### DIFF
--- a/cincinnati/src/plugins/internal/graph_builder/openshift_secondary_metadata_parser/plugin.rs
+++ b/cincinnati/src/plugins/internal/graph_builder/openshift_secondary_metadata_parser/plugin.rs
@@ -204,6 +204,9 @@ where
     Ok(t_vec)
 }
 
+pub static BLOCKED_EDGES_DIR: &str = "blocked-edges";
+pub static CHANNELS_DIR: &str = "channels";
+
 impl OpenshiftSecondaryMetadataParserPlugin {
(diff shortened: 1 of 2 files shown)


steveej (commented):
I'm a bit wary to make this code public as I have not especially designed it to be used by other code. Could we maintain the consumer of it in this repo so that we can ensure by CI tests that we don't break the consumer accidentally?

vrutkovs:
> Could we maintain the consumer of it in this repo so that we can ensure by CI tests that we don't break the consumer accidentally?

The consumer of this repo is https://github.com/openshift/cincinnati-graph-data/pull/433. It would verify that blocked-edge / channels are valid and scrape Quay images. This is covered by e2e tests.

I'm not too much worried about breaking future cincinnati-graph-data CI - it could point to a particular commit and update on request

steveej (approved):
/lgtm

PR #305, opened 2020-09-16
Add rh-manifest-generator and dist/vendor script

steveej:
These tools can be used to prepare the source code for Red Hat's internal build system which works offline and expects a security manifest.

/cc @LalatenduMohanty @vrutkovs

### DIFF
--- a/dist/vendor.sh
+++ b/dist/vendor.sh
@@ -1,0 +1,33 @@
+#!/usr/bin/env sh
+
+set -ueo pipefail
+
+CARGO_CONFIG=".cargo/config.toml"
+FILES_TO_COMMIT=(
+  "${CARGO_CONFIG}"
+  "vendor"
+  "rh-manifest.txt"
+)
+
+# Don't overwrite any existing config
+[ ! -f "${CARGO_CONFIG}" ] || {
+  echo WARNING ${CARGO_CONFIG} exists, moving away..
+  mv --backup=existing ${CARGO_CONFIG}{,.prev}
+}
+
+set -x
+
+# generate the rh-manifest.txt
+cargo run --bin rh-manifest-generator
+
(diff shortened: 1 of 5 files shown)


dist/vendor.sh
```diff
@@ -0,0 +1,31 @@
...
+  "rh-manifest.txt"
+)
+
+# Don't overwrite any existing config
+[ ! -f "${CARGO_CONFIG}" ] || {
+  echo WARNING ${CARGO_CONFIG} exists, moving away..
+  mv --backup=existing ${CARGO_CONFIG}{,.prev}
+}
+
+set -xueo pipefail
```
vrutkovs: Lets move this on line 2?

steveej: I purposely moved this past the variable definitions and checks. Ideally we'd probably `set -ueo pipefail` at line 2 and only leave `set -x` here?

vrutkovs: Right, I'm mostly concerned about having `-ueo` part as early as possible

vrutkovs:
Looks good, but I'm not sure how its expected to run: should the developer run it or the automated script would run this on CI?

This is a bit different from the workflow other repos are using:
* developer updates code and runs the tool to include vendoring
* CI would ensure vendored code is updated and no git diff is produced.

This would make sure the commit along with the vendored sources is tested. If we plan to have tool which makes commits with vendoring code behind our backs we might miss some e2e test failures.

steveej:
> should the developer run it or the automated script would run this on CI?

While I initially thought of automating it, I also reached the conclusion of having the developer run this manually. In our repository I would suggest that we only run it on release branches and not for the master branch.

LalatenduMohanty (approved):
/lgtm

PR #306, opened 2020-09-17
hack/e2e: copy kubeconfig to /tmp so that it can be mutated

vrutkovs:
Switching projects mutates kubeconfig, which can be mounted as read-only. e2e script should copy it to writable location.

See https://github.com/openshift/release/pull/11904, which fails with
```
error: open /var/run/secrets/ci.openshift.io/multi-stage/kubeconfig: read-only file system
error: failed to execute wrapped command: exit status 1 
```


vrutkovs:
Install issue

/test e2e

steveej (approved):
/lgtm

PR #307, opened 2020-09-23
plugins/openshift_secondary_metadata_parser: add option to abort on error

steveej:
This adds a configuration option to set conditions as errors which were
previously unconditionally tolerated.

This is meant to complement https://github.com/openshift/cincinnati-graph-data/pull/433/

TODO:
* [x] Discuss whether or not to include a _run-once_ option to the graph-builder which would further reduce the size of the linked PR

/cc @vrutkovs


cincinnati/src/plugins/internal/graph_builder/openshift_secondary_metadata_parser/plugin.rs
```diff
@@ -148,41 +154,103 @@ impl PluginSettings for OpenshiftSecondaryMetadataParserSettings {
...
     T: DeserializeOwned,
 {
+    use std::sync::Arc;
+    use std::sync::Mutex;
     use tokio::stream::Stream;
     use tokio::stream::StreamExt;
 
+    // Even though we don't use concurrent threads, the usage of async forces us
+    // to guarantee that the error container is thread-safe
+    let error: Arc<Mutex<Option<Error>>> = Arc::new(Mutex::new(None));
```
vrutkovs, line 189: I used [appendlist](https://docs.rs/appendlist/1.4.0/appendlist/) in cincinnati-graph-data PR - perhaps its would be useful here?

steveej: I'm happy to discuss that approach as well.

I think I want to start backwards, focusing on what Error type do we want in the Result of the function? I see the following options:

(a) A single Error -- this is what I chose
(b) A collection of Error

Choosing (a) it makes sense to me to use [anyhow::Context](https://docs.rs/anyhow/1.0.32/anyhow/trait.Context.html) to append all the errors into a single instance of `anyhow::Error`. This has the advantage of not requiring to iterate through the list of errors later to consolidate them into a single Error.

If we would choose (b) I think using the _appendlist_ crate would make sense as we could likely transform it into a `Vec`.

So the question we're dealing with now: what's better for the consumer of the function?

vrutkovs:
>Discuss whether or not to include a run-once option to the graph-builder which would further reduce the size of the linked PR

I'd prefer this to happen after Rust CI is working on cincinnati-graph-data PRs. It would be easier to verify it

cincinnati/src/plugins/internal/graph_builder/openshift_secondary_metadata_parser/plugin.rs
vrutkovs: Returning a single error seems to be the obvious choice - we don't need special treatment for the `Vec` of errors, other functions are returning error message as a string.

Its possible that `cincinnati-graph-data` PR will make a change to both "blocked-edges" and "channels" with invalid yamls in both, but it has been extremely rare. As a result I'm fine with throwing a string error and exiting early

vrutkovs: Also `appendlist` is great for appending without mutation, but I'm not sure if its thread-safe. it cannot be easily converted to `Vec` though, so +1 to current implementation

steveej:
> I'd prefer this to happen after Rust CI is working on cincinnati-graph-data PRs. It would be easier to verify it

I'm not sure what you want to verify. The tests I added cover the new functionality and the existing tests ensure that the changes don't break existing code.

vrutkovs (approved):
/lgtm

Lets leave run-once option for later, when cincinnati-graph-data CI is setup

steveej:
> Lets leave run-once option for later, when cincinnati-graph-data CI is setup

:+1: 

While CI is still stuck on e2e tests, do you want to bikeshed a bit about the names of the `deserialize_directory_files` function name and the `DeserializeDirectoryFilesError` enum name or shall we leave it as is?

vrutkovs:
names look good to me

PR #309, opened 2020-09-26
dist,hack: add scripts to manage CI credentials

steveej:


dist/prepare_ci_credentials.sh
```diff
@@ -0,0 +1,15 @@
...
+set -xeuo pipefail
+
+test -f "${CINCINNATI_CI_DOCKERCFG_PATH}"
+test -f "${CINCINNATI_CI_PUBLIC_DOCKERCFG_PATH}"
+
+touch "${CINCINNATI_CI_DOCKERJSON_PATH}"
+touch "${CINCINNATI_CI_PUBLIC_DOCKERJSON_PATH}"
+
+# consolidate the quay credentials with the private CI registry credentials to allow a seamless transition to the latter
+jq -s '{ "auths": .[0] } * .[1]' "${CINCINNATI_CI_DOCKERCFG_PATH}" "${CINCINNATI_TEST_CREDENTIALS_PATH}" > "${CINCINNATI_CI_DOCKERJSON_PATH}"
```
LalatenduMohanty, line 14: What is `${CINCINNATI_TEST_CREDENTIALS_PATH}"` ?

dist/prepare_ci_credentials.sh
```diff
@@ -0,0 +1,15 @@
...
+jq <"${CINCINNATI_CI_PUBLIC_DOCKERCFG_PATH}" '{ "auths": . }' > "${CINCINNATI_CI_PUBLIC_DOCKERJSON_PATH}"
```
LalatenduMohanty, line 15: Without looking in to ` hack/get_ci_secrets.sh` it is hard to understand this script. Can you add a comment that the variables can be set running ` hack/get_ci_secrets.sh` for CI runs?

LalatenduMohanty (changes requested):

dist/prepare_ci_credentials.sh
steveej: > hack/get_ci_secrets.sh

That's only meant for local development to set up the environment that exists in CI. So on CI we don't want to run that hack script.

hack/get_ci_secrets.sh
```diff
@@ -0,0 +1,31 @@
+#!/usr/bin/env bash
+#
+# This script fetches the secrets from the CI cluster and populates local files with it.
+# This script is meant to be sourced.
```
LalatenduMohanty: Can we mention that this file is mostly for the local development to set up the environment that exists in CI?

LalatenduMohanty (changes requested):

hack/get_ci_secrets.sh
LalatenduMohanty: In its current form it would be hard to comprehend with respect to its use by anyone who is new to Cincinnati.

steveej: To me it's hard to comprehend how we would rely on anything in a directory called `hack` for anything other than local development environments :wink: I did change the description though. PTAL

steveej:
Seems to be some unusual latency on the CI node.

/retest

LalatenduMohanty (approved):
/lgtm

PR #311, opened 2020-09-29
e2e: deref copied kubeconfig

vrutkovs:
In multistage test kubeconfig is mounted from a secret, so its a reference to `../data/kubeconfig`.
Copy kubeconfig with `-L` so that `/tmp/kubeconfig` would be a plain file.

This also updates the location for pull-secret

--

Closes #310

### DIFF
--- a/hack/e2e.sh
+++ b/hack/e2e.sh
@@ -27,7 +27,9 @@ PULL_SECRET=${PULL_SECRET:-/tmp/cluster/pull-secret}
 
 set -euo pipefail
 # Copy KUBECONFIG so that it can be mutated
-cp -rvf $KUBECONFIG /tmp/kubeconfig
+cp -Lrvf $KUBECONFIG /tmp/kubeconfig
+ls -la /tmp/kubeconfig
+cat /tmp/kubeconfig
 export KUBECONFIG=/tmp/kubeconfig
 
 # Create a new project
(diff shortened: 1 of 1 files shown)


hack/e2e.sh
```diff
@@ -23,11 +23,11 @@ echo "IMAGE=${IMAGE}"
 echo "IMAGE_TAG=${IMAGE_TAG}"
 
 # Use defined PULL_SECRET or fall back to CI location
-PULL_SECRET=${PULL_SECRET:-/tmp/cluster/pull-secret}
+PULL_SECRET=${PULL_SECRET:-/var/run/secrets/ci.openshift.io/cluster-profile/pull-secret}
```
steveej, line 26: out of curiosity, where's this documented?

steveej (approved):
/lgtm

hack/e2e.sh
vrutkovs: [ipi-install-install step](https://ci-operator-configresolver-ui-ci.apps.ci.l2s4.p1.openshiftapps.com/reference/ipi-install-install) mentions pull secret location at line 15 (not sure if we have access to `CLUSTER_PROFILE_DIR` in this step, so I used the path where it gets mounted in the test container)

vrutkovs:
Tests are still failing (duplicate secrets), investigating why in https://github.com/openshift/release/pull/12291

/override ci/prow/e2e

PR #308, opened 2020-09-26
plugins/release_scrape_dockerv2: switch to using the CI registry

steveej:
This changes the plugin to use the images stored on the CI registry.
CI can then be configured to automatically update these images from the
content which is currently stored in `graph-builder/tests/images`.

TODO:
- [x] Split out test refactors into a separate PR #312 and #313 
- [x] Requires #312
- [x] Requires #313
- [x] Requires https://github.com/openshift/release/pull/12267.
- [x] Split out the e2e parts to #340

PR #312, opened 2020-10-02
Refactor the registry tests to operate on the plugin scope 

steveej:
- [x] Requires #313.

Split from #308.


steveej:
/hold

- [x] #313 shall go in first.

e2e/tests/e2e.rs
```diff
@@ -58,10 +58,14 @@ fn e2e_channel_success(channel: &'static str, arch: &'static str) {
...
-            "io.openshift.upgrades.graph.previous.remove",
-        ],
+        cincinnati::testing::CompareGraphsVerboseSettings {
+            unwanted_metadata_keys: &[
+                "io.openshift.upgrades.graph.previous.remove_regex",
+                "io.openshift.upgrades.graph.previous.remove",
+            ],
+
+            ..Default::default()
+        },
```
LalatenduMohanty, line 68: @steveeJ I am not sure what is this for? Why is this called `unwanted_metadata_keys`? Can you add more context to the commit?

LalatenduMohanty: This came from https://github.com/openshift/cincinnati/pull/312/commits/4cefba246adc9d37ef8965634c405e7a1b569a6d but still, a bit explanation/documentation/comment will help.

LalatenduMohanty (changes requested):
We should squash https://github.com/openshift/cincinnati/pull/312/commits/ebceead1d7f448f4d0d07511202d8b2ed22657ad and  https://github.com/openshift/cincinnati/pull/312/commits/4cefba246adc9d37ef8965634c405e7a1b569a6d  as there are essentially related code changes.

e2e/tests/e2e.rs
steveej: It effectively allows us to omit arguments to the `graph_compare_graph` and leaving them at their default.

steveej: We don't want to include metadata with the given keys in the graph comparison, so we remove these before comparing the graphs.

steveej:
> We should squash [ebceead](https://github.com/openshift/cincinnati/commit/ebceead1d7f448f4d0d07511202d8b2ed22657ad) and [4cefba2](https://github.com/openshift/cincinnati/commit/4cefba246adc9d37ef8965634c405e7a1b569a6d) as there are essentially related code changes.

All comments addressed, thanks for the review!

cincinnati/src/lib.rs
```diff
@@ -1017,6 +1040,45 @@ pub mod testing {
...
+                            Error::msg(format!("invalid version string {:?}", version))
+                        })?;
+                        release.payload = format!("{}:{}", source_front, version);
+
+                        Ok(())
+                    }
+                    _ => panic!("should not get here"),
+                }
+            })
+            .context("Post-processing the received graph to match the test expectations")
```
LalatenduMohanty: This message `Post-processing the received graph to match the test expectations` is only specific to tests and does not sound right.

cincinnati/src/lib.rs
```diff
@@ -1017,6 +1040,45 @@ pub mod testing {
...
+                        // replace digest by tag to match expectency
+                        let version = release.version.to_string();
+                        release.payload = version;
+
+                        Ok(())
+                    }
+                    _ => panic!("should not get here"),
+                }
+            })
+            .context("Post-processing the received graph to match the test expectations")
```
LalatenduMohanty: same here. https://github.com/openshift/cincinnati/pull/312/commits/02f94bc9d668970731d224b9db755f3cbf878e03#r515120447

cincinnati/src/lib.rs
```diff
@@ -1017,6 +1040,45 @@ pub mod testing {
...
+            .iter_releases_mut(|ref mut release| {
+                match release {
+                    Release::Concrete(ref mut release) => {
+                        // replace digest by tag to match expectency
+                        let version = release.version.to_string();
+                        release.payload = version;
+
+                        Ok(())
+                    }
+                    _ => panic!("should not get here"),
```
LalatenduMohanty: Another console message which I think should be modified to be more meaningful. Forgot to point this out in last review.

LalatenduMohanty (changes requested):

cincinnati/src/lib.rs
```diff
@@ -1017,6 +1040,45 @@ pub mod testing {
...
+                        // replace digest by tag to match expectency
+                        let version = release.version.to_string();
+                        let source_front = release.payload.split('@').next().ok_or_else(|| {
+                            Error::msg(format!("invalid version string {:?}", version))
+                        })?;
+                        release.payload = format!("{}:{}", source_front, version);
+
+                        Ok(())
+                    }
+                    _ => panic!("should not get here"),
```
LalatenduMohanty: https://github.com/openshift/cincinnati/pull/312/files#r515210503

LalatenduMohanty (approved):
/lgtm

PR #313, opened 2020-10-02
Bump dkregistry and detect when authentication without credentials is required when scraping a registry

steveej:
See https://github.com/camallo/dkregistry-rs/pull/168 for the upstream change and the problem description.

- [x] Requires https://github.com/camallo/dkregistry-rs/pull/180
- [x] Requires https://github.com/camallo/dkregistry-rs/pull/181

Testing the scraper plugin against the OpenShift internal registry revealed that it requires anonymous authentication as well, so this will be needed for the effort of moving our test images there.

### DIFF
--- a/cincinnati/src/plugins/internal/graph_builder/release_scrape_dockerv2/registry/mod.rs
+++ b/cincinnati/src/plugins/internal/graph_builder/release_scrape_dockerv2/registry/mod.rs
@@ -210,22 +210,32 @@ pub async fn fetch_releases(
     concurrency: usize,
 ) -> Result<Vec<cincinnati::plugins::internal::graph_builder::release::Release>, Error> {
     let registry_client = {
-        let authenticate = username.is_some() && password.is_some();
-
         let client_builder = dkregistry::v2::Client::configure()
             .registry(&registry.host_port_string())
             .insecure_registry(registry.insecure);
+        let scope = format!("repository:{}:pull", &repo);
 
-        if authenticate {
+        if username.is_some() && password.is_some() {
             client_builder
                 .username(username.map(ToString::to_string))
                 .password(password.map(ToString::to_string))
                 .build()?
-                .authenticate(&[&format!("repository:{}:pull", &repo)])
-                .await
+                .authenticate(&[&scope])
+                .await?
         } else {
-            client_builder.build()
-        }?
+            let client = client_builder.build()?;
+
+            if client
+                .is_v2_supported_and_authorized()
+                .await
+                .map(|(_, authorized)| authorized)?
+            {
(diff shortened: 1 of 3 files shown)


cincinnati/src/plugins/internal/graph_builder/release_scrape_dockerv2/registry/mod.rs
```diff
@@ -208,9 +208,10 @@ pub async fn fetch_releases(
     cache: cache::Cache,
     manifestref_key: &str,
     concurrency: usize,
+    force_auth: bool,
 ) -> Result<Vec<cincinnati::plugins::internal::graph_builder::release::Release>, Error> {
     let registry_client = {
-        let authenticate = username.is_some() && password.is_some();
+        let authenticate = force_auth || (username.is_some() && password.is_some());
```
vrutkovs: Which username and password are sent when `force_auth` is true?

steveej: The ones which are set, which could be none.

vrutkovs: In that case why can't `None` be sent right away, trap 401 and try without username/password?

steveej: I'll try to fill in your question:
In the case of username and password being `None`, we try to fetch the releases without authentication first, and if it tails with a 401 error code we try to get a token anonymously and then retry the release fetching.
I think that can be done. I have just kept it simpler and left the decision for anonymous authentication to the configuration.

steveej: To drive this conversation towards resolve, @vrutkovs are you asking for a change in the above described manner or were you merely curious?

vrutkovs (changes requested):
anonymous_auth is not necessary, it can be derived based username/password and registry responses

steveej:
I want to land some dkregistry-rs PRs first which are now linked in the first comment.

/hold

steveej:
The dkregistry-rs PRs have landed and this should be good to go when CI is green.

/hold

steveej:
I think this timing issue could be caused by high load on the CI node:

```
 thread 'plugins::tests::process_blocking_succeeds' panicked at 'took 2.139688319s despite timeout of 2s', cincinnati/src/plugins/mod.rs:675:9 
```

/retest

vrutkovs (approved):
/lgtm

PR #323, opened 2020-10-23
bump dkregistry for 1.43.1 compatbility; dist/Dockerfile.build: also install 1.43.1

steveej:
Required because we use Rust 1.43.1 in the release pipeline.

/cc @LalatenduMohanty

### DIFF
--- a/dist/Dockerfile.build/Dockerfile
+++ b/dist/Dockerfile.build/Dockerfile
@@ -11,7 +11,8 @@ ENV HOME="/root"
 ENV PATH="${HOME}/.cargo/bin:${PATH}"
 
 # build: Rust stable toolchain
-RUN curl https://sh.rustup.rs -sSf | sh -s -- --default-toolchain 1.46.0 -y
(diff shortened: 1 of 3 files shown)


dist/Dockerfile.build/Dockerfile
```diff
@@ -11,7 +11,8 @@ ENV HOME="/root"
 ENV PATH="${HOME}/.cargo/bin:${PATH}"
 
 # build: Rust stable toolchain
-RUN curl https://sh.rustup.rs -sSf | sh -s -- --default-toolchain 1.46.0 -y
+RUN curl https://sh.rustup.rs -sSf | sh -s -- --default-toolchain 1.46.0 -y && \
+  rustup install 1.43.1
```
LalatenduMohanty, line 15: Why do we need to install rust 1.43.1 explicitly? Just to maintain backward compatibility?

steveej: Yes.

LalatenduMohanty (approved):
/lgtm

PR #277, opened 2020-05-08
[Do not merge] Cargo lockfile V2 format.Generated  using cargo generate-lockfile

LalatenduMohanty:
https://github.com/rust-lang/cargo/pull/7070

Testing if it works in our CI

Signed-off-by: Lalatendu Mohanty <lmohanty@redhat.com>

PR #324, opened 2020-10-23
Cargo.lock: update to new format

steveej:
This is a syntax only update that was accomplished using
`cargo generate-lockfile --offline`.


steveej:
I think it's easier to hold this back until we fully switch the switch to Rust 1.47.

/hold

LalatenduMohanty:
@steveeJ this is similar to https://github.com/openshift/cincinnati/pull/277/, isn't it?

steveej:
> this is similar to #277, isn't it?

It is indeed. I'm fine closing either or both. WDYT?

LalatenduMohanty:
> It is indeed. I'm fine closing either or both. WDYT?

I am willing to update #277 . Not sure why you suggested to wait till Rust 1.47

steveej:
> Not sure why you suggested to wait till Rust 1.47

When I updated to that version it automatically updated the lockfile to the new format, that's what initiated the PR. I have since pinned Rust to 1.46 in the cincinnati repo locally to workaround that issue. I'm fine closing this and letting you update #277

PR #329, opened 2020-10-28
dist/Dockerfiles: switch from docker.io to quay.io

steveej:
This is done because the app-sre pipeline has recently introduced
blocking connections to docker.io to prevent hitting the ratelimit.


wking (approved):
/lgtm

LalatenduMohanty (approved):
/lgtm

PR #330, opened 2020-10-28
[release-4.6]: catch up with master

steveej:


steveej:
Closing because this will fail once #333 is merged because it lacks the vendored files and can't add them due to the nature of the PR.

PR #333, opened 2020-10-28
dist/vendor: gate git commands with an env var

steveej:
This is so that CI can most easily inspect whether running the script
caused a change or not.

/cc @wking

### DIFF
--- a/dist/vendor.sh
+++ b/dist/vendor.sh
@@ -1,4 +1,4 @@
-#!/usr/bin/env sh
+#!/usr/bin/env bash
 
 set -ueo pipefail
 
--- a/dist/vendor.sh
+++ b/dist/vendor.sh
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-set -ueo pipefail
+set -ue
 
 CARGO_CONFIG=".cargo/config.toml"
 FILES_TO_COMMIT=(
@@ -29,7 +29,7 @@ find ./vendor -type f -exec chmod a+r {} \;
 # check the a variable as we don't want to commit new files on CI
 # it can also be set by developers which want to skip the git operations
 if [ "${OPENSHIFT_CI}" != "true" ]; then
-  git add --force "${FILES_TO_COMMIT[@]}"
+  git add "${FILES_TO_COMMIT[@]}"
   git diff --quiet --staged -- "${FILES_TO_COMMIT[@]}" || {
     echo Files changed, committing...
     git commit -m "Update vendored dependencies and rh-manifest.txt" -- "${FILES_TO_COMMIT[@]}"
--- a/dist/vendor.sh
+++ b/dist/vendor.sh
@@ -26,8 +26,12 @@ cargo vendor > "${CARGO_CONFIG}"
 # some vendored files aren't world readable which has lead to issues in container builds when trying out the workflow
 find ./vendor -type f -exec chmod a+r {} \;
 
-git add --force "${FILES_TO_COMMIT[@]}"
-git diff --quiet --staged -- "${FILES_TO_COMMIT[@]}" || {
-  echo Files changed, committing...
-  git commit -m "Update vendored dependencies and rh-manifest.txt" -- "${FILES_TO_COMMIT[@]}"
-}
+# check the a variable as we don't want to commit new files on CI
+# it can also be set by developers which want to skip the git operations
+if [ "${OPENSHIFT_CI}" != "true" ]; then
+  git add --force "${FILES_TO_COMMIT[@]}"
+  git diff --quiet --staged -- "${FILES_TO_COMMIT[@]}" || {
+    echo Files changed, committing...
+    git commit -m "Update vendored dependencies and rh-manifest.txt" -- "${FILES_TO_COMMIT[@]}"
+  }
+fi


dist/vendor.sh
```diff
@@ -26,8 +26,10 @@ cargo vendor > "${CARGO_CONFIG}"
 # some vendored files aren't world readable which has lead to issues in container builds when trying out the workflow
 find ./vendor -type f -exec chmod a+r {} \;
 
-git add --force "${FILES_TO_COMMIT[@]}"
-git diff --quiet --staged -- "${FILES_TO_COMMIT[@]}" || {
-  echo Files changed, committing...
-  git commit -m "Update vendored dependencies and rh-manifest.txt" -- "${FILES_TO_COMMIT[@]}"
-}
+if [ "${COMMIT}" = "1" ]; then
```
LalatenduMohanty: I do not see `COMMIT` defined anywhere.

steveej: that's correct. it's expected to be passed in the environment to make the script commit. would you prefer having `COMMIT=0` somewhere?

LalatenduMohanty: Can you give some information about how the environment is going call the script and how COMMIT is going to be defined? May be a comment should be ok too.

LalatenduMohanty: My intent is to just have required information so that when someone want to reuse this script she has the required information.

dist/vendor.sh
```diff
@@ -26,8 +26,10 @@ cargo vendor > "${CARGO_CONFIG}"
 # some vendored files aren't world readable which has lead to issues in container builds when trying out the workflow
 find ./vendor -type f -exec chmod a+r {} \;
 
-git add --force "${FILES_TO_COMMIT[@]}"
```
vrutkovs, line 29: CI would run the job with [OPENSHIFT_CI=true env var](https://docs.ci.openshift.org/docs/architecture/step-registry/#available-environment-variables), so if its set we should exit with 0

vrutkovs (changes requested):

dist/vendor.sh
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env sh
+#!/usr/bin/env bash
```
LalatenduMohanty, line 1: I am fine with `bash` but wondering what made you to switch?

steveej: I usually use neovim for editing shell scripts but now I happened to open it in vscode with ShellCheck integration enabled. So I just followed the warnings given by it.

LalatenduMohanty (approved):
/lgtm

dist/vendor.sh
wking: `set -o pipefail` is a Bashism (it's not [in POSIX][1]).  That said, I see no pipes in this script today, so we could probably go with:

```sh
#!/usr/bin/env sh

set -eu
```

and see what vscode had to say about it ;).

[1]: https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#set

vrutkovs: no need to introduce bash dependency

dist/vendor.sh
```diff
@@ -26,8 +26,12 @@ cargo vendor > "${CARGO_CONFIG}"
...
 find ./vendor -type f -exec chmod a+r {} \;
 
-git add --force "${FILES_TO_COMMIT[@]}"
-git diff --quiet --staged -- "${FILES_TO_COMMIT[@]}" || {
-  echo Files changed, committing...
-  git commit -m "Update vendored dependencies and rh-manifest.txt" -- "${FILES_TO_COMMIT[@]}"
-}
+# check the a variable as we don't want to commit new files on CI
+# it can also be set by developers which want to skip the git operations
+if [ "${OPENSHIFT_CI}" != "true" ]; then
```
vrutkovs, line 31:
```suggestion
if [ "${OPENSHIFT_CI}" != "true" ]; then exit 0; fi
```

so that extra nesting would be avoided

dist/vendor.sh
```diff
@@ -26,8 +26,12 @@ cargo vendor > "${CARGO_CONFIG}"
...
+  git add --force "${FILES_TO_COMMIT[@]}"
```
vrutkovs: Why `--force` btw?

dist/vendor.sh
steveej: > no need to introduce bash dependency

You can argue with shellcheck about this:

```console
$ shellcheck dist/vendor.sh

In dist/vendor.sh line 6:
FILES_TO_COMMIT=(
                ^-- SC2039: In POSIX sh, arrays are undefined.


In dist/vendor.sh line 15:
  mv --backup=existing ${CARGO_CONFIG}{,.prev}
                                      ^------^ SC2039: In POSIX sh, brace expansion is undefined.


In dist/vendor.sh line 32:
  git add --force "${FILES_TO_COMMIT[@]}"
                   ^-------------------^ SC2039: In POSIX sh, array references are undefined.


In dist/vendor.sh line 33:
  git diff --quiet --staged -- "${FILES_TO_COMMIT[@]}" || {
                                ^-------------------^ SC2039: In POSIX sh, array references are undefined.


In dist/vendor.sh line 35:
    git commit -m "Update vendored dependencies and rh-manifest.txt" -- "${FILES_TO_COMMIT[@]}"
                                                                         ^-------------------^ SC2039: In POSIX sh, array references are undefined.

For more information:
  https://www.shellcheck.net/wiki/SC2039 -- In POSIX sh, array references are...
```

These warnings all go away if I change the shebang to refernece `bash`.

vrutkovs: These are issues to fix, not a requirement to have bash

dist/vendor.sh
steveej: Nesting helps readability by making it obvious that the nested code is conditional. I regard it coincidental that this is code is at the end of the file. If the script is ever extended with more code after this condition your suggestion would require changing it just for the sake of being able to add more code.

dist/vendor.sh
steveej: I was bitten by a local gitignore file at one point with this script so this is a guarantee it doesn't happen to anyone else.

vrutkovs: What if I have a file with credentials saved there? The script tries to be unnecessarily smart. Let remove `git add/commit` part - I'd rather do it manually tbh

steveej: Since we have another safety net in the build pipeline, which is it will simply fail if these files aren't up to date, I'm fine removing this one.

dist/vendor.sh
steveej: They are features of bash and other shells and were only displayed by shellcheck because of the `sh` shebang. So the fix is simply to switch to bash, since I think depending on bash is a very reasonable requirement. How do you justify the restriction your suggesting?

LalatenduMohanty: I can not think of any downside of switching to bash. Also I am fine with bash as it will reduce the effort we need to put to fix issues if we want to stick to `sh`

dist/vendor.sh
LalatenduMohanty: @vrutkovs you mean `if [ "${OPENSHIFT_CI}" = "true" ]; then exit 0; fi` right? you suggesting we should exit if `"${OPENSHIFT_CI}"` is set to true?

steveej: @LalatenduMohanty 

> @vrutkovs you mean `if [ "${OPENSHIFT_CI}" = "true" ]; then exit 0; fi` right? you suggesting we should exit if `"${OPENSHIFT_CI}"` is set to true?

Note that your correcting the details of the suggestion is merely a distraction from the discussion I'm trying to have, which is about maintainability.

LalatenduMohanty: Personally I will also prefer `if [ "${OPENSHIFT_CI}" = "true" ]; then exit 0; fi` as this is more easy to understand. But I this is not an issue which should stop this PR from merging.

LalatenduMohanty: Because it is more of a programming style.

LalatenduMohanty: > Note that your correcting the details of the suggestion is merely a distraction from the discussion I'm trying to have, which is about maintainability.

I do not think maintainability is an issue with either styles if I specifically think about this bash script.

steveej:
> The following test **failed**, say `/retest` to rerun all failed tests:

Probably an SLO flake:

```
 test check_slo::_max_over_time_cincinnati_gb_graph_upstream_errors_total_1h_0_ ... FAILED 
```

/retest

dist/vendor.sh
wking: Can we just drop `FILES_TO_COMMIT`?  It is an attempt to allow the vendor script to run in the face of an otherwise dirty working directory without committing the dirt, but it's not clear to me that folks would want to do that (and `git stash ...` provides a fairly convenient workaround if they do have dirt they don't want committed).  If we want to keep it and stick with POSIX, we could use `set` and `${@}` to maintain this array.

PR #344, opened 2020-10-30
cincinnati/src/lib: Fix "Abstrat" -> "Abstract" typos

wking:
These snuck in with b77e137dbd (#312).

/assign @steveeJ


wking:
In case this is a cased-username or sleepy-Prow thing:

/override ci/prow/e2e

PR #346, opened 2020-10-31
cincinnati/src/lib: Fix diff edge sense (correcting left vs. right)

wking:
The mixed inputs I'm correcting landed with the methor in 84faa290fc (#250).

/assign @steveeJ


steveej (approved):
/lgtm

PR #349, opened 2020-12-10
WIP Dockerfiles: improve dockerfiles to be compatible with rust-toolset images

vrutkovs:
Rework dockerfiles so that we could use `registry.redhat.io/rhel8/rust-toolset` as a builder image. This ensures workdir is set once in our builder image, root user set explicitly and `jq` fetched as a binary.

TODO:
* [ ] No `yamllint` in UBI8 images

Precondition for https://github.com/openshift/release/pull/13990


vrutkovs:
No longer necessary

PR #354, opened 2021-01-06
Bug 1913442: Dockerfile.e2e: add rustfmt and yaml lint scripts

vrutkovs:
These scripts should be part of e2e image so that the test could use prebuilt e2e image instead of `src`


LalatenduMohanty (approved):
/lgtm

vrutkovs:
single SLO test failed, seem to be a flake

/retest

PR #366, opened 2021-01-25
Bug 1916188: Dockerfile: remove extra copies

vrutkovs:
This is no longer required, as builder already contains those


wking (approved):
/lgtm

PR #367, opened 2021-01-28
Bug 1916188: Dockerfile.deploy: copy sourcecode

vrutkovs:
CI tests don't require the source code to be copied here, but without it 
stage deploy procedure is broken

### DIFF
--- a/dist/Dockerfile.deploy/Dockerfile
+++ b/dist/Dockerfile.deploy/Dockerfile
@@ -1,5 +1,6 @@
 FROM quay.io/app-sre/cincinnati:builder AS builder
 
+COPY . .
 # copy git information for built crate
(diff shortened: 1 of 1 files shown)


dist/Dockerfile.deploy/Dockerfile
```diff
@@ -1,5 +1,6 @@
 FROM quay.io/app-sre/cincinnati:builder AS builder
 
+COPY . .
```
LalatenduMohanty, line 3: We do not need to set the workdir ?

vrutkovs: Its already set in quay.io image and CI's replacement, no need to override it

LalatenduMohanty (approved):
/lgtm

PR #370, opened 2021-02-16
No error on multiple tags referencing the same pullspec

PratikMahajan:
Do not throw the error
```
Concrete release with the same version (4.1.30+amd64) already exists
```

As quay tags are manifested by the hash, we can safely ignore
multiple tags referencing the same pullspec.

### DIFF
--- a/cincinnati/src/lib.rs
+++ b/cincinnati/src/lib.rs
@@ -73,6 +73,18 @@ impl Release {
             Release::Concrete(release) => Some(&mut release.metadata),
         }
     }
+
+    /// Returns the `manifestref` of a given `Release`
+    pub fn manifestref(&self) -> Result<&String, Error> {
+        let digestkey = String::from("io.openshift.upgrades.graph.release.manifestref");
+        match self {
+            Release::Concrete(release) => {
+                let digest = release.metadata.get(&digestkey);
+                Ok(digest.map(|d| d).unwrap())
+            }
+            _ => bail!("could not get manifest reference"),
+        }
+    }
 }
 
 /// Type to represent a Release with all its information.
@@ -188,10 +200,15 @@ impl Graph {
             Some(id) => {
                 let node = self.dag.node_weight_mut(id.0).expect(EXPECT_NODE_WEIGHT);
                 if let Release::Concrete(_) = node {
-                    bail!(
-                        "Concrete release with the same version ({}) already exists",
-                        release.version()
-                    );
+                    // check if release digest and node digest are same
+                    if release.manifestref().unwrap() != node.manifestref().unwrap() {
+                        bail!(
+                            "mismatched manifest ref for concrete release {}: {}, {}",
+                            release.version(),
+                            release.manifestref().unwrap(),
+                            node.manifestref().unwrap()
+                        )
+                    }
                 }
                 *node = release;
                 Ok(id)
--- a/cincinnati/src/plugins/internal/graph_builder/release.rs
+++ b/cincinnati/src/plugins/internal/graph_builder/release.rs
@@ -162,6 +162,7 @@ pub fn create_graph(releases: Vec<Release>) -> Result<cincinnati::Graph, Error> 
 #[cfg(test)]
 mod tests {
     use super::*;
+    use std::collections::BTreeMap;
 
     #[test]
     fn create_graph_tolerates_nonexistent_edges() -> Fallible<()> {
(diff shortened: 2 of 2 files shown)


cincinnati/src/lib.rs
```diff
@@ -187,12 +187,6 @@ impl Graph {
         match self.find_by_version(&release.version()) {
             Some(id) => {
                 let node = self.dag.node_weight_mut(id.0).expect(EXPECT_NODE_WEIGHT);
-                if let Release::Concrete(_) = node {
```
vrutkovs, line 190: We should ensure we don't add `Release::Abstract` here

vrutkovs (commented):
This looks good to me. Lets add some unit tests to verify that?

cincinnati/src/lib.rs
```diff
@@ -188,10 +201,11 @@ impl Graph {
             Some(id) => {
                 let node = self.dag.node_weight_mut(id.0).expect(EXPECT_NODE_WEIGHT);
                 if let Release::Concrete(_) = node {
-                    bail!(
-                        "Concrete release with the same version ({}) already exists",
-                        release.version()
-                    );
+                    // check if release digest and node digest are same
+                    if release.manifestref().unwrap() != node.manifestref().unwrap(){
```
wking: Doesn't this independently ingest both references?  E.g.:

1. Cincinnati sees reference A for digest D.
2. Cincinnati pulls the manifest and all layers from digest D to construct a concrete release A' for digest D.
3. Cincinnati inserts A' in the graph.
4. Cincinnati sees reference B for digest D.
5. Cincinnati pulls the manifest and all layers from digest D to construct a concrete release B' for digest D.
6. Cincinnati sees that the version and manifest references for A' and B' match, so... B' does not get inserted in the graph?

I'd rather avoid 5 entirely.  Hmm... I wonder if we have logging in CI that shows images being ingested...

cincinnati/src/lib.rs
vrutkovs: Covered in `manifestref()` func

vrutkovs:
/lgtm
/hold

for Trevor to review, IUUC https://github.com/openshift/cincinnati/pull/370#discussion_r581441848 is addressed in `manifestref()`

cincinnati/src/lib.rs
PratikMahajan: B’ will get inserted into the graph.
Before:
1. release comes to add_release
2. cincinnati sees a concrete release already exists.
3. throws error

Now:
1. release comes to add_release
2. cincinnati sees that a concrete release already exists
3. checks if the sha of previous and new match.
4. if yes, adds the release, if no, throws error

I’ve not implemented the mapping function in add_release as we are already getting a constructed release when when it comes to the function.
Logical change only includes, checking if the new release that we are getting is not conflicting with an existing release. in case of multiple releases with same SHA, the ones which we encounter the latest is present in the graph.

wking: > I’ve not implemented the mapping function in add_release as we are already getting a constructed release when when it comes to the function.

Right, that's the redundant-fetch business I'm concerned about ;).  Poking around, looks like we're [calling `add_release` from `create_graph`][1].  We're calling `create_graph` with [data pulled from `fetch_releases`][2].  `fetch_releases` gets manifests over [here][3].  So I'd rather have our guard living [before `tags.try_for_each_concurrent`][4] or thereabouts where we drop any tags whose manifest references are redundant.  Confirming we should theoretically have that information available at the tag-listing stage:

```console
$ curl -s 'https://quay.io/api/v1/repository/openshift-release-dev/ocp-release/tag/?page=3' | jq -r '.tags[] | select(.expiration == null) | .manifest_digest + " " + .name' | sort | head -n2
sha256:08b06c446c8388c0ca3b32e6ed9d43e835a20fea72e078579cde2b2304b284d2 4.4.32-s390x
sha256:0975188e83f8688f97180b408a447b41f492ee35d1dacd43a826b14db7d486e5 4.6.9-ppc64le
```

Although I confess to not knowing exactly how to get at it in the `get_tags` result.

[1]: https://github.com/openshift/cincinnati/blob/9a55fde2c8a7746d1b8e99d72e3ce5daa7aba837/cincinnati/src/plugins/internal/graph_builder/release.rs#L89
[2]: https://github.com/openshift/cincinnati/blob/9a55fde2c8a7746d1b8e99d72e3ce5daa7aba837/cincinnati/src/plugins/internal/graph_builder/release_scrape_dockerv2/plugin.rs#L141-L151
[3]: https://github.com/openshift/cincinnati/blob/9a55fde2c8a7746d1b8e99d72e3ce5daa7aba837/cincinnati/src/plugins/internal/graph_builder/release_scrape_dockerv2/registry/mod.rs#L269-L271
[4]: https://github.com/openshift/cincinnati/blob/9a55fde2c8a7746d1b8e99d72e3ce5daa7aba837/cincinnati/src/plugins/internal/graph_builder/release_scrape_dockerv2/registry/mod.rs#L262

PratikMahajan: Unfortunately, we can only get the image tag names and not the manifest ref. 
dkregistry crate uses, `GET /v2/<name>/tags/list?n=<integer>`

see: https://docs.docker.com/registry/spec/api/#listing-image-tags
https://github.com/camallo/dkregistry-rs/blob/master/src/v2/tags.rs#L26

wking: So:

```console
$ curl -s https://quay.io/v2/openshift-release-dev/ocp-release/tags/list | jq -r '.tags[]' | head -n2
4.0.0-2
4.0.0-3
$ curl -i https://quay.io/v2/openshift-release-dev/ocp-release/manifests/4.0.0-2
HTTP/2 200 
server: nginx/1.12.1
date: Fri, 19 Mar 2021 03:45:16 GMT
content-type: application/vnd.docker.distribution.manifest.v1+prettyjws
content-length: 2868
docker-content-digest: sha256:324e9e586ed110e2d63ca5edc86d6ddf15ffd46fa53fe17a59aa8948f8a9b8ff
x-frame-options: DENY
strict-transport-security: max-age=63072000; preload

{
   "schemaVersion": 1,
   "name": "openshift-release-dev/ocp-release",
...
```

dkregistry [captures that header][1] (and probably does more intelligent media type content negotiation).  And then [after `get_manifest_and_ref`][2], we perform a lookup in `cache` or in a local locked map or some such to check if the slot is already populated, or to stake our claim that we are about to attempt to populate it with a release (so that we don't have parallel workers check the cache and find it empty while another worker is busy pulling layers)?

[1]: https://github.com/camallo/dkregistry-rs/blob/762ef80e1200966b44290f6b4e0ac6af052c0ff6/src/v2/manifest/mod.rs#L56
[2]: https://github.com/openshift/cincinnati/blob/9a55fde2c8a7746d1b8e99d72e3ce5daa7aba837/cincinnati/src/plugins/internal/graph_builder/release_scrape_dockerv2/registry/mod.rs#L272

cincinnati/src/lib.rs
```diff
@@ -188,10 +200,15 @@ impl Graph {
...
                 let node = self.dag.node_weight_mut(id.0).expect(EXPECT_NODE_WEIGHT);
                 if let Release::Concrete(_) = node {
-                    bail!(
-                        "Concrete release with the same version ({}) already exists",
-                        release.version()
-                    );
+                    // check if release digest and node digest are same
+                    if release.manifestref().unwrap() != node.manifestref().unwrap() {
+                        bail!(
+                            "mismatched manifest ref for concrete release {}:{}, {}",
```
wking: nit: missing a space after the colon?  `mismatched manifest ref for concrete release {}: {}, {}`

cincinnati/src/lib.rs
```diff
@@ -188,10 +200,15 @@ impl Graph {
...
+                            release.version(),
+                            release.manifestref().unwrap(),
+                            node.manifestref().unwrap()
+                        )
+                    }
                 }
                 *node = release;
```
wking, line 213: I don't understand what this line does.  If we get to this point, don't we know that we already have this node in the graph?  Can we just return `Ok(id)` without... whatever this line is up to?

PratikMahajan: That will be handled while handling the cache issue

PR #374, opened 2021-03-03
add tests to check liveness and readiness

PratikMahajan:


graph-builder/src/main.rs
```diff
@@ -157,15 +157,16 @@ mod tests {
     use commons::metrics::HasRegistry;
     use commons::metrics::RegistryWrapper;
     use commons::testing;
+    use graph_builder::status::{serve_liveness, serve_readiness};
     use parking_lot::RwLock;
     use prometheus::Registry;
     use std::collections::HashSet;
     use std::sync::Arc;
 
     fn mock_state() -> State {
```
vrutkovs: let split this into
* `mock_state` - default setting, liveness/readiness returns 503
* `mock_state_live` - `is_live` is set to true, liveness is 200, readiness is 503
* `mock_state_ready_live` - liveness/readiness return 200

vrutkovs (changes requested):

vrutkovs (approved):
/lgtm
/hold cancel

PR #381, opened 2021-03-23
reduce redundant pulling of layer digest

PratikMahajan:
In a scenario when a worker is actively pulling the layer
digest and another worker checks the cache for the same release,
it resulted in a cache miss and redundant downloading of the
layer digest.
this commit fixes the issue with the use of a placeholder which
takes place of the release till all the layer digests are downloaded.
This will not result in a cache miss thus stopping the redundant
pulling of layers.
The placeholder in the cache will be replaced with updated values
once the layer digest is pulled.


vrutkovs (approved):
/lgtm

PR #383, opened 2021-03-25
e2e ubi8: copy cargo_test.sh

vrutkovs:
In UBI8 setup we have to re-use e2e container to run unit tests, so the 
image has to have all the necessary scripts


LalatenduMohanty (approved):
/lgtm

PR #384, opened 2021-03-29
Update CI registry path

vrutkovs:
`registry.svc.ci` was sunset, use `registry.ci` instead.

This also marks three network tests as ignored, as OCP 4.x registry acts differently on unauthenticated requests.


PratikMahajan (approved):
/lgtm

PR #385, opened 2021-03-31
extend policy-engine error logging

PratikMahajan:
Cincinnati should now log incorrect API requests.
earlier incorrect requests were returned with error
but the error was only logged in case the user-provided incorrect
parameters, other errors were not logged.
this PR makes sure we are catching left-out error requests.

Also, added a default response in case someone ends up hitting an
endpoint that does not exist. Also logging these errors.

### DIFF
--- a/policy-engine/src/graph.rs
+++ b/policy-engine/src/graph.rs
@@ -1,6 +1,7 @@
 //! Cincinnati graph service.
 
 use crate::AppState;
+use actix_web::http::HeaderValue;
 use actix_web::web::Query;
 use actix_web::{HttpRequest, HttpResponse};
 use cincinnati::plugins::BoxedPlugin;
@@ -40,6 +41,15 @@ pub(crate) async fn index(
     req: HttpRequest,
     app_data: actix_web::web::Data<AppState>,
 ) -> Result<HttpResponse, GraphError> {
+    _index(&req, app_data)
+        .await
+        .map_err(|e| api_response_error(&req, e))
+}
+
+async fn _index(
+    req: &HttpRequest,
+    app_data: actix_web::web::Data<AppState>,
+) -> Result<HttpResponse, GraphError> {
     let span = get_tracer().start("index", None);
 
     V1_GRAPH_INCOMING_REQS.inc();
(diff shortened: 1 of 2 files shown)


policy-engine/src/main.rs
```diff
@@ -129,6 +131,18 @@ fn main() -> Result<(), Error> {
     Ok(())
 }
 
+// log errors in case an incorrect endpoint is called
+fn default_response(req: HttpRequest) -> HttpResponse {
+    error!(
+        "Error serving request '{}' from '{}': Incorrect Endpoint",
+        format!("{:?}", req).replace("\n", " ").replace("\t", " "),
+        req.peer_addr()
```
wking, line 139: I dunno if we care about [`peer_addr`][1], since that will almost always be the router proxy, right?  I expect we want to include the forwarder chain, or assume the forwarder chain doesn't contain spoofed input and use [`realip_remote_addr`][2]?

[1]: https://docs.rs/actix-web/3.3.2/actix_web/struct.HttpRequest.html#method.peer_addr
[2]: https://docs.rs/actix-web/3.3.2/actix_web/dev/struct.ConnectionInfo.html#method.realip_remote_addr

wking: Never mind, I see `peer_addr` is being recycled from #186.  If we want to adjust that, we can punt to follow-up work.

policy-engine/src/graph.rs
```diff
@@ -45,37 +45,65 @@ pub(crate) async fn index(
...
     // Check that the client can accept JSON media type.
-    commons::ensure_content_type(req.headers(), CONTENT_TYPE)?;
+    commons::ensure_content_type(req.headers(), CONTENT_TYPE)
+        .map_err(|e| api_response_error(&req, e))?;
 
     // Check for required client parameters.
     let mandatory_params = &app_data.mandatory_params;
-    commons::ensure_query_params(mandatory_params, req.query_string())?;
+    commons::ensure_query_params(mandatory_params, req.query_string())
+        .map_err(|e| api_response_error(&req, e))?;
```
wking: Instead of wrapping each possible error generator, should we just make an internal `_index(...)` (or some such, I dunno what Rust's conventions are for naming private functions) and then have:

```rust
pub(crate) async fn index(
    req: HttpRequest,
    app_data: actix_web::web::Data<AppState>,
) -> Result<HttpResponse, GraphError> {
    _index(req, app_data).map_err(|e| api_response_error(&reg, e))?
}
```

?  Otherwise, I'm worried that there will be errors we forget to wrap.  I guess #185 is the furthest up we can bubble things, and I'm agnostic on whether we want to do something at the `App::new().wrap(...)` level or the `index` level, but I'm pretty sure we don't want to have multiple `map_err` within one function, unless only a subset of the functions errors deserve the `api_response_error` mapping.

wking (approved):
/lgtm

PR #388, opened 2021-04-14
improve e2e experience 

PratikMahajan:
makes sure that e2e works seamlessly on local machine.
also, added flag `-- --nocapture` to print the logs for tests.

### DIFF
--- a/Justfile
+++ b/Justfile
@@ -47,7 +47,8 @@ test: format
 _test component cargoargs='--features test-net,test-net-private' rustcargs='--ignored':
 	#!/usr/bin/env bash
 	set -xe
-	(pushd {{component}} && cargo test {{cargoargs}} -- {{rustcargs}})
+	export RUST_BACKTRACE=1 RUST_LOG="graph-builder=trace,cincinnati=trace,dkregistry=trace"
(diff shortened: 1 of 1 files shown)


Justfile
```diff
@@ -206,12 +203,9 @@ run-daemons-e2e:
 	just \
 		registry="{{registry}}" repository="{{repository}}" \
 		run-graph-builder-e2e 2>&1 &
-	PG_PID=$!
 
 	just run-policy-engine 2>&1 &
-	PE_PID=$!
 
-	trap "kill $PG_PID $PE_PID" EXIT
 	sleep infinity
```
wking: It feels strange to me drop drop this kill but still keep the infinite `sleep`.  Don't you want killing the `just run-daemons-e2e` call to clean up the daemons it created?

Justfile
```diff
@@ -79,12 +80,8 @@ run-e2e:
 	#!/usr/bin/env bash
 	set -e
 
-	just \
```
vrutkovs, line 82: Not sure what this change does

PratikMahajan: reverted the change

vrutkovs (approved):
/lgtm
/hold cancel

PR #391, opened 2021-04-16
Fix up available-updates.sh script

cblecker:
Shoving the JSON data into a variable was causing escaping/syntax errors with JQ when it was echoed back.

This swaps the script to generate and use a tempfile to pass the data to JQ. Retains error checking, and traps cleanup of the file.


PratikMahajan (approved):
/lgtm

PR #416, opened 2021-04-27
Remove dependency on dockerfile.build

PratikMahajan:
we are doing a multi-stage build in dockerfile.deploy to
remove the dependency on dockerfile.build

this is because we are installing certain modules in dockerfile.build
that are introducing vulnerabilities. We are not carrying these
vulnerabilities in the final container that we are shipping, we are just
using dockerfile.build container as a builder image. This can be achieved
in a multi-stage build, thus not requiring a separate build dockerfile.

### DIFF
--- a/dist/Dockerfile.deploy/Dockerfile
+++ b/dist/Dockerfile.deploy/Dockerfile
@@ -1,4 +1,27 @@
-FROM quay.io/app-sre/cincinnati:builder AS builder
+FROM registry.centos.org/centos/centos:7 as builder
+
+# base: EPEL repo for extra tools
+RUN yum -y install epel-release
+
+# build: system utilities and libraries
+RUN yum update -y && \
+    yum -y groupinstall 'Development Tools' && \
+    yum -y install gcc openssl-devel protobuf-compiler jq skopeo buildah libgit2 && \
+    yum -y install yamllint && \
+    yum -y install cmake elfutils-libelf-devel libcurl-devel binutils-devel elfutils-devel && \
+    yum clean all
+
+ENV HOME="/root"
+ENV PATH="${HOME}/.cargo/bin:${PATH}"
+
+# build: Rust stable toolchain
+RUN curl https://sh.rustup.rs -sSf | sh -s -- --default-toolchain 1.50.0 -y && \
+  rustup install 1.46.0
+
+RUN \
+  mkdir -p $HOME/.cargo/git/ && \
+  find $HOME/. -type d -exec chmod 777 {} \; && \
+  find $HOME/. -type f -exec chmod ugo+rw {} \;
 
 COPY . .
 # copy git information for built crate
--- a/dist/Dockerfile.e2e/Dockerfile
+++ b/dist/Dockerfile.e2e/Dockerfile
@@ -1,4 +1,29 @@
-FROM quay.io/app-sre/cincinnati:builder AS rust_builder
+FROM registry.centos.org/centos/centos:7 as rust_builder
+
+# base: EPEL repo for extra tools
+RUN yum -y install epel-release
+
+# build: system utilities and libraries
+RUN yum update -y && \
+    yum -y groupinstall 'Development Tools' && \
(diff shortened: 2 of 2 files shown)


dist/Dockerfile.e2e/Dockerfile
```diff
@@ -1,4 +1,31 @@
-FROM quay.io/app-sre/cincinnati:builder AS rust_builder
+FROM registry.centos.org/centos/centos:8 as rust_builder
+
+# base: EPEL repo for extra tools
+RUN yum -y install epel-release
+
+# build: system utilities and libraries
+RUN yum update -y && \
+    yum -y install gcc openssl-devel jq skopeo buildah && \
```
vrutkovs: `/usr/bin/cincinnati-e2e-test: error while loading shared libraries: libssl.so.1.1: cannot open shared object file: No such file or directory`

I think we want `openssl`, not `-devel` here

PratikMahajan: updated

vrutkovs: Wrong container, these are installed in `builder`, not in resulting ubi8

LalatenduMohanty:
@PratikMahajan Are you planing to remove the rpms coming from epel from https://github.com/openshift/cincinnati/blob/master/dist/Dockerfile.build/Dockerfile  after this PR merges?

vrutkovs (approved):
/lgtm

vrutkovs:
/hold

Not sure why is it failing to start UWM. Lets see if a retest after 4 hrs helps

vrutkovs:
/lgtm cancel

Now its a valid failure - seems `customrust` deploy image doesn't have `git2` installed, so git version is incorrectly being reported

dist/Dockerfile.e2e/Dockerfile
```diff
@@ -24,14 +52,14 @@ RUN yum update -y && \
 COPY --from=rust_builder /opt/cincinnati/bin/e2e /usr/bin/cincinnati-e2e-test
 COPY --from=rust_builder /opt/cincinnati/bin/prometheus_query /usr/bin/cincinnati-prometheus_query-test
 COPY --from=rust_builder /opt/cincinnati/bin/slo /usr/bin/cincinnati-e2e-slo
-COPY --from=rust_builder /go/src/github.com/openshift/cincinnati/hack/e2e.sh hack/
```
vrutkovs: Previous `RUN` needs `libgit2` to fix e2e and report valid commit message

dist/Dockerfile.deploy/Dockerfile
```diff
@@ -1,4 +1,27 @@
-FROM quay.io/app-sre/cincinnati:builder AS builder
+FROM registry.centos.org/centos/centos:7 as builder
+
+# base: EPEL repo for extra tools
+RUN yum -y install epel-release
+
+# build: system utilities and libraries
+RUN yum update -y && \
+    yum -y groupinstall 'Development Tools' && \
+    yum -y install gcc openssl-devel protobuf-compiler jq skopeo buildah && \
```
vrutkovs:
```suggestion
    yum -y install gcc openssl-devel protobuf-compiler jq skopeo buildah libgit2 && \
```

dist/Dockerfile.e2e/Dockerfile
vrutkovs: I was wrong - `libgit2` must be present in build container, not necessary in e2e/deploy images

dist/Dockerfile.e2e/Dockerfile
```diff
@@ -18,20 +46,20 @@ ENV PATH="${PATH}:${HOME}/bin"
 
 # Install container tools
 RUN yum update -y && \
-    yum install -y skopeo buildah && \
+    yum install -y skopeo buildah libgit2 && \
```
vrutkovs, line 49:
```suggestion
    yum install -y skopeo buildah && \
```

vrutkovs:
That looks great! Lets squash the commits and I'll lgtm it

/hold cancel

vrutkovs (approved):
/lgtm

PR #419, opened 2021-05-03
hack/e2e.sh: drop tech preview UWM

vrutkovs:
User Workload Monitoring in 4.5 required tech preview flag to be set. e2e tests moved to 4.7, so this flag is no longer needed.

Follow-up for https://github.com/openshift/release/pull/16424


wking (approved):
/lgtm

PR #437, opened 2021-05-21
replace rust-toolset image with UBI

PratikMahajan:
Now that rust-toolset image is deprecated, we need to replace it
with UBI image.
We can now install rust directly in UBI image


wking (approved):
/lgtm

PR #456, opened 2021-06-09
update opentelemetry to v0.14.0

PratikMahajan:
update the `opentelemetry` dependency from v0.4.0 to v0.14.0
this change includes updating the `actix-web` dependency to 4.0.0-beta.5

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -26,9 +26,9 @@ dependencies = [
 
 [[package]]
 name = "actix-codec"
-version = "0.4.0-beta.1"
+version = "0.4.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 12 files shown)


vrutkovs:
That looks good, thanks!

I'll check if traces are still being set locally

vrutkovs:
Works with `podman run --rm -u0 --name jaeger -p6831:6831/udp -p16686:16686 jaegertracing/all-in-one:latest`

However two regressions are introduced:
* traces are disconnected. Each component is reporting these independently, instead of adding parent-child relationship.
Screenshot-20210622103031-1271x665

* policy-engine / graph-builder traces don't have necessary params:
Screenshot-20210622103321-862x371

See https://github.com/openshift/cincinnati/pull/266 for expected view

vrutkovs (changes requested):
see previous comment

vrutkovs (approved):
That looks great, thanks!
/lgtm

PR #463, opened 2021-06-16
[dockerfile][deploy][build] use /root as WORKDIR

PratikMahajan:
use `/root` as WORKDIR instead of current `/`.
this change is particularly needed as using `/` is the cause
for missing git information from `graph-builder` and
`policy-engine` binary.

### DIFF
--- a/dist/Dockerfile.deploy/Dockerfile
+++ b/dist/Dockerfile.deploy/Dockerfile
@@ -14,6 +14,8 @@ RUN yum update -y && \
 ENV HOME="/root"
 ENV PATH="${HOME}/.cargo/bin:${PATH}"
 
+WORKDIR ${HOME}
+
 # build: Rust stable toolchain
 RUN curl https://sh.rustup.rs -sSf | sh -s -- --default-toolchain 1.50.0 -y && \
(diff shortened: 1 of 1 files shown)


LalatenduMohanty (approved):
/lgtm

dist/Dockerfile.deploy/Dockerfile
```diff
@@ -23,14 +23,15 @@ RUN \
...
 COPY . .
 # copy git information for built crate
 COPY .git/ ./.git/
 
 RUN cargo build --release && \
     mkdir -p /opt/cincinnati/bin && \
-    cp -rvf target/release/graph-builder /opt/cincinnati/bin && \
-    cp -rvf target/release/policy-engine /opt/cincinnati/bin
+    cp -rvf /build/cincinnati/target/release/graph-builder /opt/cincinnati/bin && \
+    cp -rvf /build/cincinnati/target/release/policy-engine /opt/cincinnati/bin
```
wking: do we need this change?  I'd expect `WORKDIR` to be the current working directory for subsequent `RUN`

wking (approved):
/lgtm

PR #471, opened 2021-06-25
Add vendor files and rh-manifest.txt

DavidHurta:
Add files generated by the `dist/vendor.sh` script on the `release-4.8` branch.

Adds: 
- vendor files into the `vendor` directory
- `rh-manifest.txt`
- `.cargo/config.toml`

Suggested by https://github.com/openshift/cincinnati/pull/468#issuecomment-86785429.

Pull request related to https://issues.redhat.com/browse/OTA-372.

PR #470, opened 2021-06-25
Update release-4.7 branch

DavidHurta:
Update Cincinnati's release-4.7 branch.

Backport fixes from the master branch without the dependabot configuration.

Pull request related to https://issues.redhat.com/browse/OTA-372.

PR #473, opened 2021-06-28
Add vendor files and rh-manifest.txt

DavidHurta:
Add files generated by the `dist/vendor.sh` script on the `release-4.7` branch.

Adds:
- vendor files into the `vendor` directory
- `rh-manifest.txt`
- `.cargo/config.toml`

Pull request is similar to the https://github.com/openshift/cincinnati/pull/471, which added the vendor files to the `release-4.8` branch.

Pull request related to https://issues.redhat.com/browse/OTA-372


PratikMahajan (commented):
looks like the commit is same as #471. you can use `git cherry-pick -x <commit_hash>` to cherry pick commit and add it to this PR

DavidHurta:
I have used the `git rebase` to move the commits from the current `release-4.8` branch.

The reason why the [cincinnati: Remove dependabot configuration](https://github.com/openshift/cincinnati/commit/4e6239546e41985a6ab5adfcd41e7d30ae2a9e3d) commit is now present is that:
I should have used the `git cherry-pick` to move the commit in Pull request https://github.com/openshift/cincinnati/pull/470, but I have created a new commit, which then created a point in history where the branches diverged thus the checksums of the commit `Add vendor files and rh-manifest.txt` were different even though their changes were the same.

vrutkovs:
That lgtm, @PratikMahajan do we have any other concerns?

LalatenduMohanty:
> The reason why the cincinnati: Remove dependabot configuration commit is now present is that:
I should have used the git cherry-pick to move the commit in Pull request #470, but I have created a new commit, which then created a point in history where the branches diverged thus the checksums of the commit Add vendor files and rh-manifest.txt were different even though their changes were the same.

Yeah, two commits for removing the dependabot commits looks odd. But I am fine with it.

PR #475, opened 2021-06-30
Bug 1850781: [memory] define scope while running graph-builder to free up memory 

PratikMahajan:
we are creating the variable `internal_io` in an infinite loop
this variable although is decalared in this loop, never goes out
of scope, thus always staying in memory.

this commit defines scope in the infinite loop that runs every 5 minutes
variables that need not be in memory will be cleared once the scope ends

### DIFF
--- a/graph-builder/src/graph.rs
+++ b/graph-builder/src/graph.rs
@@ -181,6 +181,9 @@ pub fn run(settings: &config::AppSettings, state: &State) -> ! {
 
     BUILD_INFO.inc();
 
+    // Store amount of nodes in the graph for metrics
+    let mut nodes_count: i64;
+
     loop {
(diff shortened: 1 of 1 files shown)


vrutkovs (changes requested):
Instead of explicit drop perhaps we should wrap it:
```
let mut nodes_count: i64;
...
{
    let internal_io = match scrape {
    ...
    *state.json.write() = json_graph;
    nodes_count = internal_io.graph.releases_count() as i64;
}
...
GRAPH_FINAL_RELEASES.set(nodes_count);
debug!("graph update completed, {} valid releases", nodes_count);
```
https://github.com/vrutkovs/cincinnati/commit/43589f093b3da082b38d9ce5391d4cf5fe197c96

WDYT?

vrutkovs (approved):
/lgtm
/hold

for Lala to review

PR #476, opened 2021-07-01
docs/design/openshift: Document the 'arch' query parameter

wking:
Belatedly catching up with 1b50628e5b (#180).  Linking to the image-spec for valid values, and the image-spec in turn weakly references the Go architecture identifiers.


PratikMahajan (approved):
/lgtm

PR #485, opened 2021-07-07
Bug 1979719: handle container registry error gracefully

PratikMahajan:
in case of some error with container registry auth, warns with the error
and attempt anonymous authorization.
Also, includes the exact error so it's easier to debug the issue.

### DIFF
--- a/cincinnati/src/plugins/internal/graph_builder/release_scrape_dockerv2/plugin.rs
+++ b/cincinnati/src/plugins/internal/graph_builder/release_scrape_dockerv2/plugin.rs
@@ -113,12 +113,17 @@ impl ReleaseScrapeDockerv2Plugin {
             .context(format!("Parsing {} as Registry", &settings.registry))?;
 
         if let Some(credentials_path) = &settings.credentials_path {
-            let (username, password) =
-                registry::read_credentials(Some(&credentials_path), &registry.host_port_string())
-                    .context(format!(
-                    "Reading registry credentials from {:?}",
-                    credentials_path
-                ))?;
+            let (username, password) = registry::read_credentials(
+                Some(&credentials_path),
+                &registry.host_port_string(),
+            )
+            .unwrap_or_else(|err| {
+                warn!(
+                    "Error reading registry credentials from {:?}. Access to {:?} will be unauthenticated: {} ",
+                    credentials_path, &registry.host_port_string() ,err
+                );
+                return (None, None);
+            });
 
             settings.username = username;
             settings.password = password;


wking:
[e2e][1]:

```
error: timed out waiting for the condition on deploymentconfigs/cincinnati
...
9m43s       Normal    Started             pod/cincinnati-1-pg76s               Started container cincinnati-policy-engine
4m52s       Warning   Unhealthy           pod/cincinnati-1-pg76s               Readiness probe failed: HTTP probe failed with statuscode: 503
9m56s       Normal    Scheduled           pod/cincinnati-1-zq6f8               Successfully assigned cincinnati-e2e/cincinnati-1-zq6f8 to ip-10-0-226-191.us-west-1.compute.internal
...
9m42s       Normal    Started             pod/cincinnati-1-zq6f8               Started container cincinnati-policy-engine
9m9s        Warning   Unhealthy           pod/cincinnati-1-zq6f8               Readiness probe failed: HTTP probe failed with statuscode: 503
9m56s       Normal    SuccessfulCreate    replicationcontroller/cincinnati-1   Created pod: cincinnati-1-pg76s
```

Not sure if that is reproducible or not.

/retest

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cincinnati/485/pull-ci-openshift-cincinnati-master-e2e/1412877638099275776#1:build-log.txt%3A139

cincinnati/src/plugins/internal/graph_builder/release_scrape_dockerv2/plugin.rs
```diff
@@ -115,10 +115,10 @@ impl ReleaseScrapeDockerv2Plugin {
         if let Some(credentials_path) = &settings.credentials_path {
             let (username, password) =
                 registry::read_credentials(Some(&credentials_path), &registry.host_port_string())
-                    .context(format!(
-                    "Reading registry credentials from {:?}",
-                    credentials_path
-                ))?;
+                    .unwrap_or_else(|err| {
+                        warn!("Error Reading registry credentials from {:?}, Setting credentials to 'None': {} ", credentials_path, err);
```
wking: nit: `Error Reading` -> `Error reading`.  Or just leave it at `Reading`, because the fact that it's a `warn` and there is an attached error message should be enough to let folks know there was a problem.

And also probably `, Setting` to either `, setting` or `. Setting`.  Or maybe `Access to {:?} will be unauthenticated` with `&registry.host_port_string()` added to the formatting arguments, or some such?

vrutkovs:
>in case of some error with container registry auth, warns with the error
and attempt anonymous authorization.

if I understood correctly this is implemented in #313 and it has logging setup around it. I don't quite understand why would we tolerate invalid setting instead of throwing an error there.

cincinnati/src/plugins/internal/graph_builder/release_scrape_dockerv2/plugin.rs
PratikMahajan: Updated. Kept the string `Error` to keep it obvious.

PratikMahajan:
> > in case of some error with container registry auth, warns with the error
> > and attempt anonymous authorization.
> 
> if I understood correctly this is implemented in #313 and it has logging setup around it. I don't quite understand why would we tolerate invalid setting instead of throwing an error there.

@vrutkovs 
Currently, we are setting [default config path for Cincinnati through the operator](https://github.com/openshift/cincinnati-operator/blob/1a746871611d4ccb07aa7599c9eac0ef21df56d1/controllers/new.go#L56). 
In case someone wanted to access the repo without using auth, they need not provide the creds path. But given that we are hard coding it in the operator, we need to handle it either in Cincinnati or give a way for users through Operator to unset it. 
We decided to make Cincinnati more tolerant.

vrutkovs:
Okay, that sgtm

/lgtm

PR #499, opened 2021-08-02
hack: use valid data for dummy secret

vrutkovs:
`oc create secret` no longer allows empty `--from-literal`. This change adds a dummy value for it to pass.

This PR also prints commands executed so that we'd know where it breaks next time

### DIFF
--- a/hack/deploy_cincinnati.sh
+++ b/hack/deploy_cincinnati.sh
@@ -12,7 +12,7 @@ export IMAGE_TAG=90efacd
 oc new-project cincinnati
 
 # Create a dummy secret as a workaround to not having real secrets
-oc create secret generic cincinnati-credentials --from-literal=""
+oc create secret generic cincinnati-credentials --from-literal="foo=bar"
 
 
 # Apply oc template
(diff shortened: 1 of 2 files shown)


LalatenduMohanty:
> `oc create secret` no longer allows empty `--from-literal`

Assuming OC from 4.9 has the change right? as we install the latest `dist/Dockerfile.e2e/Dockerfile`

```
# Get oc CLI
 40 RUN mkdir -p ${HOME}/bin && \
 41     curl https://mirror.openshift.com/pub/openshift-v4/clients/oc/latest/linux/oc.tar.gz 2>/dev/null | tar xzf - -C "${HOME}/bin/" oc
```

vrutkovs:
>Assuming OC from 4.9 has the change right?

Its probably a 4.8 GA change. Not sure if its useful to stick to a particular `oc` major/minor version.

LalatenduMohanty:
> Its probably a 4.8 GA change. Not sure if its useful to stick to a particular `oc` major/minor version.

I am fine with what we have today.

LalatenduMohanty (approved):
/lgtm

PR #528, opened 2021-09-10
Update Rust to 1.56/1.52 and use UBI in rust-toolset

vrutkovs:
Dedicated Rust UBI image was deprecated, we should be installing Rust for UBI

### DIFF
--- a/dist/Dockerfile.deploy/Dockerfile
+++ b/dist/Dockerfile.deploy/Dockerfile
@@ -17,8 +17,8 @@ ENV PATH="${HOME}/.cargo/bin:${PATH}"
 WORKDIR ${HOME}
 
 # build: Rust stable toolchain
-RUN curl https://sh.rustup.rs -sSf | sh -s -- --default-toolchain 1.53.0 -y && \
-  rustup install 1.49.0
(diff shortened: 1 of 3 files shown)


hack/e2e.sh
```diff
@@ -74,7 +74,7 @@ backoff oc -n openshift-monitoring \
 backoff oc apply -f /tmp/cluster-monitoring-config.yaml
 
 # Wait for user workload monitoring is deployed
-backoff oc -n openshift-user-workload-monitoring wait --for=condition=Ready pod -l app=thanos-ruler
+#backoff oc -n openshift-user-workload-monitoring wait --for=condition=Ready pod -l app=thanos-ruler
```
LalatenduMohanty: It is not clear why we need this.

vrutkovs: Investigating why UWM verification doesn't work as expected. I'll move this to a separate PR

PratikMahajan (approved):
/lgtm

vrutkovs:
/hold

Still no luck passing e2e

PratikMahajan (approved):
/lgtm

PR #531, opened 2021-09-13
e2e: don't wait for UWM to start

vrutkovs:
Skip waiting for UWM to come up as this command is flaking a lot. This is a core platform bug and should be reported instead of being retried.

Instead we'll notice failures in SLO tests

### DIFF
--- a/hack/e2e.sh
+++ b/hack/e2e.sh
@@ -73,9 +73,6 @@ backoff oc -n openshift-monitoring \
         --from-literal='config.yaml={"enableUserWorkload": true}' -o yaml --dry-run=client > /tmp/cluster-monitoring-config.yaml
 backoff oc apply -f /tmp/cluster-monitoring-config.yaml
 
-# Wait for user workload monitoring is deployed
-backoff oc -n openshift-user-workload-monitoring wait --for=condition=Ready pod -l app=thanos-ruler
-
 # Import observability template
 # ServiceMonitors are imported before app deployment to give Prometheus time to catch up with
 # metrics


hack/e2e.sh
```diff
@@ -73,9 +73,6 @@ backoff oc -n openshift-monitoring \
         --from-literal='config.yaml={"enableUserWorkload": true}' -o yaml --dry-run=client > /tmp/cluster-monitoring-config.yaml
 backoff oc apply -f /tmp/cluster-monitoring-config.yaml
 
-# Wait for user workload monitoring is deployed
-backoff oc -n openshift-user-workload-monitoring wait --for=condition=Ready pod -l app=thanos-ruler
```
wking, line 77: is there an example of what this flake looks like?  Because [the official docs][1] include a step:

> Check that the prometheus-operator, prometheus-user-workload and thanos-ruler-user-workload pods are running in the openshift-user-workload-monitoring project. It might take a short while for the pods to start...

and we probably want to follow that flow here.  Or tie into a workflow that enables user-workload monitoring for us (and performs the recommended wait?)?

[1]: https://docs.openshift.com/container-platform/4.8/monitoring/enabling-monitoring-for-user-defined-projects.html

vrutkovs: >is there an example of what this flake looks like

https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cincinnati/529/pull-ci-openshift-cincinnati-master-customrust-e2e/1437399103607672832 fails on this step, but all UWM parts are running correctly by `gather-extra` step time

>we probably want to follow that flow here

I don't think we should be duplicating UWM verification in cincinnati e2e. If its broken QE would have noticed it, as we would simply retest it instead of report it as a bug.

vrutkovs:
No longer required

PR #544, opened 2021-10-01
change the prometheus `endpoint` label to `uri_path` due to conflict

PratikMahajan:
there is a prometheus label with `endpoint`.
Before this commit, series labels looked like:
```
cincinnati_pe_graph_incoming_requests_total{container="cincinnati-policy-engine",endpoint="status-pe",exported_endpoint="/api/upgrades_info/graph",instance="10.130.2.191:9081",job="cincinnati-policy-engine",namespace="cincinnati-stage",pod="cincinnati-248-9k69d",service="cincinnati-policy-engine"}
```
this commit will change the `exported_endpoint` label to `uri_path`


wking (approved):
/lgtm

PR #545, opened 2021-10-05
Version the Cincinnati Graph

PratikMahajan:
This PR versions the cincinnati graph and embeds
the version information into cincinnati graph json.

This commit includes changes that will let us accept
proactive negotiation using ACCEPT header value.
eg. `ACCEPT: application/vnd.redhat.cincinnati.v1+json`

The resultant graph should look like this 
```json
{
"edges":[...],
"nodes": [...],
"version":1
}
```

### DIFF
--- a/cincinnati/src/lib.rs
+++ b/cincinnati/src/lib.rs
@@ -751,6 +751,7 @@ impl From<Graph> for plugins::interface::Graph {
 #[cfg(any(test, feature = "test"))]
 pub mod testing {
     use super::*;
+    use crate::plugins::internal::versioned_graph::VersionedGraph;
 
     pub fn generate_graph() -> Graph {
         let mut graph = Graph::default();
@@ -925,6 +926,22 @@ pub mod testing {
         pub payload_replace_sha_by_tag_left: bool,
         pub payload_replace_sha_by_tag_right: bool,
         pub payload_remove_registry_and_repo: bool,
+    }
+
+    /// Compares two Versioned Graphs and gives a verbose error if not equal
+    pub fn compare_versioned_graphs_verbose(
+        versioned_left: VersionedGraph,
+        versioned_right: VersionedGraph,
+        settings: CompareGraphsVerboseSettings,
+    ) -> Fallible<()> {
+        if versioned_left.version != versioned_right.version {
+            bail!(
+                "graph version mismatch left: {}, right {}",
+                versioned_left.version,
+                versioned_right.version
+            )
+        }
+        return compare_graphs_verbose(versioned_left.graph, versioned_right.graph, settings);
     }
 
     /// Compares two Graphs and gives a verbose error if not equal.
--- a/cincinnati/src/lib.rs
+++ b/cincinnati/src/lib.rs
@@ -43,8 +43,7 @@ pub use std::collections::BTreeMap as MapImpl;
 pub use std::collections::BTreeSet as SetImpl;
 
 /// Graph type which stores `Release` as node-weights and `Empty` as edge-weights.
-#[derive(Debug, Default)]
-#[cfg_attr(any(test, feature = "test"), derive(Clone))]
+#[derive(Debug, Default, Clone)]
 pub struct Graph {
     dag: Dag<Release, Empty>,
 }
--- a/cincinnati/src/plugins/internal/versioned_graph.rs
+++ b/cincinnati/src/plugins/internal/versioned_graph.rs
@@ -1,0 +1,32 @@
+use self::cincinnati::plugins::prelude_plugin_impl::*;
+use crate as cincinnati;
+use commons::{CINCINNATI_VERSION, MIN_CINCINNATI_VERSION};
+
+#[derive(Debug, Serialize, Deserialize, SmartDefault)]
+#[serde(default)]
+pub struct VersionedGraph {
+    version: i32,
+    #[serde(flatten)]
+    graph: cincinnati::Graph,
+}
+
+impl VersionedGraph {
+    pub const PLUGIN_NAME: &'static str = "versioned-graph";
+
+    pub fn versioned_graph(io: &InternalIO) -> Fallible<VersionedGraph> {
+        let min_version = match CINCINNATI_VERSION.get(*MIN_CINCINNATI_VERSION) {
+            Some(version) => *version,
+            None => bail!("error parsing minimum cincinnati version"),
+        };
+        Ok(VersionedGraph {
+            version: match io.parameters.get("content_type") {
+                None => min_version,
+                Some(v) => match CINCINNATI_VERSION.get(v.as_str()) {
+                    Some(version) => *version,
+                    None => min_version,
+                },
+            },
+            graph: io.graph.clone(),
+        })
+    }
+}
--- a/cincinnati/src/plugins/internal/versioned_graph.rs
+++ b/cincinnati/src/plugins/internal/versioned_graph.rs
@@ -5,9 +5,9 @@ use commons::{CINCINNATI_VERSION, MIN_CINCINNATI_VERSION};
 #[derive(Debug, Serialize, Deserialize, SmartDefault)]
 #[serde(default)]
 pub struct VersionedGraph {
-    version: i32,
+    pub version: i32,
     #[serde(flatten)]
-    graph: cincinnati::Graph,
+    pub graph: cincinnati::Graph,
 }
 
 impl VersionedGraph {
@@ -28,5 +28,86 @@ impl VersionedGraph {
             },
             graph: io.graph.clone(),
         })
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use crate as cincinnati;
+    use std::collections::HashMap;
+
+    use super::*;
+    use cincinnati::testing::{generate_custom_graph, TestMetadata};
+    use commons::testing::init_runtime;
+    use commons::MIN_CINCINNATI_VERSION;
+
+    fn get_min_version() -> Fallible<i32> {
+        match CINCINNATI_VERSION.get(*MIN_CINCINNATI_VERSION) {
+            Some(version) => Ok(*version),
+            None => bail!("error parsing minimum cincinnati version"),
+        }
+    }
+
+    #[test]
+    fn min_version_if_missing() -> Fallible<()> {
+        let _ = init_runtime()?;
+
+        let input_graph: cincinnati::Graph = {
+            let metadata: TestMetadata = vec![(1, [].iter().cloned().collect())];
+            generate_custom_graph("image", metadata, None)
+        };
+
+        let versioned_graph = VersionedGraph::versioned_graph(&InternalIO {
+            graph: input_graph,
+            parameters: Default::default(),
+        })
+        .unwrap();
+
+        assert_eq!(versioned_graph.version, get_min_version().unwrap());
+        Ok(())
+    }
+
+    #[test]
+    fn ensure_min_on_unsupported() -> Fallible<()> {
+        let _ = init_runtime()?;
+
+        let input_graph: cincinnati::Graph = {
+            let metadata: TestMetadata = vec![(1, [].iter().cloned().collect())];
+            generate_custom_graph("image", metadata, None)
+        };
+
+        let mut plugin_params: HashMap<String, String> = HashMap::new();
+        plugin_params.insert(String::from("version"), "application/json".to_string());
+
+        let versioned_graph = VersionedGraph::versioned_graph(&InternalIO {
+            graph: input_graph,
+            parameters: plugin_params,
+        })
+        .unwrap();
+
+        assert_eq!(versioned_graph.version, get_min_version().unwrap());
+        Ok(())
+    }
+
+    #[test]
+    fn ensure_version_1() -> Fallible<()> {
+        let _ = init_runtime()?;
+
+        let input_graph: cincinnati::Graph = {
+            let metadata: TestMetadata = vec![(1, [].iter().cloned().collect())];
+            generate_custom_graph("image", metadata, None)
+        };
+
+        let mut plugin_params: HashMap<String, String> = HashMap::new();
+        plugin_params.insert(String::from("version"), MIN_CINCINNATI_VERSION.to_string());
+
+        let versioned_graph = VersionedGraph::versioned_graph(&InternalIO {
+            graph: input_graph,
+            parameters: plugin_params,
+        })
+        .unwrap();
+
+        assert_eq!(versioned_graph.version, 1);
+        Ok(())
     }
 }
--- a/commons/src/lib.rs
+++ b/commons/src/lib.rs
@@ -206,20 +206,93 @@ mod tests {
 
     #[test]
     fn test_validate_content_type() {
+        let most_recent_version = "application/vnd.redhat.cincinnati.v1+json";
+        let all_supported_versions: Vec<actix_web::http::HeaderValue> = CINCINNATI_VERSION
+            .keys()
+            .map(|val| header::HeaderValue::from_static(val))
+            .collect();
+
+        // Test for empty header
+        // No accept value provided with header, server accepts `application/json` and defaults to `application/json`
         let mut headers = actix_web::http::HeaderMap::new();
-        validate_content_type(&headers, "application/json").unwrap(); // if the request leaves Accept empty, we can return whatever we want
+        let accept_default = header::HeaderValue::from_str("application/json").unwrap();
+        let version = validate_content_type(
+            &headers,
+            vec![accept_default.clone()],
+            accept_default.clone(),
+        )
+        .unwrap(); // if the request leaves Accept empty, we return the minimum supported cincinnati version as that's the lowest version we support
+        assert_eq!(version, MIN_CINCINNATI_VERSION.to_string());
+
+        // Support old clients with older cincinnati config.
+        // `application/json` provided with header, server accepts `application/json` and defaults to `application/json`
         headers.insert(
             header::ACCEPT,
             //"application/json, text/*; q=0.2".parse().unwrap(), // prefer JSON, but also accept any text/* after an 80% markdown in quality.  FIXME: needs a smarter parser in validate_content_type
             "application/json".parse().unwrap(),
         );
-        validate_content_type(&headers, "application/json").unwrap();
+        let version = validate_content_type(
+            &headers,
+            vec![accept_default.clone()],
+            accept_default.clone(),
+        )
+        .unwrap();
+        assert_eq!(version, "application/json");
+
+        // `application/*` provided with header, server accepts `application/json` and defaults to `application/json`
+        headers.insert(header::ACCEPT, "application/*".parse().unwrap());
+        let version = validate_content_type(
+            &headers,
+            vec![accept_default.clone()],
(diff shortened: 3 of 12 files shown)


commons/src/lib.rs
```diff
@@ -25,9 +25,21 @@ pub mod prelude_errors {
...
 
+lazy_static! {
+    /// list of cincinnati versions
+    pub static ref CINCI_VERSION: HashMap<&'static str, i32> =
+        [("application/vnd.redhat.cincinnati.v1+json", 1)]
+            .iter()
+            .cloned()
+            .collect();
+    /// minimum cincinnati version supported
+    pub static ref MIN_CINCI_VERSION: i32 = 1;
```
wking: nit: are we already scoped to Cincinnati by our package?  I am not clear on why we need `CINCI` here.  And while I like the Cincy abbreviation internally, I'd rather spell out `Cincinnati` in formal public APIs instead of abbreviating, if it turns out that package scoping isn't enough to allow `MIN_VERSION` here.

commons/src/lib.rs
```diff
@@ -104,6 +116,10 @@ pub fn validate_content_type(
     let full_type = header::HeaderValue::from_static(content_type);
     let wildcard = header::HeaderValue::from_static("*");
     let double_wildcard = header::HeaderValue::from_static("*/*");
+    let mut cinci_versions: Vec<actix_web::http::HeaderValue> = CINCI_VERSION
```
wking: nit: this local variable definitely doesn't need `cinci_`, because any readers will know they're looking at Cincinnati code.

commons/src/lib.rs
```diff
@@ -104,6 +116,10 @@ pub fn validate_content_type(
     let full_type = header::HeaderValue::from_static(content_type);
```
wking: We should make up our mind about whether `validate_content_type` accepts supported types from the caller (which is how `content_type` is currently working) or via an internal global (which is how you're setting up the `CINCI_VERSION` consumption).  Personally I'd rather have callers pass this in explicitly, but to do that they'd need to build a vector with `application/json` and the set of supported `application/vnd.redhat.cincinnati.v1+json`, `application/vnd.redhat.cincinnati.v2+json`, etc.  So I'm not strongly against porting to using the global and dropping the `content_type` argument here.  But using both pathways at once seems like it would be too confusing.

commons/src/lib.rs
```diff
@@ -127,6 +144,26 @@ pub fn validate_content_type(
     }
 }
 
+/// get the requested version from accept header
+pub fn get_version(headers: &HeaderMap) -> i32 {
```
wking: It might be worth dropping `validate_content_type` and replacing it with a fallible `get_version`.  Because I don't see much use-case for a separate `validate_content_type` call.  Won't all the folks who call `validate_content_type` today call `get_version` right after?

policy-engine/src/graph.rs
```diff
@@ -144,9 +147,30 @@ where
...
         .map_err(|e| GraphError::FailedJsonOut(e.to_string()))?;
 
+    let v =
+        serde_json::from_str(&graph_json).map_err(|e| GraphError::FailedJsonOut(e.to_string()))?;
+    let version: i32 = match internal_io.parameters.get("version") {
+        None => *commons::MIN_CINCI_VERSION as i32,
+        Some(v) => v.parse::<i32>().unwrap(),
+    };
+
+    let updated_json = add_version_information(&v, &version);
```
wking: This feels like it's cheating a bit and adding a hard-coded plugin to the tail of the plugin chain.  On the one hand, we have a number of hard-coded policy-engine configs in places like cincinnati-operator where this is deployed.  On the other, we can bump all those locations to configure a new version endpoint.  I'd rather make this an official plugin, since it's consuming information from `plugin_params`.

policy-engine/src/graph.rs
```diff
@@ -144,9 +147,30 @@ where
...
+
     Ok(HttpResponse::Ok()
         .content_type(CONTENT_TYPE)
```
wking: This should be consuming the version from `get_version` (possibly by looking in `plugin_params`).  So `get_version` might need to return both a version integer and a recommended media type string, so we can satisfy folks who ask for v1 with `application/json` and folks who ask for that same content via `application/vnd.redhat.cincinnati.v1+json`.

commons/src/lib.rs
PratikMahajan: Agree, i tried to go with passing the vector at first, but later decided on internal global. I'll dig more into this.

commons/src/lib.rs
```diff
@@ -25,9 +25,21 @@ pub mod prelude_errors {
...
 
 use actix_web::http::{header, HeaderMap};
+use std::collections::HashMap;
 use std::collections::HashSet;
 use url::form_urlencoded;
 
+lazy_static! {
+    /// list of cincinnati versions
+    pub static ref CINCINNATI_VERSION: HashMap<&'static str, i32> =
+        [("application/vnd.redhat.cincinnati.v1+json", 1)]
```
vrutkovs, line 35: Extract it dynamically using regexp?

commons/src/lib.rs
```diff
@@ -25,9 +25,21 @@ pub mod prelude_errors {
...
+            .iter()
+            .cloned()
+            .collect();
+    /// minimum cincinnati version supported
+    pub static ref MIN_CINCINNATI_VERSION: &'static str = "application/vnd.redhat.cincinnati.v1+json";
```
vrutkovs, line 40: Perhaps we want `default` here instead? `min` should be decided from `CINCINNATI_VERSION` hashmap?

vrutkovs (changes requested):

commons/src/lib.rs
```diff
@@ -172,19 +196,44 @@ mod tests {
...
+            // FIXME: drop once validate_content_type gets a smarter parser and the previous insert can include the text/* entry
+            header::ACCEPT,
+            "application/vnd.redhat.cincinnati.v1+json".parse().unwrap(),
+        );
+        let cincinnati_v1: Vec<actix_web::http::HeaderValue> =
+            vec![
+                header::HeaderValue::from_str("application/vnd.redhat.cincinnati.v1+json").unwrap(),
+            ];
+        let version = validate_content_type(&headers, cincinnati_v1).unwrap(); // if the request leaves Accept empty, we can return whatever we want
+        assert_eq!(version, 1);
```
vrutkovs:
```suggestion
        assert_eq!(version, "1");
```
Comparing strings to strings

cincinnati/src/lib.rs
```diff
@@ -43,7 +43,7 @@ pub use std::collections::BTreeMap as MapImpl;
 pub use std::collections::BTreeSet as SetImpl;
 
 /// Graph type which stores `Release` as node-weights and `Empty` as edge-weights.
-#[derive(Debug, Default)]
+#[derive(Debug, Default, Clone)]
 #[cfg_attr(any(test, feature = "test"), derive(Clone))]
```
vrutkovs: needs to be removed so that `test` features would not attempt to re-implement Clone

Same for InternalIO:
```
#[derive(Debug, Clone)]
#[cfg_attr(test, derive(PartialEq))]
pub struct InternalIO {
    pub graph: cincinnati::Graph,
    pub parameters: HashMap<String, String>,
}
```

cincinnati/src/plugins/internal/json_append.rs
```diff
@@ -0,0 +1,119 @@
+use crate as cincinnati;
+// use serde::de::{self, Deserialize, Deserializer, MapAccess, Visitor};
+// use serde::ser::{Serialize, SerializeStruct, Serializer};
```
wking: nit: we should drop any commented-out code before landing this.

commons/src/lib.rs
```diff
@@ -172,19 +196,44 @@ mod tests {
...
             // FIXME: drop once validate_content_type gets a smarter parser and the previous insert can include the text/* entry
             header::ACCEPT,
             "text/*".parse().unwrap(),
         );
-        validate_content_type(&headers, "text/plain").unwrap();
-        validate_content_type(&headers, "image/png").unwrap_err();
+        let text_type: Vec<actix_web::http::HeaderValue> =
+            vec![header::HeaderValue::from_str("text/plain").unwrap()];
+        let version = validate_content_type(&headers, text_type).unwrap();
+        assert_eq!(version, "text/*");
```
wking: This is "server supports `text/plain`; client requests `text/*`", and should be returning "we're planning on responding with `text/plain`", not `text/*`.

commons/src/lib.rs
```diff
@@ -172,19 +196,44 @@ mod tests {
...
+        validate_content_type(&headers, image_type).unwrap_err();
+
+        headers.insert(
+            // FIXME: drop once validate_content_type gets a smarter parser and the previous insert can include the text/* entry
+            header::ACCEPT,
+            "application/vnd.redhat.cincinnati.v1+json".parse().unwrap(),
+        );
+        let cincinnati_v1: Vec<actix_web::http::HeaderValue> =
+            vec![
+                header::HeaderValue::from_str("application/vnd.redhat.cincinnati.v1+json").unwrap(),
```
wking: can we also accept `application/json` here?  Because in production, the server will support both `application/json` (for backwards-compat with existing CVOs) and this versioned type, and we'll want to ensure that we respond with `application/json` to those older CVOs and `application/vnd.redhat.cincinnati.v1+json` to newer CVOs out of the same server config.

graph-builder/src/graph.rs
```diff
@@ -108,8 +109,11 @@ pub async fn index(
     let path = req.uri().path();
     GRAPH_INCOMING_REQS.with_label_values(&[path]).inc();
 
+    let generic_header: Vec<actix_web::http::HeaderValue> =
+        vec![header::HeaderValue::from_static(CONTENT_TYPE)];
+
     // Check that the client can accept JSON media type.
-    commons::validate_content_type(req.headers(), CONTENT_TYPE)?;
+    let _version: String = commons::validate_content_type(req.headers(), generic_header)?;
```
wking: `content_type` is probably a better name for this variable, since it's holding `application/json` or other supported media type string.  And then we should pass this through to the `.content_type(...)` method being called a few lines down.

policy-engine/src/graph.rs
```diff
@@ -60,17 +61,30 @@ async fn _index(
...
 
+    let mut generic_header: Vec<actix_web::http::HeaderValue> =
+        vec![header::HeaderValue::from_static(CONTENT_TYPE)];
+
+    let mut version_headers: Vec<actix_web::http::HeaderValue> = commons::CINCINNATI_VERSION
+        .keys()
+        .map(|val| header::HeaderValue::from_static(val))
+        .collect();
+
+    version_headers.append(&mut generic_header);
```
wking: do we need to declare the `generic_header` variable?  Instead of inlining here with:

```rust
version_headers.append(header::HeaderValue::from_static(CONTENT_TYPE))
```

policy-engine/src/graph.rs
```diff
@@ -60,17 +61,30 @@ async fn _index(
...
     // Check for required client parameters.
     let mandatory_params = &app_data.mandatory_params;
     commons::ensure_query_params(mandatory_params, req.query_string())?;
 
-    let plugin_params = Query::<HashMap<String, String>>::from_query(req.query_string())
+    let mut plugin_params = Query::<HashMap<String, String>>::from_query(req.query_string())
         .map(|query| query.into_inner())
         .map_err(|e| commons::GraphError::InvalidParams(e.to_string()))?;
 
+    plugin_params.insert(String::from("version"), cincinnati_version);
```
wking: you're injecting the media type here, not the version integer, right?  Probably call this `content_type` or some such (or convert it to the version integer/string, without the rest of the `application/vnd.redhat.cincinnati...` context).

cincinnati/src/plugins/internal/json_append.rs
```diff
@@ -0,0 +1,119 @@
...
+#[derive(Debug, Serialize, Deserialize, SmartDefault)]
+#[serde(default)]
+pub struct VersionedGraph {
+    version: i32,
+    #[serde(flatten)]
+    graph: cincinnati::Graph,
+}
+
+impl VersionedGraph {
+    pub const PLUGIN_NAME: &'static str = "json-version-append";
```
vrutkovs: `<bikeshedding>`graph-version maybe?`</bikeshedding>`
I imagine in the future this plugin would also convert existing graph into previous versions maybe?

commons/src/lib.rs
wking: ok, `validate_content_type(&headers, cincinnati_v1, accept_default.clone())` below now, and the default type gets injected into the allowed list, so I'm good here.

cincinnati/src/plugins/internal/versioned_graph.rs
```diff
@@ -0,0 +1,116 @@
...
+        let input_graph: cincinnati::Graph = {
+            let metadata: TestMetadata = vec![(1, [].iter().cloned().collect())];
+            generate_custom_graph("image", metadata, None)
+        };
+
+        let mut plugin_params: HashMap<String, String> = HashMap::new();
+        plugin_params.insert(
+            String::from("version"),
+            "application/vnd.redhat.cincinnati.v1+json".to_string(),
+        );
```
vrutkovs:
```suggestion
        plugin_params.insert(String::from("version"), MIN_CINCINNATI_VERSION.to_string());
```

cincinnati/src/plugins/internal/versioned_graph.rs
```diff
@@ -0,0 +1,116 @@
...
+        })
+        .unwrap();
+
+        assert_eq!(versioned_graph.version, get_min_version().unwrap());
+        Ok(())
+    }
+
+    #[test]
+    fn ensure_version_1() -> Fallible<()> {
+        let runtime = init_runtime()?;
```
vrutkovs:
```suggestion
        let _ = init_runtime()?;
```
`runtime` is not used here

cincinnati/src/plugins/internal/versioned_graph.rs
```diff
@@ -0,0 +1,116 @@
...
+        })
+        .unwrap();
+
+        assert_eq!(versioned_graph.version, get_min_version().unwrap());
+        Ok(())
+    }
+
+    #[test]
+    fn ensure_min_on_unsupported() -> Fallible<()> {
+        let runtime = init_runtime()?;
```
vrutkovs: Same, unused `runtime`

cincinnati/src/plugins/internal/versioned_graph.rs
```diff
@@ -0,0 +1,116 @@
...
+    fn get_min_version() -> Fallible<i32> {
+        match CINCINNATI_VERSION.get(*MIN_CINCINNATI_VERSION) {
+            Some(version) => Ok(*version),
+            None => bail!("error parsing minimum cincinnati version"),
+        }
+    }
+
+    #[test]
+    fn min_version_if_missing() -> Fallible<()> {
+        let runtime = init_runtime()?;
```
vrutkovs:
```suggestion
        let _ = init_runtime()?;
```

commons/src/lib.rs
vrutkovs: We can improve this later on

commons/src/lib.rs
```diff
@@ -94,34 +106,59 @@ pub fn ensure_query_params(
...
+            .any(|i| i == "*")
+        {
+            Ok(header::HeaderValue::to_str(&accept_default)
+                .unwrap()
+                .parse()
+                .unwrap())
+        } else {
+            let minimum_version: String = MIN_CINCINNATI_VERSION.to_string();
+            let accept = header::HeaderValue::to_str(header_value);
+            match accept {
```
vrutkovs:
```suggestion
            match header::HeaderValue::to_str(header_value) {
```
No need for additional `let`

commons/src/lib.rs
```diff
@@ -94,34 +106,59 @@ pub fn ensure_query_params(
...
+                Ok(a) => Ok(a.parse().unwrap()),
+                Err(_e) => Ok(minimum_version),
```
vrutkovs:
```suggestion
                Err(_e) => Ok(MIN_CINCINNATI_VERSION.to_string()),
```
same, extra `let` is not necessary

commons/src/lib.rs
```diff
@@ -171,20 +208,83 @@ mod tests {
...
+        //server should throw error on non-supported ACCEPT
+        validate_content_type(&headers, image_type, accept_default.clone()).unwrap_err();
+
+        //`application/vnd.redhat.cincinnati.v1+json` provided with header, server accepts
+        //`application/vnd.redhat.cincinnati.v1+json` and defaults to `application/json`
+        headers.insert(
+            header::ACCEPT,
+            "application/vnd.redhat.cincinnati.v1+json".parse().unwrap(),
+        );
+        let cincinnati_v1: Vec<actix_web::http::HeaderValue> =
```
vrutkovs, line 258:
```suggestion
            MIN_CINCINNATI_VERSION.to_string().parse().unwrap(),
```

commons/src/lib.rs
vrutkovs:
```suggestion
            MIN_CINCINNATI_VERSION.to_string().parse().unwrap(),
```
less hardcodes

commons/src/lib.rs
```diff
@@ -171,20 +208,83 @@ mod tests {
...
+            vec![
+                header::HeaderValue::from_str("application/vnd.redhat.cincinnati.v1+json").unwrap(),
+            ];
+        let version =
+            validate_content_type(&headers, cincinnati_v1.clone(), accept_default.clone()).unwrap();
+        // Server returns the response with content_type `application/vnd.redhat.cincinnati.v1+json`
+        assert_eq!(version, "application/vnd.redhat.cincinnati.v1+json");
```
vrutkovs:
```suggestion
        assert_eq!(version, MIN_CINCINNATI_VERSION.to_string());
```

vrutkovs (changes requested):
Looks good, some code changes to reuse `commons` a bit more

commons/src/lib.rs
```diff
@@ -171,20 +208,83 @@ mod tests {
...
+        let image_type: Vec<actix_web::http::HeaderValue> =
+            vec![header::HeaderValue::from_str("image/png").unwrap()];
+        //server should throw error on non-supported ACCEPT
+        validate_content_type(&headers, image_type, accept_default.clone()).unwrap_err();
+
+        //`application/vnd.redhat.cincinnati.v1+json` provided with header, server accepts
+        //`application/vnd.redhat.cincinnati.v1+json` and defaults to `application/json`
+        headers.insert(
+            header::ACCEPT,
+            "application/vnd.redhat.cincinnati.v1+json".parse().unwrap(),
```
vrutkovs:
```suggestion
            MIN_CINCINNATI_VERSION.to_string().parse().unwrap(),
```

commons/src/lib.rs
PratikMahajan: updated the comments for tests to make it more clear and reduced hardcodes.

vrutkovs (approved):
LGTM, thanks!

PR #557, opened 2021-10-14
fix manifestref error if digest value missing

PratikMahajan:
Currently if a ConcreteRelease is missing the manifestref digest,
Cincinnati tries to unwrap() this value which causes it to panic
becuase of the absense of digest.

This commit returns an invalid value for manifestref digest if
it is missing in the ConcreteRelease.
This will cause a mismatch in fn add_release() which will prevent
cincinnati from adding this release. Thus preventing the panic
and throwing an easily readable error.

### DIFF
--- a/cincinnati/src/lib.rs
+++ b/cincinnati/src/lib.rs
@@ -75,12 +75,16 @@ impl Release {
     }
 
     /// Returns the `manifestref` of a given `Release`
-    pub fn manifestref(&self) -> Result<&String, Error> {
+    pub fn manifestref(&self) -> Result<String, Error> {
         let digestkey = String::from("io.openshift.upgrades.graph.release.manifestref");
         match self {
             Release::Concrete(release) => {
                 let digest = release.metadata.get(&digestkey);
-                Ok(digest.map(|d| d).unwrap())
+                let manifestref = match digest.map(|d| d) {
+                    Some(manifestref) => manifestref.clone(),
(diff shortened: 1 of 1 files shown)


cincinnati/src/lib.rs
```diff
@@ -75,12 +75,16 @@ impl Release {
...
-    pub fn manifestref(&self) -> Result<&String, Error> {
+    pub fn manifestref(&self) -> Result<String, Error> {
         let digestkey = String::from("io.openshift.upgrades.graph.release.manifestref");
         match self {
             Release::Concrete(release) => {
                 let digest = release.metadata.get(&digestkey);
-                Ok(digest.map(|d| d).unwrap())
+                let manifestref = match digest.map(|d| d) {
+                    Some(manifestref) => manifestref.clone(),
+                    None => String::from("no manifestref for release"),
```
lucab, line 85: If all the logic expects a valid `manifestref` but some `ConcreteRelease` do not have one, it means that either 1) the design/assumption is invalid, or 2) the input was invalid and it has not been properly validated.
Creating a fake arbitrary string here won't do any good to the situation.

Is a release without a manifestref a valid one? If yes, then this method should reflect that through an `Option<_>` result. If not, then the input parsing logic should be enhanced to detect and highlight the invalid input.

Getting rid of this `unwrap` would be overall good, but likely it should become an `ok_or_else(...)?`.

lucab (changes requested):

PR #559, opened 2021-10-18
OTA-498: Use OSUS operator to setup test Cincinnati

vrutkovs:
Instead of shipping our template and custom config, we should use OSUS operator to run Cincinnati.

TODO:
* [x] https://github.com/openshift/release/pull/22805 so that we could use 4.9 operator
* [x] https://github.com/openshift/release/pull/23242 so that graphdata image would always be fresh
* [x] https://github.com/openshift/release/pull/23954
* [x] Fix observability template / loadtesting
* [x] `r_min_over_time_up_job_e2e_graph_builder_1h_expects_1` frequently fails on rust or customrust, Removed this test case, as operator is watching the deployment is available


LalatenduMohanty:
Once we merge this, this is going to impact the SRE folks right?

vrutkovs:
>Once we merge this, this is going to impact the SRE folks right?

It shouldn't, as it only touches the e2e part not the deployement. IIRC Trevor has MRs to update stage/prod to use operator - these are separate tasks anyway

vrutkovs:
`test check_slo_numeric::_cincinnati_pe_http_upstream_errors_total_is_less_than_1_ ... FAILED`
Flake

/retest

LalatenduMohanty (approved):
/lgtm

PR #561, opened 2021-10-21
e2e: use Cincinnati config closer to production

vrutkovs:
Run with fetch_concurrency=16 and bump readiness/liveness/deployment timeouts, so that initial scrape would surely succeed on busy AWS account in CI


vrutkovs:
/retest

Infra problems

PratikMahajan (commented):
should we also reduce the fetch-concurrency to 16? that's what we use in production and stage.

LalatenduMohanty (approved):
/lgtm
/hold cancel

PR #574, opened 2021-11-12
Targeted edge blocking

PratikMahajan:
Add support for v2 of cincinnati and start supporting 
targeted edge blocking by adding conditional edges 
to cincinnati v2 response.


cincinnati/src/plugins/internal/graph_builder/openshift_secondary_metadata_parser/plugin.rs
```diff
@@ -478,6 +491,52 @@ impl OpenshiftSecondaryMetadataParserPlugin {
         Ok(())
     }
 
+    async fn process_conditional_edges(
+        &self,
+        graph: &mut cincinnati::Graph,
+        data_dir: &PathBuf,
+    ) -> Fallible<()> {
+        let blocked_edges_dir = data_dir.join(BLOCKED_EDGES_DIR);
```
vrutkovs, line 500: nitpick: lets split out the deserialization part in a function and reuse it in `process_blocked_edges`?

cincinnati/src/plugins/internal/graph_builder/openshift_secondary_metadata_parser/plugin.rs
```diff
@@ -478,6 +491,52 @@ impl OpenshiftSecondaryMetadataParserPlugin {
...
+        conditional_edges
+            .into_iter()
+            .try_for_each(|cey| -> Fallible<()> {
+                let ce: ConditionalEdge = ConditionalEdge {
+                    edge_regex: ConditionalUpdateEdge {
+                        from: cey.from.to_string(),
+                        to: cey.to.to_string(),
+                    },
+                    edges: vec![],
+                    risks: vec![ConditionalUpdateRisk {
```
vrutkovs, line 527: We should add a TODO here that some edges could be blocked on several risks

cincinnati/src/plugins/internal/edge_add_remove.rs
```diff
@@ -304,6 +312,83 @@ impl EdgeAddRemovePlugin {
...
+            .conditional_edges
+            .iter()
+            .try_for_each(|ce| -> Fallible<()> {
+                let mut conditional_edge = ConditionalEdge {
+                    edge_regex: ce.edge_regex.clone(),
+                    edges: vec![],
+                    risks: ce.risks.clone(),
+                };
+
+                let to_regex_string = ce.edge_regex.to.clone();
```
vrutkovs: nitpick: we don't need to clone it here, strictly speaking - but I'm fine with it as a convenience

cincinnati/src/plugins/internal/edge_add_remove.rs
```diff
@@ -304,6 +312,83 @@ impl EdgeAddRemovePlugin {
...
+
+                let to_regex = regex::Regex::new(&to_regex_string)
+                    .context(format!("Parsing {} as Regex", &to_regex_string))?;
+                let from_regex = regex::Regex::new(&from_regex_string)
+                    .context(format!("Parsing {} as Regex", &from_regex_string))?;
+
+                let to = graph.find_by_version_regex(&to_regex);
+                to.iter().try_for_each(|(to, to_version)| -> Fallible<()> {
+                    if from_regex_string == ".*" {
+                        let mut froms: Vec<String> = graph
```
vrutkovs:
```suggestion
                        let froms: Vec<String> = graph
```
Vectors don't need to be mutable (they can be extended)

cincinnati/src/plugins/internal/edge_add_remove.rs
```diff
@@ -304,6 +312,83 @@ impl EdgeAddRemovePlugin {
...
+                            .previous_releases(to)
+                            .map(|(_, _, release)| release.version().to_string())
+                            .collect();
+                        trace!(
+                            "adding conditional edge by regex for '{}': {:?}",
+                            to_version,
+                            froms
+                        );
+                        froms.iter_mut().for_each(|from| {
```
vrutkovs:
```suggestion
                        froms.iter().for_each(|from| {
```
ditto

cincinnati/src/plugins/internal/edge_add_remove.rs
```diff
@@ -304,6 +312,83 @@ impl EdgeAddRemovePlugin {
...
+                            conditional_edge
+                                .edges
+                                .append(&mut vec![ConditionalUpdateEdge {
+                                    from: from.to_string(),
+                                    to: to_version.clone(),
+                                }]);
+                        });
+                    } else {
+                        let mut froms: Vec<String> = graph
```
vrutkovs:
```suggestion
                        let froms: Vec<String> = graph
```
ditto

cincinnati/src/plugins/internal/edge_add_remove.rs
```diff
@@ -304,6 +312,83 @@ impl EdgeAddRemovePlugin {
...
+                            .previous_releases(to)
+                            .map(|(_, _, release)| release.version().to_string())
+                            .collect();
+                        debug!(
+                            "adding conditional edge by regex for '{}': {:?}",
+                            to_version, froms
+                        );
+                        froms
+                            .iter_mut()
```
vrutkovs:
```suggestion
                            .iter()
```
ditto

cincinnati/src/plugins/internal/edge_add_remove.rs
```diff
@@ -304,6 +312,83 @@ impl EdgeAddRemovePlugin {
...
+                            .collect();
+                        trace!(
+                            "adding conditional edge by regex for '{}': {:?}",
+                            to_version,
+                            froms
+                        );
+                        froms.iter_mut().for_each(|from| {
+                            conditional_edge
+                                .edges
+                                .append(&mut vec![ConditionalUpdateEdge {
```
vrutkovs:
```suggestion
conditional_edge.edges.push(ConditionalUpdateEdge {
```
We're adding a single `ConditionalUpdateEdge`, so instead of creating a mutable Vector we could use `edges.push`

cincinnati/src/plugins/internal/edge_add_remove.rs
```diff
@@ -304,6 +312,83 @@ impl EdgeAddRemovePlugin {
...
+                        froms
+                            .iter_mut()
+                            .for_each(|from| match from_regex.is_match(from) {
+                                true => {
+                                    debug!(
+                                        "Regex '{}' matches version '{}'",
+                                        &from_regex,
+                                        from.to_string(),
+                                    );
+                                    conditional_edge.edges.append(&mut vec![
```
vrutkovs: Same, `edges.push(ConditionalUpdateEdge {` is sufficient

cincinnati/src/plugins/internal/edge_add_remove.rs
```diff
@@ -304,6 +312,83 @@ impl EdgeAddRemovePlugin {
...
+                                            to: to_version.clone(),
+                                        },
+                                    ])
+                                }
+                                _ => {}
+                            });
+                    }
+                    Ok(())
+                })?;
+                conditional_edges.append(&mut vec![conditional_edge]);
```
vrutkovs:
```suggestion
                conditional_edges.push(conditional_edge);
```

cincinnati/src/lib.rs
```diff
@@ -583,11 +608,22 @@ impl<'a> Deserialize<'a> for Graph {
                     .map_err(|_| {
                         de::Error::invalid_value(serde::de::Unexpected::StructVariant, &self)
                     })?;
+
+                graph.conditional_edges.append(
+                    &mut conditional_edges
+                        .iter()
+                        .map(|ce| ce.clone())
+                        .collect::<Vec<ConditionalEdge>>(),
+                );
```
vrutkovs:
```suggestion
                graph.conditional_edges.extend(conditional_edges);
```
to copy all elements from `conditional_edges` to `graph.conditional_edges`

vrutkovs:
Looking really good, needs a few unit tests and some code clean up

vrutkovs (commented):
looking really good, needs a few more unit tests

cincinnati/src/plugins/internal/graph_builder/openshift_secondary_metadata_parser/plugin.rs
PratikMahajan: I thought of splitting it out, but it'll be just using a function to call another function as we'll be basically splitting out `data_dir.join` and calling the function `deserialize_directory_files`, rest everything will remain as it is.

cincinnati/src/plugins/internal/graph_builder/openshift_secondary_metadata_parser/plugin.rs
PratikMahajan: makes sense, I've left the work to remove edge-risk duplication for later. Added a TODO here.

cincinnati/src/plugins/internal/edge_add_remove.rs
PratikMahajan: fixed

cincinnati/src/plugins/internal/arch_filter.rs
```diff
@@ -51,6 +51,46 @@ impl ArchFilterPlugin {
...
+
+    pub fn remove_conditional_edges(
+        &self,
+        ce: &mut Vec<cincinnati::ConditionalEdge>,
+        arch: String,
+    ) -> Fallible<()> {
+        ce.into_iter()
+            .for_each(|ce: &mut cincinnati::ConditionalEdge| {
+                ce.edges
+                    .retain(|e| e.from.contains(&arch) || e.to.contains(&arch));
```
vrutkovs: I think we want `e.from.contains(&arch) && e.to.contains(&arch)` here. This would protect us from mistakes in rules like "block upgrades from 4.2.3-s390x to 4.2.5-amd64"
Ideally we'd throw an error here, so that we could create a new cincinnati-graph-data test to verify that

cincinnati/src/plugins/internal/arch_filter.rs
```diff
@@ -51,6 +51,46 @@ impl ArchFilterPlugin {
...
+
+        Ok(())
+    }
+
+    fn remove_arch_info(&self, v: &String, arch: String) -> Result<String, Error> {
+        semver::Version::parse(v)
+            .context(v.clone())
+            .map(|mut version| {
+                version.build.retain(|elem| elem.to_string() != arch);
+                trace!("CE: rewriting version {} ->  {}", v, version);
```
vrutkovs:
```suggestion
                trace!("CE: rewriting version {} -> {}", v, version);
```
world's smallest nitpick - removed extra space

cincinnati/src/plugins/internal/arch_filter.rs
```diff
@@ -123,6 +163,7 @@ impl InternalPlugin for ArchFilterPlugin {
 
         // remove all matches from the Graph
         let removed = graph.remove_releases(to_remove);
+        self.remove_conditional_edges(&mut graph.conditional_edges, arch.clone())?;
 
         trace!("removed {} releases", removed);
```
vrutkovs: Add a number of releases removed conditionally (after this trace)

cincinnati/src/plugins/internal/channel_filter.rs
```diff
@@ -86,13 +102,16 @@ impl InternalPlugin for ChannelFilterPlugin {
...
                 .collect()
         };
 
         // remove all matches from the Graph
-        let removed = graph.remove_releases(to_remove);
+        let removed = graph.remove_releases(to_remove.clone());
+        let _removed_ce =
+            self.remove_conditional_edges(&mut graph.conditional_edges, releases_version)?;
 
         trace!("removed {} releases", removed);
```
vrutkovs: Same here - report number of unconditially removed and conditionally removed

vrutkovs:
few more tweaks

cincinnati/src/plugins/internal/arch_filter.rs
PratikMahajan: here we are removing conditional edges not associated with any release in the arch, and similarly for channel in its plugin. We are not removing any release conditionally, we are just keeping the edges for the releases present in the channel or with the arch conditionally available and removing the rest.

cincinnati/src/plugins/internal/arch_filter.rs
```diff
@@ -51,6 +51,46 @@ impl ArchFilterPlugin {
...
+        ce.into_iter()
+            .for_each(|ce: &mut cincinnati::ConditionalEdge| {
+                ce.edges
+                    .retain(|e| e.from.contains(&arch) || e.to.contains(&arch));
+            });
+
+        Ok(())
+    }
+
+    pub fn rm_ce_arch_identifier(
```
wking: nit: I'd rather replace `ce` with `conditional_edge` here, because we'll presumably have more folks reading the code than writing it, and had a brief "what's `ce`?" moment, before I got to the function's argument list.  I don't mind abbreviations in local variables, but function names have a wider reach and may have less context.

cincinnati/src/plugins/internal/arch_filter.rs
wking: so a trace message for the number of conditional edges that existed before and after `remove_conditional_edges`?  `removed {} releases and reduced conditional edges from {} to {}` or some such?  Might read strangely if we hadn't actually removed any edges...

cincinnati/src/plugins/internal/channel_filter.rs
```diff
@@ -41,6 +42,20 @@ impl ChannelFilterPlugin {
...
+
+    pub fn remove_conditional_edges(
+        &self,
+        ce: &mut Vec<cincinnati::ConditionalEdge>,
+        releases: HashSet<String>,
+    ) -> Fallible<()> {
+        ce.into_iter()
+            .for_each(|ce: &mut cincinnati::ConditionalEdge| {
+                ce.edges
+                    .retain(|e| !releases.contains(&e.from) || !releases.contains(&e.to));
```
wking: [De Morgan's laws][1] have `!a || !b = !(a && b)`, and I think the form with fewer `!` is easier to read.  Although in this case, I think you may have the sense flipped.  Shouldn't it actually be `releases.contains(&e.from) && releases.contains(&e.to)` to only retain edges where both the source and target are in the graph?

[1]: https://en.wikipedia.org/wiki/De_Morgan%27s_laws

cincinnati/src/plugins/internal/channel_filter.rs
```diff
@@ -41,6 +42,20 @@ impl ChannelFilterPlugin {
 
         Ok(Box::new(plugin))
     }
+
+    pub fn remove_conditional_edges(
+        &self,
+        ce: &mut Vec<cincinnati::ConditionalEdge>,
+        releases: HashSet<String>,
+    ) -> Fallible<()> {
+        ce.into_iter()
```
wking, line 54: I think this function and it's arch-filter analog need a separate layer somewhere to completely remove `ConditionalEdge` entries if edge-filtering remove all of their edges.

cincinnati/src/plugins/internal/channel_filter.rs
wking: Ah, perhaps the `releases` set being passed in is the set of releases that are being pruned out?  Can we pivot to have it be the set of releases that are being preserved?  With an explicit name like `preserved_releases`?  Because that's a well-defined set based on current graph state, while "releases we should remove" depends on either tracking history or comparing the set of all conditional-edge-mentioned releases with the set of all the current graph's releases.

cincinnati/src/plugins/internal/graph_builder/openshift_secondary_metadata_parser/plugin.rs
```diff
@@ -478,6 +491,54 @@ impl OpenshiftSecondaryMetadataParserPlugin {
...
+            .into_iter()
+            .try_for_each(|cey| -> Fallible<()> {
+                let ce: ConditionalEdge = ConditionalEdge {
+                    edge_regex: ConditionalUpdateEdge {
+                        from: cey.from.to_string(),
+                        to: cey.to.to_string(),
+                    },
+                    edges: vec![],
+                    // Some edges could be blocked for multiple risks, here we are parsing each yaml
+                    // file separately. TODO: remove edge-risk duplication
```
wking: You need to do at least some condensing here.  Constraints:

* Each YAML file contains a single risk, mentioning a single target release and a regexp of source releases.
* By the time you're serving consumers, each `(from, to)` tuple may only occur in a single `conditionalEdges` entry, covering the complete set of relevant `risks` for that edge.  You don't want to have one entry saying that A -> B is vulnerable to risk C, and recommended if risk C doesn't apply, while a second entry says A -> B is vulnerable to risk D, and recommended if risk D doesn't apply.  Because then a cluster might be tempted to update based on "C doesn't apply" despite D applying to that cluster.  Collapsing all the risks for a single edge into one `conditionalEdges` entry is required, not optional.
* The risk-set may be shared between multiple edges (e.g. A -> C and B -> C could have the same risk sets).  Collapsing multiple edges into the same `conditionalEdges` entry is an optional, space-saving mechanism.

Assuming we defer multi-edge collapse, you'll still need to collapse multiple risks for each single edge.  You can do a bit of that here, by collapsing to a single `conditionalEdges` entry for each `(from, to)` tuple.  But because `from` is a regular expression here, and different regular expressions might match the same actual-`from` release, you'll still need a separate multi-risk collapse later on, after the regular expressions have been validated.

cincinnati/src/plugins/internal/edge_add_remove.rs
```diff
@@ -304,6 +312,79 @@ impl EdgeAddRemovePlugin {
...
+                    edge_regex: ce.edge_regex.clone(),
+                    edges: vec![],
+                    risks: ce.risks.clone(),
+                };
+
+                let to_regex_string = &ce.edge_regex.to;
+                let from_regex_string = &ce.edge_regex.from;
+
+                let to_regex = regex::Regex::new(to_regex_string)
+                    .context(format!("Parsing {} as Regex", to_regex_string))?;
```
wking: At the moment, [`from` is a regexp, but `to` is a non-regexp release reference][1].  `to` will need [arch-suffix handling][2], and the regexp in `from` is expected to apply [before the arch suffix is stripped][3]

[1]: https://github.com/openshift/cincinnati-graph-data/tree/bfd85c895952e2517b5593798c3a7b2c51faff43#block-edges
[2]: https://github.com/openshift/cincinnati-graph-data/tree/bfd85c895952e2517b5593798c3a7b2c51faff43#release-names
[3]: https://github.com/openshift/cincinnati-graph-data/blame/bfd85c895952e2517b5593798c3a7b2c51faff43/README.md#L134-L137

cincinnati/src/plugins/internal/arch_filter.rs
PratikMahajan: renamed

cincinnati/src/plugins/internal/channel_filter.rs
PratikMahajan: handled this case, by removing conditional edges where the edges is an empty array

cincinnati/src/plugins/internal/graph_builder/openshift_secondary_metadata_parser/plugin.rs
PratikMahajan: handled this use case in `edge-add-remove` plugin

cincinnati/src/plugins/internal/edge_add_remove.rs
LalatenduMohanty: We should add an unit test to cover the scenario.

cincinnati/src/plugins/internal/arch_filter.rs
PratikMahajan: done

cincinnati/src/plugins/internal/arch_filter.rs
PratikMahajan: changed the statement to include the number of conditional edges removed as well.

cincinnati/src/plugins/internal/channel_filter.rs
PratikMahajan: done

cincinnati/src/plugins/internal/channel_filter.rs
PratikMahajan: like we are doing to the releases, the graph contains all the releases and all the conditional edges for all available architectures. 
here we are removing the releases and removing the conditional edges for the removed releases. 
this is just an extension of the logic that we are using to remove releases and their edges.

cincinnati/src/plugins/internal/edge_add_remove.rs
PratikMahajan: changed it to a string

LalatenduMohanty (changes requested):
We should update tests for openshift_secondary_metadata_parser
with  "conditionalEdges" yamls

cincinnati/src/plugins/internal/channel_filter.rs
```diff
@@ -41,6 +42,25 @@ impl ChannelFilterPlugin {
 
         Ok(Box::new(plugin))
     }
+
+    /// Takes a HashSet of release names and removes the corresponding conditional edges
+    /// containing the releases.  
+    pub fn remove_conditional_edges(
+        &self,
+        ce: &mut Vec<cincinnati::ConditionalEdge>,
+        releases: HashSet<String>,
```
vrutkovs, line 51: This function is duplicated in two plugins - lets move it to `commons`?
Since it simply mutates `ce` and just 4 lines maybe we don't need it to be a function at all

e2e/tests/e2e.rs
```diff
@@ -29,7 +42,7 @@ fn e2e_channel_success(channel: &'static str, arch: &'static str) {
     let expected: cincinnati::plugins::internal::versioned_graph::VersionedGraph =
```
vrutkovs, line 35: We're importing two versions of the graphs for 4.7/4.8, but testing just the stable-4.2/4.3

vrutkovs (changes requested):

cincinnati/src/plugins/internal/channel_filter.rs
PratikMahajan: These are 2 different functions with different params, so cannot move to commons. 
Actually, we don't need this function but created it so that the existing function does not become too long and cluttered.

commons/src/lib.rs
```diff
@@ -32,7 +32,8 @@ use url::form_urlencoded;
 lazy_static! {
     /// list of cincinnati versions
     pub static ref CINCINNATI_VERSION: HashMap<&'static str, i32> =
-        [("application/vnd.redhat.cincinnati.v1+json", 1)]
+        [("application/vnd.redhat.cincinnati.v1+json", 1),
+         ("application/vnd.redhat.cincinnati.v2+json", 2)]
```
wking: I'm not clear on why we need a v2 for this, since adding a new `conditionalEdges` property is a backwards-compatible change.  Can we keep the conditional-edge handling in v1?

LalatenduMohanty:
@PratikMahajan  Targeted edge blocking is not supposed to be in v2 graph. As per https://github.com/openshift/enhancements/blob/2cc2d9b331532c852878a7c793f3a754914c824e/enhancements/update/targeted-update-edge-blocking.md#enhanced-graph-data-schema-for-blocking-edges it would be in the v1 graph. Only the graph data schema needs to be changed to 1.1.0

commons/src/lib.rs
wking: I'll mark this thread resolved, and we can sort this out in [the main comment stream][1].

[1]: https://github.com/openshift/cincinnati/pull/574#issuecomment-984053375

cincinnati/src/lib.rs
```diff
@@ -799,8 +799,12 @@ pub mod testing {
     use super::*;
     use crate::plugins::internal::versioned_graph::VersionedGraph;
 
-    pub fn generate_graph() -> Graph {
+    pub fn generate_graph(include_conditional_edge: bool) -> Graph {
```
LalatenduMohanty, line 802: We should add a comment here about why are adding this option of generating conditional edges based on `include_conditional_edge: bool`

LalatenduMohanty: I meant we should communicate that because it is not possible to test the conditional edges in tests we will use generate_graph(false) for the tests and in future we will fix this

LalatenduMohanty (changes requested):

PR #584, opened 2021-11-23
Do not replace regex if .*

PratikMahajan:
if the block regex is `.*`, do not replace it with other entries.
this comes into play when we are blocking an edge by using multiple
regex.
for eg. for 4.7.4, there are two regex used, `.*` and `4\.6\..*`
In some cases, cincinnati would replace `.*` with `4\.6\..*` which
was causing cincinnati to create an incorrect and inconsistent graph.
this PR adds a condition to check if the existing regex is `.*` and
does not replace it with any other regex.

### DIFF
--- a/cincinnati/src/plugins/internal/graph_builder/openshift_secondary_metadata_parser/plugin.rs
+++ b/cincinnati/src/plugins/internal/graph_builder/openshift_secondary_metadata_parser/plugin.rs
@@ -455,13 +455,38 @@ impl OpenshiftSecondaryMetadataParserPlugin {
                                 &to.to_string()
                             )) {
                                 Ok(metadata) => {
-                                    metadata.insert(
-                                        format!(
-                                            "{}.{}",
-                                            self.settings.key_prefix, "previous.remove_regex"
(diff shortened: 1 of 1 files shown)


cincinnati/src/plugins/internal/graph_builder/openshift_secondary_metadata_parser/plugin.rs
```diff
@@ -455,13 +455,38 @@ impl OpenshiftSecondaryMetadataParserPlugin {
...
+                                                format!(
+                                                    "{}.{}",
+                                                    self.settings.key_prefix,
+                                                    "previous.remove_regex"
+                                                ),
+                                                blocked_edge.from.to_string(),
+                                            );
+                                        }
+                                        Some(value) => {
+                                            if value == ".*" {
```
vrutkovs, line 473: We could create another, more specific arm:
```
Some(".*") => {
    warn!(
        "not replacing `.*` with {}",
        blocked_edge.from.to_string()
    );
}
Some(value) => {
```

PR #592, opened 2021-11-30
e2e: build operator yamls using a template

vrutkovs:
Generate operator yamls using Template, so that we could manage all changes in one file, avoid patching Subscription and have configuration options listed. This also ensures our stage deployment procedure is not wildly different


dist/openshift/cincinnati.yaml
```diff
@@ -0,0 +1,60 @@
...
+  - name: IMAGE
+    value: "quay.io/app-sre/cincinnati"
+    displayName: cincinnati image
+    description: cincinnati docker image. Defaults to quay.io/app-sre/cincinnati
+  - name: IMAGE_TAG
+    value: "latest"
+    displayName: cincinnati version
+    description: cincinnati version which defaults to latest
+  - name: GRAPHDATA_IMAGE
+    value: "registry.ci.openshift.org/cincinnati-ci-public/cincinnati-graph-data:stable"
```
PratikMahajan, line 48: can we fix this to a specific commit that is referenced in `metadata_revision`?

vrutkovs: Yes, however we'd need to build one first (manually I guess)

LalatenduMohanty (approved):
/lgtm

PR #595, opened 2021-12-03
OTA-498: dist: prepare two templates

vrutkovs:
`cincinnati.yaml` is a template which deploys Cincy traditionally (via DeploymentConfig) and used on stage/prod, `cincinnati-e2e` is an operator-based deployment for e2e tests


LalatenduMohanty:
This is better. Because I was also not sure if the changing cincinnati.yaml will impact stage and production.

LalatenduMohanty (approved):
/lgtm

vrutkovs:
/hold

Needs manifestref fix

PR #598, opened 2021-12-07
Bug 2008125: Cargo: use latest dkregistry-rs

vrutkovs:
Bump dkregistry-rs to latest as of Dec 7, so that we could make use of two important fixes:
* Allow passing custom set of accepted types with optional weight
* make Www-Authentication header parsing case-insensitive


wking (approved):
/lgtm

vrutkovs:
/hold

This PR fixes two bugs, so we might want to retitle it to ease the process

vrutkovs:
Actually, only one bug would be directly fixed by this bump, the other one requires additional commits

/retitle Bug 2008125: Cargo: use latest dkregistry-rs

PR #599, opened 2021-12-07
add graph-data dockerfile for e2es

PratikMahajan:
add the graph-data dockerfile to create a graph-data container
for cincinnati e2es.
this graph-data container created will be pinned to a specific
commit that we will be using for e2es.
the graphs for e2es will be generated based on this pinned
graph-data commit.


PratikMahajan:
decided to use a manually uploaded graph-data image to `registry.ci.openshift.org/cincinnati-ci-public/cincinnati-graph-data:<hash>`

PR #600, opened 2021-12-07
start using graph-data container with e2es commit

PratikMahajan:
start using the graph-data container which is created
with the commit that we are using in reference_revision.


PratikMahajan (commented):
/close this PR is being replaced by #663

PR #610, opened 2021-12-15
Rework load testing

vrutkovs:
* run scripts in parallel as cluster pods, so that router would not limit us
* save metrics is cluster Prometheus, so that it could be inspected by PromeCIeus
* bump worker/rate to emulate 300rps


PratikMahajan (approved):
/lgtm

PR #611, opened 2021-12-15
cincinnati_graph_fetch: Cache graph responses

vrutkovs:
Use `cached` crate to save successful graph requests for 60 secs. This should prevent PE from bashing GB when rps grows.

TODO:
* [ ] Expose this as PE config parameter 
       Its apparently not possible - Cache settings have to be consts and are immutable

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -367,6 +367,25 @@ dependencies = [
 ]
 
 [[package]]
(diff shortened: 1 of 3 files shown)


LalatenduMohanty (approved):
/lgtm

PR #627, opened 2022-01-03
Fix issues found by rust-clippy

vrutkovs:
[rust-clippy](https://github.com/rust-lang/rust-clippy) is a Rust linter. It is useful to find possible optimizations: extra clones, refs and other code improvements.

List of issues and improvements found by clippy:
https://rust-lang.github.io/rust-clippy/master/index.html#bool_assert_comparison
https://rust-lang.github.io/rust-clippy/master/index.html#comparison_chain
https://rust-lang.github.io/rust-clippy/master/index.html#from_over_into
https://rust-lang.github.io/rust-clippy/master/index.html#into_iter_on_ref
https://rust-lang.github.io/rust-clippy/master/index.html#iter_cloned_collect
https://rust-lang.github.io/rust-clippy/master/index.html#iter_nth_zero
https://rust-lang.github.io/rust-clippy/master/index.html#manual_unwrap_or
https://rust-lang.github.io/rust-clippy/master/index.html#match_like_matches_macro
https://rust-lang.github.io/rust-clippy/master/index.html#match_single_binding
https://rust-lang.github.io/rust-clippy/master/index.html#needless_arbitrary_self_type
https://rust-lang.github.io/rust-clippy/master/index.html#needless_borrow
https://rust-lang.github.io/rust-clippy/master/index.html#needless_question_mark
https://rust-lang.github.io/rust-clippy/master/index.html#needless_return
https://rust-lang.github.io/rust-clippy/master/index.html#option_as_ref_deref
https://rust-lang.github.io/rust-clippy/master/index.html#or_fun_call
https://rust-lang.github.io/rust-clippy/master/index.html#ptr_arg
https://rust-lang.github.io/rust-clippy/master/index.html#redundant_clone
https://rust-lang.github.io/rust-clippy/master/index.html#self_named_constructors
https://rust-lang.github.io/rust-clippy/master/index.html#single_char_pattern
https://rust-lang.github.io/rust-clippy/master/index.html#single_component_path_imports
https://rust-lang.github.io/rust-clippy/master/index.html#single_match
https://rust-lang.github.io/rust-clippy/master/index.html#try_err
https://rust-lang.github.io/rust-clippy/master/index.html#unnecessary_mut_passed
https://rust-lang.github.io/rust-clippy/master/index.html#unnecessary_sort_by
https://rust-lang.github.io/rust-clippy/master/index.html#useless_conversion
https://rust-lang.github.io/rust-clippy/master/index.html#useless_format
https://rust-lang.github.io/rust-clippy/master/index.html#vec_init_then_push

### DIFF
--- a/cincinnati/src/plugins/internal/graph_builder/commons.rs
+++ b/cincinnati/src/plugins/internal/graph_builder/commons.rs
@@ -18,11 +18,4 @@ pub mod tests {
             registry::cache::new(),
         )
     }
-
-    #[allow(dead_code)]
-    pub fn remove_metadata_by_key(releases: &mut Vec<release::Release>, key: &str) {
-        for release in releases.iter_mut() {
-            release.metadata.metadata.remove(key);
-        }
-    }
 }
--- a/cincinnati/src/plugins/internal/graph_builder/commons.rs
+++ b/cincinnati/src/plugins/internal/graph_builder/commons.rs
@@ -19,6 +19,7 @@ pub mod tests {
         )
     }
 
+    #[allow(dead_code)]
     pub fn remove_metadata_by_key(releases: &mut Vec<release::Release>, key: &str) {
         for release in releases.iter_mut() {
             release.metadata.metadata.remove(key);
--- a/Justfile
+++ b/Justfile
@@ -20,6 +20,9 @@ metadata_reference_revision:
 format:
 	cargo fmt --all
 
+clippy:
+	cargo clippy --all-targets --all-features
+
 commit +args="": format
 	git commit {{args}}
 
--- a/cincinnati/src/lib.rs
+++ b/cincinnati/src/lib.rs
@@ -208,7 +208,7 @@ impl Graph {
     {
         let missing_manifest_ref = String::from("none");
         let release = release.into();
-        match self.find_by_version(&release.version()) {
+        match self.find_by_version(release.version()) {
             Some(id) => {
                 let node = self.dag.node_weight_mut(id.0).expect(EXPECT_NODE_WEIGHT);
                 if let Release::Concrete(_) = node {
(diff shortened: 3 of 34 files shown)


cincinnati/src/plugins/internal/graph_builder/commons.rs
```diff
@@ -19,6 +19,7 @@ pub mod tests {
         )
     }
 
+    #[allow(dead_code)]
```
wking: f9076203267ac doesn't explain why we're allowing dead code here or elsewhere.  Can you add a note with motivation?  I'd expect there is some reason we can't just drop this dead code, and it would be good to have that available via `git blame ...`.

LalatenduMohanty:
Can we add more explanation for the commit `versioned_graph: avoid self-named constructors` i.e. why avoiding self-named constructors is a good thing.  Also `Implement fixes suggested by rust-clippy ` looks fine but we should approach it from a point of view of what we are fixing from a good coding standard or coding practice point of view.

cincinnati/src/plugins/internal/graph_builder/commons.rs
vrutkovs: This function is not being used now, and I don't think it has ever been. I guess its safe to be removed, but I'd rather mark it as dead code in this PR and remove it in the follow up

vrutkovs:
Updated the PR description with a list of issue classes and their description

cincinnati/src/plugins/internal/graph_builder/commons.rs
LalatenduMohanty: @vrutkovs I would suggest that we can remove the code (this and others) completely. This is not a large function so we can add it back if we need it in future.

LalatenduMohanty (changes requested):

LalatenduMohanty:
@vrutkovs  I see that you have removed some dead code in https://github.com/openshift/cincinnati/pull/627/commits/943b33bcb1bc75f2599644ae92b3b570379060a0 but there are some more which you have not removed. I am not sure why.

vrutkovs:
There are plenty dead_code annotation in the structures - as we don't read them / some properties. We should keep them though, as cincinnati-graoh-data may use them and to maintain the protocol stability

LalatenduMohanty (approved):
/lgtm

PR #656, opened 2022-02-02
OTA-561: Make actix-web HttpServer settings configurable

vrutkovs:
Allow changing actix-web settings without recompiling OSUS.

/cc @LalatenduMohanty @wking 
/hold

until load testing on stage is complete

### DIFF
--- a/policy-engine/src/config/settings.rs
+++ b/policy-engine/src/config/settings.rs
@@ -8,6 +8,7 @@ use custom_debug_derive::Debug as CustomDebug;
 use hyper::Uri;
 use std::collections::HashSet;
 use std::net::{IpAddr, Ipv4Addr};
+use std::time::Duration;
 use structopt::StructOpt;
 
 /// Default URL to upstream graph provider.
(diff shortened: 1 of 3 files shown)


wking (approved):
/lgtm

I expect the results of load testing to help inform settings that are specific to our deploy.  That shouldn't block the addition of knobs here, but I don't mind leaving the hold in place if that's more convenient.

policy-engine/src/config/settings.rs
```diff
@@ -51,6 +52,17 @@ pub struct AppSettings {
...
     pub tracing_endpoint: Option<String>,
+
+    #[default(10)]
+    pub backlog: u32,
+    #[default(10)]
+    pub max_connections: usize,
+    #[default(64)]
+    pub max_connection_rate: usize,
+    #[default(Duration::new(5, 0))]
+    pub keep_alive: Duration,
```
vrutkovs: This needs to be `Option<Duration>` to handle `0` correctly

policy-engine/src/config/settings.rs
```diff
@@ -51,6 +52,17 @@ pub struct AppSettings {
...
+    #[default(10)]
+    pub backlog: u32,
+    #[default(10)]
+    pub max_connections: usize,
+    #[default(64)]
+    pub max_connection_rate: usize,
+    #[default(None)]
+    pub keep_alive: Option<Duration>,
+    #[default(Duration::new(5, 0))]
+    pub client_timeout: Duration,
```
LalatenduMohanty, line 70: @vrutkovs Looks like we are setting client_timeout as 5 seconds.  which looks fine to me.  What will be the value for keep_alive?

LalatenduMohanty: I remember that we set the keep_alive somewhere else but can not seem to find it now.

vrutkovs: Default for PE would be 0. We're setting metrics server keep_alive to [10 seconds](https://github.com/openshift/cincinnati/blob/master/policy-engine/src/main.rs#L132) (so that kubelet requests would re-use the same connection) and [in GB main server](https://github.com/openshift/cincinnati/blob/master/graph-builder/src/main.rs#L131)

wking: probably worth documenting where the new defaults are coming from in the commit message, even if the reasoning is just "passing through the current actix default" or "wildly guessing" ;)

LalatenduMohanty (approved):
/lgtm

wking (approved):
/lgtm

vrutkovs:
/hold cancel

Lets have this merged for another round of load testing

PR #663, opened 2022-02-08
Add e2e to check Targeted Edge Blocks

PratikMahajan:


hack/e2e.sh
```diff
@@ -77,10 +77,18 @@ export E2E_TESTDATA_DIR
 read -r E2E_METADATA_REVISION <"${E2E_TESTDATA_DIR}"/metadata_revision
 export E2E_METADATA_REVISION
 
+CI_REGISTRY_URL="registry.ci.openshift.org/cincinnati-ci-public/cincinnati-graph-data"
+
+if [[ ! -z "${GRAPHDATA_IMAGE}" ]]; then
```
vrutkovs:
```suggestion
export GRAPH_SOURCE="${GRAPHDATA_IMAGE:-registry.ci.openshift.org/cincinnati-ci-public/cincinnati-graph-data:${E2E_METADATA_REVISION}}"
```
Similar to what we do in [cincinnati-operator](https://github.com/openshift/cincinnati-operator/blob/master/hack/deploy.sh#L10)

vrutkovs:
LGTM

I don't think there is a way to emulate conditional edge hit, right?

PratikMahajan:
> LGTM
> 
> I don't think there is a way to emulate conditional edge hit, right?

this PR relaces the existing e2es with the e2es for the latest versions of OpenShift. 
We cannot emulate specific conditional hit in Cincinnati, but we might be able to do it on the CVO side (?)

vrutkovs (approved):
/lgtm

PR #681, opened 2022-03-02
add additional optional e2es for testing graph output

PratikMahajan:
add test that checks the graph output for a valid graph.
this makes sure that the graph output conforms to the
versioned graph format.


LalatenduMohanty (approved):
/lgtm

PR #688, opened 2022-03-14
use in cluster pull-secret instead of default ci location

PratikMahajan:


hack/e2e.sh
```diff
@@ -57,8 +57,10 @@ function backoff() {
     return $failed
 }
 
-# Use defined PULL_SECRET or fall back to CI location
-PULL_SECRET=${PULL_SECRET:-/var/run/secrets/ci.openshift.io/cluster-profile/pull-secret}
+oc -n openshift-config get --template='{{index .data ".dockerconfigjson" | base64decode}}' secret pull-secret > /tmp/pull_secret
```
wking: to skip this call when `PULL_SECRET` is already set, maybe something like:

```sh
if test -z "${PULL_SECRET}"
then
  PULL_SECRET=/tmp/pull_secret
  oc -n openshift-config get --template='{{index .data ".dockerconfigjson" | base64decode}}' secret pull-secret > "${PULL_SECRET}"
fi
```

wking (approved):
/lgtm

PR #690, opened 2022-03-21
add grafana dashboard and creation script

PratikMahajan:
this PR adds the grafana dashboard to the cincinnati repo
also adds a `just` command to create the grafana config map.

To create a configmap for grafana dashboard or to update existing configmap
1. insert the `<dashboard-name>.json` grafana file to `dist/grafana` folder
2. run `just dashboards` from the root directory to create corresponding
config maps.
3. the config maps are stored in `/dist/grafana/dashboards` folder

### DIFF
--- a/dist/grafana/cincinnati-sli.json
+++ b/dist/grafana/cincinnati-sli.json
@@ -1,0 +1,2235 @@
+{
+  "annotations": {
+    "list": [
+      {
+        "builtIn": 1,
+        "datasource": "-- Grafana --",
+        "enable": true,
+        "hide": true,
+        "iconColor": "rgba(0, 211, 255, 1)",
+        "name": "Annotations & Alerts",
+        "target": {
+          "limit": 100,
+          "matchAny": false,
+          "tags": [],
+          "type": "dashboard"
+        },
+        "type": "dashboard"
+      }
+    ]
+  },
+  "editable": true,
+  "fiscalYearStartMonth": 0,
+  "gnetId": null,
+  "graphTooltip": 0,
+  "iteration": 1647548089423,
+  "links": [],
+  "liveNow": false,
+  "panels": [
+    {
+      "collapsed": false,
+      "datasource": null,
+      "fieldConfig": {
+        "defaults": {},
+        "overrides": []
+      },
+      "gridPos": {
+        "h": 1,
+        "w": 24,
+        "x": 0,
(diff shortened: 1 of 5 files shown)


dist/grafana/cincinnati-sli.json
```diff
@@ -0,0 +1,2235 @@
...
+      "6h",
+      "12h",
+      "24h",
+      "2d",
+      "7d",
+      "30d"
+    ]
+  },
+  "timezone": "utc",
+  "title": "Cincinnati SLI",
```
wking: nit: can we spell out "Cincinnati Service Level Indicators (SLIs)" or some such here.  I don't do enough ops to have SLI internalized as an acronym.

dist/grafana/grafana-dashboard.configmap.yaml.tpl
```diff
@@ -0,0 +1,13 @@
...
+kind: ConfigMap
+metadata:
+  name: GRAFANA_METADATA
+  labels:
+    grafana_dashboard: "true"
+  annotations:
+    grafana-folder: /grafana-dashboard-definitions/Cincinnati
+data:
+  cincinnati.json: |-
+GRAFANA_MANIFEST
```
wking: nit: can we indent `GRAFANA_MANIFEST` here and then adjust the `sed` to preserve that indentation when injecting the JSON?  That seems like it would be slightly less magical.

dist/grafana/grafana-dashboard.configmap.yaml.tpl
```diff
@@ -0,0 +1,13 @@
+# WARNING: This is generated from dist/grafana/
+# please modify there and run `just dashboards`
```
wking: The `tpl` isn't auto-generated.  Maybe shift this bit into the `Justfile` so it only gets injected into the rendered ConfigMap YAML?

dist/grafana/grafana-dashboard.configmap.yaml.tpl
PratikMahajan: I tried indenting here itself, but it still requires `'s/^/    /'` bit in the sed to make sure it is correctly indented in the file.  indenting `GRAFANA_MANIFEST` would just affect the 1st line of the json.

wking (approved):
/lgtm

PR #702, opened 2022-06-06
fix version matching for conditional edges

PratikMahajan:
we were using `.contains` method as the versions are suffixed with
architectures like `+amd64`, etc.
This inadvertently led to blocking versions like `4.7.45` which matched
`4.7.4` which was conditionally blocked.
This commit fixes the issue.


wking (approved):
/lgtm

wking:
Failure mode seems common among Cincinnati PRs, the change is straightforward enough, and getting this code into staging is important enough, that I don't want to block this PR on a rotted presubmit:

/override ci/prow/e2e

PR #711, opened 2022-07-07
add multi-arch images support

PratikMahajan:
adds support for multi arch images.
we are using ManifestLists to represent multi arch images

As manifest list images only contain digests of the
images contained in the manifest, the layers_digests
function from dkregistry returns the digest of all the images
contained in the ManifestList instead of individual
layers of the manifests.
The layers of a specific image from manifest list is
obtained using the digest of the image from the
manifest list and getting it's manifest and manifestref
(get_manifest_and_ref()) and using this manifest of
the individual image to get the layers.

### DIFF
--- a/cincinnati/src/plugins/internal/graph_builder/release_scrape_dockerv2/registry/mod.rs
+++ b/cincinnati/src/plugins/internal/graph_builder/release_scrape_dockerv2/registry/mod.rs
@@ -34,7 +34,8 @@ use std::string::String;
 use std::sync::Arc;
 use tar::Archive;
 
-use dkregistry::mediatypes::MediaTypes::{ManifestV2S1Signed, ManifestV2S2};
+use dkregistry::mediatypes::MediaTypes::{ManifestList, ManifestV2S1Signed, ManifestV2S2};
+use dkregistry::v2::Client;
 
 /// Module for the release cache
 pub mod cache {
@@ -215,6 +216,7 @@ pub async fn new_registry_client(
             .accepted_types(Some(vec![
                 (ManifestV2S2, None),
                 (ManifestV2S1Signed, Some(0.8)),
+                (ManifestList, Some(0.5)),
             ]));
         let scope = format!("repository:{}:pull", &repo);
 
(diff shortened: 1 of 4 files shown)


cincinnati/src/plugins/internal/graph_builder/release_scrape_dockerv2/registry/mod.rs
```diff
@@ -244,6 +246,51 @@ pub async fn new_registry_client(
...
+) -> Result<(Option<String>, String, Vec<String>), Error> {
+    trace!("[{}] Fetching release", tag);
+    let (tag, manifest, manifestref) =
+        get_manifest_and_ref(tag, repo.to_owned(), &registry_client).await?;
+
+    // Try to read the architecture from the manifest
+    let arch = match manifest.architectures() {
+        Ok(archs) => {
+            // We don't support ManifestLists now, so we expect only 1
+            // architecture for the given manifest
```
wking: This comment is now stale, and can be dropped :)

cincinnati/src/plugins/internal/graph_builder/release_scrape_dockerv2/registry/mod.rs
```diff
@@ -274,43 +321,25 @@ pub async fn fetch_releases(
...
-                }
-                Err(e) => {
-                    error!(
-                        "could not get architecture from manifest for tag {}: {}",
-                        tag, e
+            let (arch, manifestref, mut layers_digests) =
+                get_manifest_layers(tag.to_owned(), &repo, &registry_client).await?;
+
+            // if the image is multi arch, we will have to get one image from the manifest list and
+            // use its metadata as manifest list is just a collection of manifests
```
wking: nit:

```rust
// if the image is multi arch, we will have to get one image from the manifest list and
// use its metadata, because manifest lists are just collections of manifests and don't
// have their own layers with metadata files.
```

or some such.  Mostly because my first read of "use its metadata as..." left me expecting more about how we would consume the metadata, but in fact you are pivoting to explain why we need to get the metadata from the referenced image, and not talking about what we do with the metadata once we've found it.

wking:
Oh, can we also update [these docs][1] to something like:

```markdown
[the architecture identifier][image-config-properties] for the currently installed cluster, or `multi` for payloads that support heterogeneous clusters
```

[1]: https://github.com/openshift/cincinnati/blob/master/docs/design/openshift.md#request

wking (approved):
/lgtm

LalatenduMohanty:
As discussed we are merging #718 first then we will merge this PR with tests.

LalatenduMohanty:
Removed the hold as it is not possible to add tests for multi-arch images now. So we will track it is a different ticket.

PratikMahajan:
/override ci/prow/e2e 

ci/prow/e2e is expected to fail as its not getting the prometheus token. 
ci/prow/customrust-e2e is the same test which is able to get the prometheus token and has passed. Safe to assume that ci/prow/e2e will pass as well.

PR #717, opened 2022-07-20
update api version for pod disruption budget

PratikMahajan:
policy/v1beta1 PodDisruptionBudget is deprecated in v1.21+,
unavailable in v1.25+; use policy/v1 PodDisruptionBudget

### DIFF
--- a/dist/openshift/cincinnati.yaml
+++ b/dist/openshift/cincinnati.yaml
@@ -201,7 +201,7 @@ objects:
           targetPort: ${{PE_STATUS_PORT}}
       selector:
         deploymentconfig: cincinnati
-  - apiVersion: policy/v1beta1
+  - apiVersion: policy/v1
     kind: PodDisruptionBudget
     metadata:
       name: cincinnati-pdb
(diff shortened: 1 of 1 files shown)


dist/openshift/cincinnati.yaml
```diff
@@ -201,13 +201,13 @@ objects:
...
         deploymentconfig: cincinnati
-  - apiVersion: policy/v1beta1
+  - apiVersion: policy/v1
     kind: PodDisruptionBudget
     metadata:
       name: cincinnati-pdb
       namespace: cincinnati
     spec:
-      minAvailable: 1
+      minAvailable: 3
```
LalatenduMohanty: For disconnected environments we do not need replica count as 3. So I do not think we need to change this here.

LalatenduMohanty (changes requested):

dist/openshift/cincinnati.yaml
PratikMahajan: removed

LalatenduMohanty (approved):
/lgtm

PR #718, opened 2022-07-28
do not panic on sha-mismatch for multi-arch images

PratikMahajan:
check the architecture of images before panicking for sha
mismatch.
this is needed as the multi-arch images for single-arch
have the same version as single-arch images but a different sha.
multi-arch images for single-arch contain additional information
and are different from the images built specifically for single-arch

cincinnati will ignore single-arch tagged images
from multi-arch manifest and will only keep single-arch images in
the graph.

### DIFF
--- a/cincinnati/src/lib.rs
+++ b/cincinnati/src/lib.rs
@@ -87,6 +87,20 @@ impl Release {
             _ => bail!("could not get manifest reference"),
         }
     }
+
+    /// return the arch identifier set in metadata or returns `none`
+    pub fn metadata_arch_id(&mut self) -> String {
+        let metadata_arch_key = "release.openshift.io/architecture";
+        let no_arch = String::from("none");
+        let arch = self
+            .get_metadata_mut()
+            .map(|metadata| metadata.get(metadata_arch_key))
+            .unwrap_or(None);
+        match arch {
+            Some(arch) => arch.to_string(),
+            _ => no_arch,
+        }
+    }
 }
 
 /// Type to represent a Release with all its information.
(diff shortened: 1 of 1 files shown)


cincinnati/src/lib.rs
```diff
@@ -216,12 +216,29 @@ impl Graph {
...
+                            .unwrap()
+                            .get("release.openshift.io/architecture")
+                            .unwrap_or(&no_arch);
+                        let node_arch = node
+                            .get_metadata_mut()
+                            .unwrap()
+                            .get("release.openshift.io/architecture")
+                            .unwrap_or(&no_arch);
+
+                        if release_arch == node_arch {
```
wking: I'm not clear on how this whole `Conrete` block avoids comparing the single-arch amd64 and s390x, etc. releases for the same version today.  Is there some by-arch lookup higher in the stack?

PratikMahajan: the `find_by_version()` which this code snippet is a part of uses version+arch to get to this block. 
so when we have two same versions+arch here, if its an abstract release it gets replaced by concrete release. 
But if there is concrete release already in the graph, we were checking for integrity before and bailing out on sha sum mismatch. 
now the new code block checks the `release.openshift.io/architecture` field and if both are different, it ignores the ones with `multi` set in `release.openshift.io/architecture`. 
So effectively we are ignoring single-arch images built for multi-arch clusters and keeping single arch releases specifically built for single-arch. 
the manifest list images wont be affected here because sha sum will be same and if it differs, it should bailout like it does today

wking (approved):
/lgtm

LalatenduMohanty (approved):
/lgtm

PratikMahajan:
/override ci/prow/e2e

test ci/prow/customrust-e2e uses the same script as ci/prow/e2e. Its passing. confident of e2e passing as well

PR #719, opened 2022-07-29
fix e2e prometheus, operator-group, 4.11 errors

PratikMahajan:

PR #730, opened 2022-08-17
OCPBUGS-179: fix conditional edge for type always

PratikMahajan:
conditional edge has an empty promql when the type of
conditional block is `Always`
This commit skips the serlialization if promql if the
promql condition is empty, which is the case for
type `Always`

Co-Authored-By: W. Trevor King <wking@tremily.us>

### DIFF
--- a/cincinnati/src/lib.rs
+++ b/cincinnati/src/lib.rs
@@ -818,12 +818,49 @@ pub mod testing {
     use super::*;
     use crate::plugins::internal::versioned_graph::VersionedGraph;
 
-    pub fn generate_graph(include_conditional_edge: bool) -> Graph {
+    pub fn generate_graph(include_conditional_edge: bool, include_always_condition: bool) -> Graph {
         let mut graph = Graph::default();
 
+        if include_conditional_edge {
+            let mut ce = ConditionalEdge {
+                edge_regex: ConditionalUpdateEdge {
+                    from: "1[.]0[.].*".to_string(),
+                    to: "2.0.0".to_string(),
+                },
+                edges: vec![ConditionalUpdateEdge {
+                    from: "1.0.0".to_string(),
+                    to: "2.0.0".to_string(),
+                }],
+                risks: vec![ConditionalUpdateRisk {
+                    url: "https://bug.example.com/show_bug.cgi?id=example".to_string(),
+                    name: "BrokenUpdates".to_string(),
+                    message: "Updates are broken for this provider".to_string(),
+                    matching_rules: vec![ClusterCondition {
+                        condition_type: "PromQL".to_string(),
+                        promql: PromQLClusterCondition {
+                            promql: "cluster_infrastructure_provider{type=\"CloudProvider\"}"
+                                .to_string(),
+                        },
+                    }],
+                }],
+            };
+            if include_always_condition {
+                ce.risks = vec![ConditionalUpdateRisk {
+                    url: "https://bug.example.com/show_bug.cgi?id=example".to_string(),
+                    name: "AllBrokenUpdates".to_string(),
+                    message: "All Updates are broken".to_string(),
+                    matching_rules: vec![ClusterCondition {
+                        condition_type: "Always".to_string(),
+                        promql: Default::default(),
+                    }],
+                }]
+            }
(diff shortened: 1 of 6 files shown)


cincinnati/src/lib.rs
```diff
@@ -1169,13 +1193,22 @@ mod tests {
 
     #[test]
     fn serialize_graph_with_conditional_edges() {
-        let graph = generate_graph(true);
+        let graph = generate_graph(true, false);
         assert_eq!(
             serde_json::to_string(&graph).unwrap(),
             r#"{"nodes":[{"version":"1.0.0","payload":"image/1.0.0","metadata":{}},{"version":"2.0.0","payload":"image/2.0.0","metadata":{}},{"version":"3.0.0","payload":"image/3.0.0","metadata":{}}],"edges":[[0,1],[1,2],[0,2]],"conditionalEdges":[]}"#
```
wking: `"conditionalEdges":[]` is not a very interesting set of conditional edges.  Can we use a PromQL example in this `generate_graph` case?

cincinnati/src/lib.rs
```diff
@@ -818,12 +818,36 @@ pub mod testing {
...
 
-    pub fn generate_graph(include_conditional_edge: bool) -> Graph {
+    pub fn generate_graph(include_conditional_edge: bool, include_always_condition: bool) -> Graph {
         let mut graph = Graph::default();
 
+        if include_conditional_edge && include_always_condition {
+            graph.conditional_edges = Some(vec![ConditionalEdge {
+                edge_regex: ConditionalUpdateEdge {
+                    from: "1.0.0".to_string(),
+                    to: "2\\.0\\.*".to_string(),
```
wking: This expression should probably be `2[.]0[.].*`.  `[.]` is easier for me to think about than stacking escapes with `\\.`.  And you need an un-escaped `.` for the `.*` portion.

wking: err, also, when we declare conditional risks in graph-data, we have a literal `to` and a regular expression `from`.  In this code at the moment, you have a literal `from` and a regular expression `to`, which seems backwards.

cincinnati/src/lib.rs
```diff
@@ -818,12 +818,36 @@ pub mod testing {
...
+                    from: "1.0.0".to_string(),
+                    to: "2.0.0".to_string(),
+                }],
+                risks: vec![ConditionalUpdateRisk {
+                    url: "https://bug.example.com/show_bug.cgi?id=example".to_string(),
+                    name: "AllBrokenUpdates".to_string(),
+                    message: "All Updates are broken".to_string(),
+                    matching_rules: vec![ClusterCondition {
+                        condition_type: "Always".to_string(),
+                        promql: PromQLClusterCondition {
```
wking, line 840: if we leave this unpopulated, what happens?

PratikMahajan: using `Default:default()` for `Always` condition

cincinnati/src/lib.rs
PratikMahajan: addressed

cincinnati/src/lib.rs
PratikMahajan: fixed

cincinnati/src/lib.rs
```diff
@@ -821,28 +821,41 @@ pub mod testing {
...
         let mut graph = Graph::default();
 
-        if include_conditional_edge && include_always_condition {
-            graph.conditional_edges = Some(vec![ConditionalEdge {
+        if include_conditional_edge {
+            let mut ce = ConditionalEdge {
                 edge_regex: ConditionalUpdateEdge {
-                    from: "1.0.0".to_string(),
-                    to: "2\\.0\\.*".to_string(),
+                    from: "1[.]0[.]*".to_string(),
```
wking: still needs a `.*`, like `1[.]0[.].*`

wking (approved):
/lgtm

wking (approved):
/lgtm

PratikMahajan:
/override ci/prow/e2e 

Will be fixed in https://github.com/openshift/cincinnati/pull/719

PR #733, opened 2022-08-31
update grafana dashboard to include additional metrics

PratikMahajan:
include the graph-staleness and container restarts
metrics as well as create tables to include the stats


LalatenduMohanty (approved):
/lgtm

PR #740, opened 2022-10-17
[policy-engine] add liveness and readiness probes

PratikMahajan:
This PR adds liveness and readiness probes to the
policy-engine.
The liveness holds true when the application has initialized.
The readiness endpoint waits for the graph-builder to build
the graph and for the graph to be cached on the policy engine side
before going true.

### DIFF
--- a/dist/openshift/cincinnati.yaml
+++ b/dist/openshift/cincinnati.yaml
@@ -144,8 +144,16 @@ objects:
                 - name: status-pe
                   containerPort: ${{PE_STATUS_PORT}}
               livenessProbe:
-                tcpSocket:
-                  port: ${{PE_PORT}}
+                httpGet:
+                  path: /livez
+                  port: ${{GB_STATUS_PORT}}
+                initialDelaySeconds: 60
+                periodSeconds: 30
+                timeoutSeconds: 3
+              readinessProbe:
+                httpGet:
+                  path: /readyz
+                  port: ${{GB_STATUS_PORT}}
                 initialDelaySeconds: 60
                 periodSeconds: 30
                 timeoutSeconds: 3
--- a/policy-engine/src/main.rs
+++ b/policy-engine/src/main.rs
@@ -22,13 +22,14 @@ extern crate custom_debug_derive;
 mod config;
 mod graph;
 mod openapi;
+mod status;
 
 use actix_cors::Cors;
 use actix_service::Service;
 use actix_web::http::StatusCode;
-use actix_web::{middleware, App, HttpRequest, HttpResponse, HttpServer};
+use actix_web::{http, middleware, App, HttpRequest, HttpResponse, HttpServer};
 use cincinnati::plugins::BoxedPlugin;
-use commons::metrics::{self, RegistryWrapper};
+use commons::metrics::{self, HasRegistry};
 use commons::prelude_errors::*;
 use commons::tracing::{get_tracer, init_tracer, set_span_tags};
 use futures::future;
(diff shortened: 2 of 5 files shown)


policy-engine/src/main.rs
```diff
@@ -72,33 +75,59 @@ async fn main() -> Result<(), Error> {
...
                 actix_web::web::resource("/metrics")
-                    .route(actix_web::web::get().to(metrics::serve::<RegistryWrapper>)),
+                    .route(actix_web::web::get().to(metrics::serve::<AppState>)),
+            )
+            .service(
+                actix_web::web::resource("/liveness")
+                    .route(actix_web::web::get().to(status::serve_liveness)),
+            )
+            .service(
+                actix_web::web::resource("/readiness")
```
wking: A common convention in Kube-land (apparently inherited [from Google][1] [z-pages][2]) is to use [`/readyz` and `/livez` for these][3].  I'm agnostic about including the `z`, but I have a weakly-held opinion that we should avoid striking out on our own with the `ness` suffix ;)

[1]: https://github.com/kubernetes/kubernetes/pull/419#discussion_r14854689
[2]: https://stackoverflow.com/questions/43380939/where-does-the-convention-of-using-healthz-for-application-health-checks-come-f
[3]: https://kubernetes.io/docs/reference/using-api/health-checks/#api-endpoints-for-health

policy-engine/src/main.rs
```diff
@@ -72,33 +75,59 @@ async fn main() -> Result<(), Error> {
...
+                    .route(actix_web::web::get().to(status::serve_readiness)),
             )
     })
     .bind((settings.status_address, settings.status_port))?
     .run();
 
+    // metrics endpoints has started running
+    *state.live.write() = true;
```
wking: This seems like a brittle way to say we're alive now, because it won't change if something happens in the container that would benefit from restarting the container.  But I'm not sure when the container would benefit from a restart.  Possibly if the graph-data cache gets stale, a fresh container might have better luck connecting?  But if the issue is on the graph-data side, then a fresh policy-engine container might just lose the cache from the last healthy graph-data.  Are there any deadlock-y sorts of things we're worried about?

policy-engine/src/main.rs
```diff
@@ -141,7 +170,8 @@ async fn main() -> Result<(), Error> {
     .run();
 
     BUILD_INFO.inc();
-
+    // application has initialized.
+    *state.ready.write() = true;
```
wking: We probably want to wait until we've synchronized our graph-data cache before going ready, because once we enter the load balancer, we will begin serving client requests.  Skimming the above^^, I'm not clear on whether or not that's happened by this point.

policy-engine/src/main.rs
PratikMahajan: fixed

policy-engine/src/main.rs
PratikMahajan: the liveness probe goes true when the container is able to serve the requests. 
if the endpoint is unreachable, the container will restart. 
the graph cache on the policy engine will go stale only when the graph-builder fails catastrophically, in that case restarting the policy-engine wont help the application. we dont want the policy-engine restarting in case the graph-builder fails as that will mean that we wont be able to serve any request.

policy-engine/src/main.rs
PratikMahajan: added additional logic to wait for the graph to refresh and cache before going ready

dist/openshift/cincinnati.yaml
```diff
@@ -144,8 +144,16 @@ objects:
...
                   containerPort: ${{PE_STATUS_PORT}}
               livenessProbe:
-                tcpSocket:
-                  port: ${{PE_PORT}}
+                httpGet:
+                  path: /livez
+                  port: ${{GB_STATUS_PORT}}
+                initialDelaySeconds: 60
+                periodSeconds: 30
+                timeoutSeconds: 3
```
LalatenduMohanty, line 152: 3 as timeout seconds might be less. I wonder if we can increase this to 5 to 10 seconds to reduce the churn. but I am fine with starting with 3 and then increase if required.

LalatenduMohanty (approved):
/lgtm

PR #743, opened 2022-11-21
[e2e] add test for multi arch and update e2es for 134475077ed

PratikMahajan:


petr-muller:
/lgtm
/hold

Someone more experienced with Cincy tests should probably eyeball this too. Feel free to unhold as needed.

wking (approved):
/lgtm
/hold cancel

PR #747, opened 2022-12-06
increase liveness/readiness initial delay to 5min

PratikMahajan:
this is to accommodate the change in liveness and readiness 
conditions that go true after the graph scraping is complete


LalatenduMohanty (approved):
/lgtm

PR #750, opened 2022-12-15
Add rustfmt.toml to explicitly set the root edition to 2018

jeckersb:
Some IDEs/plugins do not proprely infer the edition from the workspace crates, and instead only check the root and default to 2015 since the root Cargo.toml does not have a project in which to declare an edition.

See: https://github.com/rust-lang/rustfmt/issues/4454#issuecomment-703315392


jeckersb:
Actually hold this for now, this only seems to half-fix the issue for me.  Needs more investigation.

jeckersb:
Nevermind, turns out I had two copies of rustfmt running in parallel; one with `edition` and one without `edition` and they were racing each other.  I fixed that and now it's able to figure out the edition automatically.  Sorry for the noise.

PR #751, opened 2022-12-16
add version information to graph-normalize

PratikMahajan:


petr-muller (approved):

petr-muller:
bah, we don't map gh review to lgtm here
/lgtm

PR #769, opened 2023-01-17
Serve current graph-data tar through cincinnati

PratikMahajan:
create a tar file after the unnecessary files
are removed from the graph-data directory.
And add the tar file path to internal_io parameters
which can then be used to update the secondary_metadata
state which we use while serving the tar file through a 
new endpoint and port

### DIFF
--- a/cincinnati/src/plugins/internal/graph_builder/dkrv2_openshift_secondary_metadata_scraper/plugin.rs
+++ b/cincinnati/src/plugins/internal/graph_builder/dkrv2_openshift_secondary_metadata_scraper/plugin.rs
@@ -1,6 +1,7 @@
 use crate as cincinnati;
 use crate::plugins::internal::dkrv2_openshift_secondary_metadata_scraper::gpg;
 use crate::plugins::internal::release_scrape_dockerv2::registry;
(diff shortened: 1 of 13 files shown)


cincinnati/src/plugins/internal/graph_builder/dkrv2_openshift_secondary_metadata_scraper/plugin.rs
```diff
@@ -276,8 +274,26 @@ impl InternalPlugin for DkrV2OpenshiftSecondaryMetadataScraperPlugin {
         })
         .await?;
 
+        let graph_data_dir = data_dir.path().to_path_buf();
         self.update_cache_state(layers, data_dir).await;
 
+        let graph_data_tar_path = self.settings.output_directory.join("graph-data.tar.gz");
```
LalatenduMohanty, line 280: Wondering if we should add timestamp to the graph-data.tar.gz file?

LalatenduMohanty: We might want the commit information somewhere in the tarball as it will be useful for debugging. But that cane be handled in a later PR.

LalatenduMohanty (approved):
/lgtm

PR #779, opened 2023-01-25
fix the allowlist regex for graph-data

PratikMahajan:
fix the regex to remote the internal-channels folder
and include the LICENCE


wking (approved):
/lgtm

PR #780, opened 2023-01-25
[dist] use path prefix for graph-builder

PratikMahajan:
this is needed to enable the graph-data endpoint


dist/openshift/cincinnati.yaml
```diff
@@ -243,7 +243,7 @@ objects:
       gb.rust_backtrace: "${RUST_BACKTRACE}"
       pe.address: "0.0.0.0"
       pe.status.address: "0.0.0.0"
-      pe.upstream: "http://localhost:8080/graph"
+      pe.upstream: "http://localhost:8080/api/upgrades_info/graph"
```
wking: Did we want to use`${GB_PATH_PREFIX}` in here too?

wking (approved):
/lgtm

PR #782, opened 2023-02-06
increase loglevel to debug!

PratikMahajan:
increase the loglevel for non-critical/expected
statements to debug!

### DIFF
--- a/cincinnati/src/plugins/internal/graph_builder/openshift_secondary_metadata_parser/plugin.rs
+++ b/cincinnati/src/plugins/internal/graph_builder/openshift_secondary_metadata_parser/plugin.rs
@@ -480,11 +480,11 @@ impl OpenshiftSecondaryMetadataParserPlugin {
                                         blocked_edge.from.to_string(),
                                     );
                                 }
-                                Err(e) => warn!("{}", e),
+                                Err(e) => debug!("{}", e),
                             };
                         }
                         None => {
-                            info!("Release with version {} not found in graph", to);
+                            debug!("Release with version {} not found in graph", to);
                         }
                     };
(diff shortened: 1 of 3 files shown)


cincinnati/src/plugins/internal/graph_builder/openshift_secondary_metadata_parser/plugin.rs
```diff
@@ -281,7 +281,7 @@ where
             Ok(yaml) => match serde_yaml::from_slice(&yaml) {
                 Ok(value) => t_vec.push(value),
                 Err(e) => {
-                    warn!("Failed to deserialize file at {:?}: {}", &path, e);
+                    debug!("Failed to deserialize file at {:?}: {}", &path, e);
```
LalatenduMohanty: This looks like a real issue. IMO it should be in warning. Though I have to check the current logs to see if we see this frequently.

LalatenduMohanty (changes requested):

cincinnati/src/plugins/internal/graph_builder/openshift_secondary_metadata_parser/plugin.rs
PratikMahajan: this happens when the blocking file is not a conditional-block. Old files without conditional block throw this warning,

LalatenduMohanty: This will be useful if we get conditional edges which fails to serialize.

LalatenduMohanty (approved):
/lgtm

PR #783, opened 2023-02-07
decrease loglevel from debug to info

PratikMahajan:
Print application settings and inform when the application 
able to take requests


LalatenduMohanty (approved):
/lgtm

PR #785, opened 2023-03-01
[dist] add deployment as an alternative to deploymentconfig

PratikMahajan:


LalatenduMohanty:
Looks good to me. However I am on the fence with respect to whether we should keep the deploymentconfig as it is and new files for deployment.

PratikMahajan:
kept DeploymentConfig as it is and introduced Deployment as replacement for DeploymentConfig. This is to allow us for a smooth transfer in between the configs. 
https://github.com/openshift/cincinnati/compare/eae79e98d92743e4f5b309927474d34c07591562..55ea56e9b4605190dfb81e30c1538505522b5148

LalatenduMohanty (approved):
/lgtm

PR #787, opened 2023-03-28
fix code errors

PratikMahajan:
remove unused code segments and implement default for AppState


LalatenduMohanty:
@PratikMahajan can we remove `[chore]` from the commit message. It is not adding any value.

LalatenduMohanty:
Also please some commit description to communicate why we need these changes.

petr-muller (approved):

PR #790, opened 2023-04-10
update dkregistry to include fix for next_page

PratikMahajan:
ensures that cincinnati works with registries that do not use a "next_page" query parameter


wking (approved):

PratikMahajan:
/override ci/prow/customrust-e2e
/override ci/prow/customrust-images 

will be fixed in #786

PR #793, opened 2023-05-10
Bump Rust to 2021 edition

petr-muller:
Seeing all the dependabot bumps that fail to pass tests with

```
error: failed to parse manifest at `/root/.cargo/registry/src/github.com-1ecc6299db9ec823/hermit-abi-0.3.1/Cargo.toml`
Caused by:
  feature `edition2021` is required 
```

I wonder if there is a good reason to not bump the Rust edition?


petr-muller:
Ah I start to understand how Rust tooling works! :rofl: 

/close

PR #794, opened 2023-05-18
OTA-915: Serve signatures through a new container

PratikMahajan:
This PR adds a new container to serve the signatures. 

Signatures should be added to the directory in order to
serve the signatures. 
The dir structure should look like `{ALGO}/{DIGEST}/{SIGNATURE}`

Add this directory to toml file in-order to read the signatures.
```
[signatures]
dir = <DIR-HERE>
```
In the absence of dir in toml file, it'll create a temp directory and 
read from it

### DIFF
--- a/commons/src/lib.rs
+++ b/commons/src/lib.rs
@@ -25,6 +25,7 @@ pub mod prelude_errors {
 }
 
 use actix_web::http::header::{HeaderMap, HeaderValue, ACCEPT};
+use actix_web::HttpRequest;
 use flate2::write::GzEncoder;
 use flate2::Compression;
 use std::collections::HashMap;
@@ -182,6 +183,44 @@ pub async fn create_tar(output_path: Box<Path>, data_path: Box<Path>) -> Result<
         return Err(prelude_errors::format_err!("{:?}", tar_status.unwrap_err()));
     }
     Ok(())
+}
+
+/// format the request before logging. Include only details that we need.
+pub fn format_request(req: &HttpRequest) -> String {
+    let no_user_agent = HeaderValue::from_str("user-agent not available").unwrap();
+    let no_accept_type = HeaderValue::from_str("accept value not available").unwrap();
+    let req_type = req.method().as_str();
+    let request = req.path();
+    let query = req.query_string();
+    let user_agent = req
+        .headers()
+        .get("user-agent")
+        .unwrap_or(&no_user_agent)
+        .to_str()
+        .unwrap();
+    let accept_type = req
+        .headers()
+        .get("accept")
+        .unwrap_or(&no_accept_type)
+        .to_str()
+        .unwrap();
+    format!(
+        "Method: '{}', Request: '{}', Query: '{}', User-Agent: '{}', Accept: '{}'",
+        req_type, request, query, user_agent, accept_type
+    )
+}
+
+/// logs api request error
+pub fn api_response_error(req: &HttpRequest, e: GraphError) -> GraphError {
+    log::error!(
+        "Error serving request \"{}\" from '{}': {:?}",
+        format_request(req),
+        req.peer_addr()
+            .map(|addr| addr.to_string())
+            .unwrap_or_else(|| "<not available>".into()),
+        e
+    );
+    e
 }
 
 #[cfg(test)]
--- a/metadata-helper/Cargo.toml
+++ b/metadata-helper/Cargo.toml
@@ -1,0 +1,39 @@
+[package]
+name = "metadata-helper"
+version = "0.1.0"
+authors = ["Pratik Mahajan <pmahajan@redhat.com>"]
+edition = "2021"
+build = "src/build.rs"
+
+[dependencies]
+actix = "0.13.0"
+actix-cors = "^0.6.1"
+actix-service = "2.0.2"
+actix-web = "^4.0.0-rc.3"
+cincinnati = { path = "../cincinnati" }
+commons = { path = "../commons" }
+custom_debug_derive = "^0.5"
+env_logger = "^0.9"
+futures = "^0.3"
+hyper = "^0.14"
+lazy_static = "^1.2.0"
+log = "^0.4.17"
+opentelemetry = "0.14.0"
+parking_lot = "^0.12"
+prometheus = "0.13"
+semver = { version = "^1.0.16", features = [ "serde" ] }
+serde = "^1.0.152"
+serde_derive = "^1.0.70"
+serde_json = "^1.0.91"
+smart-default = "^0.6"
+structopt = "^0.3"
+toml = "^0.5"
+url = "^2.3"
+
+[build-dependencies]
+built = { version = "^0.5.1", features = [ "git2" ]}
+
+[dev-dependencies]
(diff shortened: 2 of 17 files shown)


metadata-helper/src/status.rs
```diff
@@ -0,0 +1,30 @@
...
+    } else {
+        HttpResponse::ServiceUnavailable().finish()
+    }
+}
+
+/// Expose readiness status.
+///
+/// Status:
+///  * Ready (200 code): the application has been initialized and is available to accept connections.
+///  * Not Ready (503 code): no JSON graph available yet.
```
wking: I'm fuzzy on what `app_data.is_ready` is actually  about, but the blocking issue for serving signatures out of this process is whether the local signature filesystem is present.  With init-container data sourcing, that's true from before this container comes up.  In a future where one of these containers is trying to poll graph-data image from the registry, it would be true after the first successful retrieval and unpacking.  Neither of those are directly about the JSON graph (that graph-builder is serving?).

metadata-helper/src/signatures.rs
```diff
@@ -0,0 +1,119 @@
...
+    let timer = SIGNATURES_SERVE_HIST.start_timer();
+
+    let params = req.match_info();
+    let algo = params.get("ALGO").unwrap();
+    let digest = params.get("DIGEST").unwrap();
+    let signature = params.get("SIGNATURE").unwrap();
+
+    let signatures_data_path = app_data.signatures_dir.clone();
+    let mut signature_path = PathBuf::from(signatures_data_path);
+    signature_path.push(&format!("{}/{}/{}", algo, digest, signature));
```
wking, line 75: Looking at RFC 3986's sections [3.3][1] and [5.2.4][2], I don't think we need to worry about `..` path segments being used to access locations outside of `signatures_data_path`, but it's probably worth testing just to make sure.

[1]: https://www.rfc-editor.org/rfc/rfc3986#section-3.3
[2]: https://www.rfc-editor.org/rfc/rfc3986#section-5.2.4

petr-muller:
Do we plan to address the issues with deps with rust 1.54? If not, should we remove the jobs? If yes, why do we need to support 1.54? :)

commons/src/lib.rs
```diff
@@ -184,6 +185,31 @@ pub async fn create_tar(output_path: Box<Path>, data_path: Box<Path>) -> Result<
     Ok(())
 }
 
+/// format the request before logging. Include only details that we need.
+pub fn format_request(req: &HttpRequest) -> String {
+    let no_user_agent = HeaderValue::from_str("user-agent not available").unwrap();
+    let no_accept_type = HeaderValue::from_str("Accept value unavailable").unwrap();
```
petr-muller: nit: make the messaging consistent, we say "not available" for one header and "value unavailable" for the other

metadata-helper/src/signatures.rs
```diff
@@ -0,0 +1,101 @@
...
+    error!(
+        "Error serving request \"{}\" from '{}': {:?}",
+        format_request(req),
+        req.peer_addr()
+            .map(|addr| addr.to_string())
+            .unwrap_or_else(|| "<not available>".into()),
+        e
+    );
+    e
+}
```
petr-muller: We have identical code in `policy-engine/src/graph.rs`, would it make sense to also extract this to where `format_request` was extracted and import `api_response_error` to be used in both places?

metadata-helper/src/status.rs
```diff
@@ -0,0 +1,30 @@
...
+/// Status:
+///  * Ready (200 code): the application has been initialized and is available to accept connections.
+///  * Not Ready (503 code): no signatures available yet.
+pub async fn serve_readiness(app_data: actix_web::web::Data<AppState>) -> HttpResponse {
+    if app_data.is_ready() {
+        HttpResponse::Ok().finish()
+    } else {
+        HttpResponse::ServiceUnavailable().finish()
+    }
+}
```
petr-muller, line 30: This is a duplicate of `policy-engine/src/status.rs`, would it make sense to extract these simple probe backends to a helper that would simply take `app_data: actix_web::web::Data<AppState>`  as a parameter?

metadata-helper/src/config/cli.rs
```diff
@@ -0,0 +1,89 @@
+//! Command-line options for policy-engine.
```
petr-muller: copy-pasta artifact ;)

metadata-helper/src/config/settings.rs
```diff
@@ -0,0 +1,88 @@
+//! Application settings for policy-engine.
```
petr-muller: copy-pasta artifact

metadata-helper/src/signatures.rs
```diff
@@ -0,0 +1,101 @@
...
+
+/// Register relevant metrics to a prometheus registry.
+pub(crate) fn register_metrics(registry: &Registry) -> Fallible<()> {
+    commons::register_metrics(registry)?;
+    registry.register(Box::new(SIGNATURES_INCOMING_REQS.clone()))?;
+    registry.register(Box::new(SIGNATURES_SERVE_HIST.clone()))?;
+    Ok(())
+}
+
+/// Serve Cincinnati graph requests.
```
petr-muller: Does "graph requests" represent well what do we do here? "signature data requests" would be better perhaps?

petr-muller (commented):
I'm very green in reviewing Rust code, so take the comments with a grain of salt and feel free to ignore them, I'm not yet familiar with Rust idioms and accepted practice :)

In general there seems to be quite a lot of duplicate code around the processing of parameters (merging params from cli and config file etc) that stems from how this was apparently created by making a gutted copy of policy-engine. It feels like we could DRY these pieces now when we have two or even three copies of them?

metadata-helper/src/status.rs
PratikMahajan: the <AppState> struct, though it has same name, the struct is a bit different in the packages, thats the reason for duplicate code

metadata-helper/src/config/cli.rs
```diff
@@ -0,0 +1,89 @@
...
+    #[structopt(short = "c")]
+    pub config_path: Option<String>,
+
+    // Status service options
+    #[structopt(flatten)]
+    pub service: options::ServiceOptions,
+
+    // Main service options
+    #[structopt(flatten)]
+    pub status: options::StatusOptions,
```
wking, line 25: I'm fuzzy  on details, but `Status service options` for `service` and `Main service options` for `status` seems like the comments may be flipped here?

petr-muller (approved):
/lgtm
/hold

I'm not the greatest Rust reviewer ever, hence holding for Trevor or Lala to have a chance if you want them to review. Feel free to unhold at will.

metadata-helper/Cargo.toml
```diff
@@ -2,12 +2,13 @@
 name = "metadata-helper"
 version = "0.1.0"
 authors = ["Pratik Mahajan <pmahajan@redhat.com>"]
-edition = "2021"
+edition = "2018"
```
LalatenduMohanty, line 5: Is it an intentional change?

PR #813, opened 2023-07-12
Dockerfiles: Use UBI to fetch vegeta and stop pretending we build Go

petr-muller:
We just `curl` the vegeta binary from a release so we can do that using the same image like we use to build Rust, avoid downloading another image and simplify maintenance of these dockerfiles.


petr-muller:
/hold 
wtf

petr-muller:
https://github.com/openshift/ci-tools/pull/3457 broke us, nothing to see here, move along

petr-muller:
/test images
/hold cancel
:monocle_face:

PratikMahajan (approved):
/lgtm

PR #816, opened 2023-07-17
OTA-915: fix configs and add signature serving deployment

PratikMahajan:
this PR fixes the existing configs/settings, removes unused imports
and adds a sample deployment yaml that deploys the metadata-helper
container along with graph-builder and policy-engine. 
deploying metadata-helper is optional and is used to serve existing signatures

### DIFF
--- a/Justfile
+++ b/Justfile
@@ -157,8 +157,6 @@ run-metadata-helper:
 		verbosity = "vvv"
 
 		[service]
-		scrape_timeout_secs = 300
-		pause_secs = {{pause_secs}}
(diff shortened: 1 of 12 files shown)


petr-muller (approved):
/lgtm
/override ci/prow/e2e
We passed all tests but then failed in a `post` must-gather step

/hold
Holding to allow Lala / Trevor to take a look, but feel free to unhold if you think it is not necessary.

LalatenduMohanty (approved):
/lgtm

PR #837, opened 2023-08-10
OTA-1005: dkrv2: Add new option graph_data_path

jeckersb:
This enables support for classic (UBI-based) as well as flat graph
data images.

By default, graph_data_path will be "/", which retains
backward-compatibility with the current state which assumes a flat
graph data image.

If using a classic UBI-based graph data image, instead the data is
typically rooted at /var/lib/cincinnati-graph-data.  Pointing
graph_data_path to this directory allows proper generation of the
graph data tarball, as well as correctly pointing the
openshift-secondary-metadata-parse plugin to the correct location
within the output directory.


PratikMahajan (approved):
/lgtm

PR #840, opened 2023-08-11
add sha256 as supported algo for metadata-helper

PratikMahajan:
most of the signatures served by metadata-helper use `sha256` algorithm. 
this PR basically adds support for `sha256` signatures. 
Earlier we used to reject requests citing `algo not supported`


LalatenduMohanty (approved):
/lgtm

PR #845, opened 2023-08-15
use correct signatures URI and other fixes

PratikMahajan:
use the correct URI for signatures. Previously we used URI that was meant to be used for file system.

Also adds docs about running metadata-helper using just command.

Removes unnecessary information transferred through error message, now just states which signature does not exist.

### DIFF
--- a/metadata-helper/src/signatures.rs
+++ b/metadata-helper/src/signatures.rs
@@ -58,8 +58,18 @@ async fn _index(
     let timer = SIGNATURES_SERVE_HIST.start_timer();
 
     let params = req.match_info();
-    let algo = params.get("ALGO").unwrap();
-    let digest = params.get("DIGEST").unwrap();
+    let algo_digest_string = params.get("ALGO_DIGEST").unwrap();
+    let algo_digest: Vec<&str> = algo_digest_string.clone().split('=').collect();
+
+    if algo_digest.len() != 2 {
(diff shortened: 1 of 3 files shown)


metadata-helper/src/signatures.rs
```diff
@@ -78,8 +88,8 @@ async fn _index(
     if f.is_err() {
         timer.observe_duration();
         return Err(GraphError::DoesNotExist(format!(
-            "signature does not exist {}",
-            f.unwrap_err()
+            "signature {}",
+            sig
```
wking: you probably want to include `{SIGNATURE}` too, maybe via `signature {}/{}`?

metadata-helper/src/signatures.rs
```diff
@@ -58,8 +58,18 @@ async fn _index(
     let timer = SIGNATURES_SERVE_HIST.start_timer();
 
     let params = req.match_info();
-    let algo = params.get("ALGO").unwrap();
-    let digest = params.get("DIGEST").unwrap();
+    let sig = params.get("ALGO_DIGEST").unwrap();
```
wking: You have `signature` for the `{SIGNATURE}` portion of the URI.  Maybe use `algo_digest_string` or something for this variable, to avoid confusion with the `{SIGNATURE}` parameter?

wking (approved):
/lgtm

PR #887, opened 2024-01-26
add support for root certificates

PratikMahajan:
uses all the certificates in the directory to make calls. 
this will ensure that, if certificates exist, we'll use them.

### DIFF
--- a/cincinnati/src/plugins/internal/graph_builder/commons.rs
+++ b/cincinnati/src/plugins/internal/graph_builder/commons.rs
@@ -1,3 +1,60 @@
+use std::io::Read;
+use std::path::{Path, PathBuf};
+
+use actix_web::Error;
+use commons::Fallible;
+use log::debug;
+use reqwest::Certificate;
+
+/// Retrives all the .pem and .crt certificate files from the directory.
+/// Scans till depth of 2, any file/directory below the depth of 2 will be ignored.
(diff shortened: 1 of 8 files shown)


cincinnati/src/plugins/internal/graph_builder/commons.rs
```diff
@@ -1,3 +1,60 @@
...
+    let mut certs: Vec<Certificate> = Vec::new();
+    let mut dirs: Vec<(PathBuf, u8)> = vec![(dir.to_path_buf(), 0)];
+    while !dirs.is_empty() {
+        let (dir, level) = dirs.pop().unwrap();
+        if let Ok(cert_paths) = std::fs::read_dir(dir) {
+            for path in cert_paths {
+                if let Ok(path) = path {
+                    if path.metadata().unwrap().is_file() {
+                        match path.path().extension() {
+                            Some(extension) => {
```
petr-muller, line 21: I'm not a fan of conditionals nested this deep but I don't see an obvious way to improve it significantly to be worth the hassle so let's keep it I guess, given it implements a rather straightforward thing :shrug:

petr-muller (approved):
LGTM, holding to allow more reviews if needed

/hold

LalatenduMohanty (approved):
/lgtm

PR #902, opened 2024-04-01
OTA-1045: Forwards & backwards compatibility for graph-data

PratikMahajan:


cincinnati/src/plugins/internal/graph_builder/openshift_secondary_metadata_parser/plugin.rs
```diff
@@ -6,11 +6,14 @@ use self::cincinnati::plugins::prelude_plugin_impl::*;
 
 use crate::conditional_edges::{ConditionalEdge, ConditionalUpdateEdge, ConditionalUpdateRisk};
 use commons::GRAPH_DATA_DIR_PARAM_KEY;
+use std::cmp::Ordering;
 use std::collections::HashSet;
 use std::path::Path;
 
 pub static DEFAULT_KEY_FILTER: &str = "io.openshift.upgrades.graph";
 static SUPPORTED_VERSIONS: &[&str] = &["1.0.0", "1.1.0", "1.2.0"];
```
wking: Can we drop this now?  I'd expect the SemVer-based `MAX_SUPPORTED_VERSION` logic to cover this use-case, and having two things that need bumping when we grow additional support seems like more work than shifting entirely to the new flow and only bumping `MAX_SUPPORTED_VERSION`.

cincinnati/src/plugins/internal/graph_builder/openshift_secondary_metadata_parser/plugin.rs
```diff
@@ -343,14 +346,29 @@ impl OpenshiftSecondaryMetadataParserPlugin {
...
+            // Utilize MIN_SEMVER_VERSION and MAX_SEMVER_VERSION as fallbacks for unsupported semver versions,
+            // avoiding iteration over the semver set to determine min/max versions due to the high cost of such operations.
+            // For unsupported semver versions, return the nearest supported semver version to the requested one.
+            let req_version = semver::Version::parse(&string_version).unwrap();
+            let resp_version =
+                match req_version.cmp(&semver::Version::parse(MAX_SUPPORTED_VERSION).unwrap()) {
+                    // the requested version is not in supported versions, means its either less than MIN_SUPPORTED or
+                    // greated than MAX_SUPPORTED, if we hit the `less` case here, it means that the version should be
+                    // less than MIN_SUPPORTED version, so return the MIN_SUPPORTED_VERSION
+                    Ordering::Less => MIN_SUPPORTED_VERSION.to_owned(),
```
wking: I'm not sure we need a `MIN_SUPPORTED_VERSION`, e.g. Cincinnati that understands 1.2.0 can happily serve 1.0.0 and 1.1.0 graph-data still.  I was expecting logic like [this][1], where we compare the major.minor portions that we read in with `MAX_SUPPORTED_VERSION` and:

* If the graph-data major version doesn't match `MAX_SUPPORTED_VERSION`'s major version, error out, complaining of an unrecognized version (`unrecognized graph-data version 2.0.0; supported version: 1.2.0 and compatible`, or similar).
* If the graph-data major does match `MAX_SUPPORTED_VERSION`'s major version:
  * If the graph-data minor version exceeds `MAX_SUPPORTED_VERSION`'s minor version, log a warning (`graph-data version 1.3.0 may have features which our 1.2.0 implementation does not support`, or similar).
  * If the graph-data minor version is <= `MAX_SUPPORTED_VERSION`'s minor version, log a happy message (`graph-data version 1.1.0 is compatible with our 1.2.0 implementation`, or similar).

[1]: https://github.com/openshift/cincinnati-graph-data/tree/9325158cefa894ae808be16639019b5326b90ab2?tab=readme-ov-file#schema-version

cincinnati/src/plugins/internal/graph_builder/openshift_secondary_metadata_parser/plugin.rs
```diff
@@ -343,14 +344,30 @@ impl OpenshiftSecondaryMetadataParserPlugin {
...
+
+        if req_version.minor > max_version.minor {
+            info!("graph-data version {} may have features which our {:?} implementation does not support", string_version, MAX_SUPPORTED_VERSION);
+            return Ok(MAX_SUPPORTED_VERSION.to_string());
+        } else {
+            info!(
+                "graph-data version {} is compatible with our {:?} implementation",
+                string_version, MAX_SUPPORTED_VERSION
+            );
+            return Ok(string_version.to_string());
```
wking: Looks like our happy-case used to be `Ok(string_version.into_owned())`, and now it's `Ok(string_version.to_string())`?  I'm not sure what the trade-offs are there, or if the two methods are basically synonyms.

cincinnati/src/plugins/internal/graph_builder/openshift_secondary_metadata_parser/plugin.rs
```diff
@@ -343,14 +344,30 @@ impl OpenshiftSecondaryMetadataParserPlugin {
...
-            Err(format_err!(
-                "unrecognized graph-data version {}; supported versions: {:?}",
+        if semver::Version::parse(&string_version).is_err() {
+            return Err(format_err!("{} is not a valid semver", string_version));
+        }
+
+        let req_version = semver::Version::parse(&string_version).unwrap();
+        let max_version = semver::Version::parse(MAX_SUPPORTED_VERSION).unwrap();
+
+        if req_version.major > max_version.major || req_version.major < max_version.major {
```
wking: nit: instead of >/<, you could probably use a more compact !=?

wking (approved):
/lgtm

PR #919, opened 2024-05-15
OCPBUGS-25833: strict filtering of BlockedEdges(old_blocks)

PratikMahajan:
strictly filter the old blocks so that conditional blocks are handled in the error section the way we were handling old blocks while parsing conditional edges.

A blocked edge will now appear either in old_blocks or conditional_blocks. Earlier all the conditional edges used to appear in the old_blocks as well.


cincinnati/src/plugins/internal/graph_builder/openshift_secondary_metadata_parser/plugin.rs
```diff
@@ -23,6 +23,7 @@ pub mod graph_data_model {
     use std::collections::HashMap;
     /// Represents the blocked edges files in the data repository.
     #[derive(Debug, Deserialize)]
+    #[serde(deny_unknown_fields)]
```
petr-muller, line 26: This seems to be a well-hidden key bit of the PR - this is what actually implements the strict filtering. The code change below just adjusts some of the related error detection behavior, not the filtering.

Do I understand correctly that without this, we previously successfully parsed all conditional edge files as blocked edges, too?

Blocked edge:
```yaml
to: 4.10.0-rc.0
from: 4\.9\..*
```

Conditional edge:
```yaml
to: 4.10.13
from: 4[.]10[.]3[+].*
url: https://bugzilla.redhat.com/show_bug.cgi?id=2076312#c9
name: CephParallelFsync
message: ...
matchingRules:
- type: Always
```

So without strict parsing, every conditional edge is a blocked edge too, and therefore we kinda depended on [OCPBUGS-25833](https://issues.redhat.com/browse/OCPBUGS-25833) behavior to even function correctly. If we blocked all blocked edge detected without strict parsing, we would block all conditional edge edges.

Do I get this right?

cincinnati/src/plugins/internal/graph_builder/openshift_secondary_metadata_parser/plugin.rs
```diff
@@ -205,11 +206,6 @@ where
     use tokio_stream::wrappers::ReadDirStream;
     use tokio_stream::StreamExt;
 
-    let mut is_conditional_edge = false;
-    if TypeId::of::<T>() == TypeId::of::<graph_data_model::ConditionalEdgeYaml>() {
-        is_conditional_edge = true;
-    }
```
petr-muller, line 211: OK, I think I get this. This tested whether "are we searching for conditional edges" by type-checking the generic type `T`. The `is_conditional_edge` name is misleading, it suggests that conditional edge is what we found, not what we are searching for.

cincinnati/src/plugins/internal/graph_builder/openshift_secondary_metadata_parser/plugin.rs
```diff
@@ -288,19 +284,13 @@ where
...
-                        // blocked-edge, so not committing error. It'll be considered as blocked-edge
-                        // till we're able to parse conditional-edge file without errors.
-                        // we'll throw an error if the file is not able to match blocked-edge
-                        // as well as conditional-edge
-                        old_block = match serde_yaml::from_slice(&yaml) {
-                            Ok(value) => value,
-                            Err(_e) => None,
-                        };
-                    }
-                    if old_block.is_none() {
```
petr-muller, line 303: OK, so previously what this did is: 1) if we failed to parse the file for whatever we are parsing for (`T`)  _and_ 2) we expected conditional edge, we try to parse as a blocked edge instead, and if we succeed, we do not fail because we place both types of edges into the same directory.

Other cases:
- if we parse for blocked edges then we never fail on conditional edges without strict
- if we parse for anything we fail on stuff thats neither a blocked or conditional edge -> this what we need to error on

cincinnati/src/plugins/internal/graph_builder/openshift_secondary_metadata_parser/plugin.rs
```diff
@@ -288,19 +284,13 @@ where
...
+                    // conditional-edge file threw error while parsing but matched blocked-edge or
+                    // blocked-edge file threw an error and matched conditional edge, dont throw a warning, these are expected
+                    // and doesnt need warning as at the end of the process, it'll match atleast one block.
+                    // if its neither blocked edge nor conditional edge, throw a warning to make sure we log this as an issue.
+                    if !(TypeId::of::<T>() == TypeId::of::<graph_data_model::ConditionalEdgeYaml>())
+                        && !(TypeId::of::<T>() == TypeId::of::<graph_data_model::BlockedEdge>())
+                    {
```
petr-muller, line 299: I don't think this logic is entirely correct then. This says "if the method is trying to find one of these two types, then do not warn". I think the check needs to be done over what was actually found - what we succeed to parse.

Something like: If we are searching for one of these two types, and it failed (this part you have in the PR already), then try to parse it with the other type and if it succeeds (this part you dont), do not error out.

petr-muller: We still need to check `T`, because we use this method to also parse channels in https://github.com/openshift/cincinnati/blob/58d65c7a7e242d2d3e7820965afc501d0e955270/cincinnati/src/plugins/internal/graph_builder/openshift_secondary_metadata_parser/plugin.rs#L587 and only "parses as" check would override errors if somehow a blocked edge ends in channel directory.

I think we need a check like this:

if 
(T is BlockedEdge and yaml does not parse as ConditionalEdge)
or
(T is ConditionalEdge and yaml does not parse as Blocked)
then error

petr-muller (commented):
The PR also panics in tests:

```
stack backtrace:
   0: rust_begin_unwind
   1: core::panicking::panic_fmt
   2: core::panicking::panic_display
             at /builddir/build/BUILD/rustc-1.71.1-src/library/core/src/panicking.rs:150:5
   3: e2e::e2e_channel_success::e2e_channel_success
             at /opt/app-root/src/e2e/tests/e2e.rs:60:9
   4: e2e::e2e_channel_success::_stable_4_7_s390x_application_
             at /opt/app-root/src/e2e/tests/e2e.rs:15:1
   5: e2e::e2e_channel_success::_stable_4_7_s390x_application_::{{closure}}
             at /opt/app-root/src/e2e/tests/e2e.rs:15:1
   6: core::ops::function::FnOnce::call_once
             at /builddir/build/BUILD/rustc-1.71.1-src/library/core/src/ops/function.rs:250:5
```

cincinnati/src/plugins/internal/graph_builder/openshift_secondary_metadata_parser/plugin.rs
PratikMahajan: this is correct

petr-muller (approved):
LGTM, this was a good catch. I'm really curious about the incoming [OTA-1043](https://issues.redhat.com/browse/OTA-1043) fix this change enables :+1:

PR #920, opened 2024-06-05
OTA-1267: ignore incorrect images instead of erroring out

PratikMahajan:
we're ignoring the images which we do not understand with this change. this change is required because we dont want cincinnati to error out on encountering cosign signatures.
unintended consequence of this change can be that cincinnati continues with a log when it encounters incorrect image whereas it should error out. Cincinnati won't include this image in the update graph. We should ideally try to ignore only the cosign .sig signatures. this change acts a stopgap till we teach cincinnati to deal with signatures

log like this will be printed when we encounter a incorrect images 
```log
[2024-06-05T20:04:15Z WARN  cincinnati::plugins::internal::graph_builder::release_scrape_dockerv2::registry] error fetching manifest and manifestref for <>/<>-releases:sha256-7c42e086007d868f8295692.sig: fetching manifest and manifestref for <>/<>-releases:sha256-7c42e086007d868f8295692.sig: unknown media type ManifestV2S1, ignoring this image
```

### DIFF
--- a/cincinnati/src/plugins/internal/graph_builder/release_scrape_dockerv2/registry/mod.rs
+++ b/cincinnati/src/plugins/internal/graph_builder/release_scrape_dockerv2/registry/mod.rs
@@ -322,7 +322,26 @@ pub async fn fetch_releases(
 
         async move {
             let (arch, manifestref, mut layers_digests) =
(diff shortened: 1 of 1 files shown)


wking (approved):
/lgtm

PR #921, opened 2024-06-06
OTA-1043: remove conditional edges from normal edges

PratikMahajan:
if a conditional edge exists, remove it from normal edges
this way it becomes more predictable to code applications
which depend on cincinnati graph.
after this commit merges, an edge will either exist in normal
edge, i.e edge with no issues
or
it'll exist in conditional edges, i.e edges with known issues.


wking (approved):
/lgtm

PR #931, opened 2024-08-28
OCPBUGS-18280: Rename instances of TRUSTED CERTIFICATE to CERTIFICATE in non standard certs

PratikMahajan:
Rename instances of TRUSTED CERTIFICATE to CERTIFICATE so that these certificates are treated as normal certs and accepted by the application. 
Libraries do not support "TRUSTED CERTIFICATE" keyword in certs

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -112,7 +112,7 @@ dependencies = [
  "mime",
  "percent-encoding",
  "pin-project-lite",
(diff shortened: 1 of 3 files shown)


petr-muller:
/hold

Discussion ongoing on [Slack](https://redhat-internal.slack.com/archives/CJ1J9C3V4/p1725003462803479)

PR #932, opened 2024-08-28
OTA-134: remove deploymentconfig from cincinnati

PratikMahajan:
deploymentconfig has been deprecated and we no longer use it. use cincinnati-deployment.yaml as a replacement to deploymentconfig


petr-muller (approved):
LGTM

petr-muller:
Failures look like the known HEAD fails investigated by David, should be safe to override

petr-muller:
/hold

Meh, the retester starts to kick-in when I LGTM, so holding, unhold as you override

PratikMahajan:
/override ci/prow/cargo-test

overriding known test failures

PR #935, opened 2024-09-02
NO-JIRA: dist/Dockerfile.e2e-ubi8/Dockerfile: Fix the `oc` binary to be compatible with the `ubi:8` base image 

DavidHurta:
Fix the `oc` binary downloaded for CI tests to be compatible with the
used base image `ubi8`.

From the release notes of OCP 4.16 [1]:
```
...With this release, the default version of OpenShift CLI (oc) is
compiled with Red Hat Enterprise Linux (RHEL) 9 and works properly when
running a cluster with FIPS enabled on RHEL 9. Additionally, a version
of oc compiled with RHEL 8 is also provided, which must be used if you
are running a cluster with FIPS enabled on RHEL 8. (OCPBUGS-23386,
OCPBUGS-28540)
```

The amd64 architecture is chosen as the vegeta binary is as well
downloaded for amd64 machines in the same Dockerfile. The
configuration [2] for the e2e test ensures the architecture to be amd64.

[1] https://docs.openshift.com/container-platform/4.16/release_notes/ocp-4-16-release-notes.html#ocp-4-16-known-issues_release-notes
[2] https://github.com/openshift/release/blob/5a96e1d5df9bd8d98b1e1c92a92f4e1841cf445d/ci-operator/config/openshift/cincinnati/openshift-cincinnati-master.yaml#L130

### DIFF
--- a/dist/Dockerfile.e2e-ubi8/Dockerfile
+++ b/dist/Dockerfile.e2e-ubi8/Dockerfile
@@ -23,7 +23,7 @@ WORKDIR "${HOME}/cincinnati"
 
 # Get oc CLI
 RUN mkdir -p ${HOME}/bin && \
-    curl -L https://mirror.openshift.com/pub/openshift-v4/clients/ocp/stable/openshift-client-linux.tar.gz 2>/dev/null | tar xzf - -C "${HOME}/bin/" oc
+    curl -L https://mirror.openshift.com/pub/openshift-v4/clients/ocp/stable/openshift-client-linux-amd64-rhel8.tar.gz 2>/dev/null | tar xzf - -C "${HOME}/bin/" oc
 ENV PATH="${PATH}:${HOME}/bin"
 
 COPY --from=rust_builder /opt/cincinnati/bin/e2e /usr/bin/cincinnati-e2e-test


petr-muller:
Alternatively, can we use ubi9 instead?

DavidHurta:
> Alternatively, can we use ubi9 instead?

I think we should use ubi9. Right now, I think it's more important to get the CI properly working to unblock any PRs and check its state. However, we should probably migrate to the ubi9 base image soon. Whether right now or in the upcoming weeks is another question.

DavidHurta:
The run https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cincinnati/935/pull-ci-openshift-cincinnati-master-e2e/1830662795826302976 passed.

However, the [e2e test logs](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_cincinnati/935/pull-ci-openshift-cincinnati-master-e2e/1830662795826302976/artifacts/e2e/openshift-e2e-test/build-log.txt) contain:
```
This file contained potentially sensitive information and has been removed.
```

Not sure whether that's expected for this job :eyes:

petr-muller:
> Right now, I think it's more important to get the CI properly working to unblock any PRs and check its state.

I would agree if this was a product base image migration but the `Dockerfile.e2e-ubi8` name indicates this dockerfile is only used in tests. Therefore I think this is the case where it's worth using the breakage as a motivation to eliminate some technical debt instead of just quickly patching it.

petr-muller (approved):
Seems like we don't feel like we have enough slack at this time to invest more effort into this, so let's just turn CI green for now

PR #936, opened 2024-09-05
OTA-1338: dist/cargo_test.sh: Remove coverage, create junits

petr-muller:
It does not seem that code coverage was actually used all, but it made the `cargo_test.sh` unnecessarily complicated by a search for individual test binaries that were then executed separately. The complexity of the bash involved caused us to not even notice when the tests actually failed to compile.

Therefore, remove the coverage-related bash and use Nextest (https://nexte.st) as a test runner. We did not use coverage, but we do use Prow, but because our tests never produced JUnit artifacts that Prow could display, we had to invest the raw test output when tests failed. If we start using tooling that produces JUnit XMLs, our CI jobs can report the individual tests that failed together with their associated output.

When run locally, the script falls back to standard runner when Nextest is not present. In CI, the script fails without Nextest (we do not want CI to silently stop producing JUnits).

### DIFF
--- a/dist/cargo_test.sh
+++ b/dist/cargo_test.sh
@@ -7,7 +7,8 @@ cargo_test_flags["cincinnati"]="--features test-net"
 cargo_test_flags["commons"]=""
 cargo_test_flags["graph-builder"]="--features test-net"
 cargo_test_flags["policy-engine"]=""
-# TODO: Are we missing metadata-helper and rh-manifest-generator here?
+cargo_test_flags["metadata-helper"]=""
+cargo_test_flags["rh-manifest-generator"]=""
 cargo_test_flags["prometheus-query"]=""
 cargo_test_flags["quay"]="--features test-net"
 
--- a/dist/cargo_test.sh
+++ b/dist/cargo_test.sh
@@ -7,6 +7,7 @@ cargo_test_flags["cincinnati"]="--features test-net"
 cargo_test_flags["commons"]=""
 cargo_test_flags["graph-builder"]="--features test-net"
 cargo_test_flags["policy-engine"]=""
+# TODO: Are we missing metadata-helper and rh-manifest-generator here?
 cargo_test_flags["prometheus-query"]=""
 cargo_test_flags["quay"]="--features test-net"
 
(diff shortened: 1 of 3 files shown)


petr-muller:
/hold

The script installs `cargo2junit` which should move to the image build

DavidHurta:
So that's how JUnits are made, cool :cool: 
/cc

petr-muller:
Submitted https://github.com/openshift/release/pull/56404 to install cargo2junit in the underlying image, but I wonder if using the [nextest runner](https://nexte.st/docs/machine-readable/junit/) could be a better option?

petr-muller:
nextest runs too few tests compared to standard `cargo test`, weird

dist/cargo_test.sh
```diff
@@ -23,52 +24,20 @@ declare -A executors
...
-              --exclude-pattern=$HOME/.cargo \
-              --verify \
-              ${CARGO_TARGET_DIR}/cov \
-              \$test
-          else
-            \$test
-          fi
-        done
-      "; then
+    if ! cargo nextest run ${cargo_test_flags[$directory]} \ --package "${directory}" --profile ci ; then
```
DavidHurta: Addressing the https://github.com/openshift/cincinnati/pull/936#issuecomment-2334239524 comment. 

I am not sure what the `\` slash's purpose is here.

In the logs we get (note that the `--package` flag is in apostrophes):
```
$ grep -E "cargo nextest run" build-log.txt 
+ cargo nextest run --features test-net,test-net-private ' --package' quay --profile ci
+ cargo nextest run ' --package' prometheus-query --profile ci
+ cargo nextest run ' --package' policy-engine --profile ci
+ cargo nextest run --features test-net,test-net-private ' --package' graph-builder --profile ci
+ cargo nextest run --features test-net,test-net-private ' --package' cincinnati --profile ci
+ cargo nextest run ' --package' commons --profile ci
```

Locally running the PR changes:
```
$ grep -Ei "Summary" PR_log.log 
     Summary [   0.000s] 0 tests run: 0 passed, 130 skipped
     Summary [   0.051s] 10 tests run: 3 passed, 7 failed, 132 skipped
     Summary [   0.000s] 0 tests run: 0 passed, 142 skipped
     Summary [   0.000s] 0 tests run: 0 passed, 130 skipped
     Summary [   0.000s] 0 tests run: 0 passed, 130 skipped
     Summary [  66.577s] 4 tests run: 3 passed, 1 failed, 138 skipped
```

By removing the slash from the script we get more tests running:
```
$ grep -Ei "Summary" PR_with_change_log.log 
     Summary [   0.003s] 1 test run: 1 passed, 0 skipped
     Summary [  67.315s] 73 tests run: 65 passed, 8 failed, 3 skipped
     Summary [   0.012s] 8 tests run: 8 passed, 0 skipped
     Summary [   0.057s] 14 tests run: 14 passed, 0 skipped
     Summary [   0.006s] 6 tests run: 6 passed, 0 skipped
     Summary [   0.966s] 2 tests run: 2 passed, 0 skipped
```

Davoska: I understand that it's an escape character; however I am not sure why

petr-muller: Good catch!

> I am not sure what the \ slash's purpose is here.

Just a leftover of multiline command; I collapsed the line when it shortened but the unfortunate `\` stayed there, resulting in this funny interaction where whatever flag parser is inside does not recognize `' --flag'` as a flag :D

petr-muller:
Needs https://github.com/openshift/release/pull/56928 to merge first. Until that PR merges, we are expecting CI to fail fast because it does not find Nextest. Once it merges, we need to retest here and expect tests to be executed and failing (unclear how many we expect, we saw ~9 failures before, but we do not know whether our failures are deterministic). We expect the JUnit XMLs to be produced and picked up by Prow. When that happens, we can override the failures and merge this PR to improve CI signal. Then we followup by fixing the tests.

petr-muller:
JUnits looking good!

DavidHurta:
The PR looks great, and the JUnits even better 🎉
/lgtm

The `/hold` is no longer needed.
/unhold

PratikMahajan:
these 13 failing JUnits are failing for me locally as well. Need to fix these tests

petr-muller:
/override ci/prow/cargo-test

These tests are failing on `master` as well, this PR just surfaces the failures better

PR #937, opened 2024-09-18
OTA-543: add autoscaler to cincinnati deployment

PratikMahajan:
Add autoscaler to cincinnati instance that adjusts the number of pods 
depending on traffic handled by policy engine. 
The traffic setting is variable and can be changed while applying the template

### DIFF
--- a/dist/openshift/cincinnati-deployment.yaml
+++ b/dist/openshift/cincinnati-deployment.yaml
@@ -11,7 +11,7 @@ objects:
         app: cincinnati
       name: cincinnati
     spec:
-      replicas: ${{REPLICAS}}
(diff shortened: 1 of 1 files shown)


petr-muller (approved):
tentative LGTM but we should test this well (meaning whether this autoscaling pattern/resource works well)

petr-muller:
/hold

Holding to avoid retester going crazy, unhold as needed (=overrides and necessary testing done)

PratikMahajan:
/override ci/prow/cargo-test

overriding known test failures

PratikMahajan:
/hold cancel

We will be testing the change in the integration environment before bumping the stage and production instance

PR #942, opened 2024-10-02
OTA-543: Add PrometheusRule to create a recording rule for calculating avg requests

PratikMahajan:
Added a PrometheusRule to create a recording rule. 
This is needed as HPA cant directly use complex PromQL queries,
and the recording rule pre-calculates the value for us.

### DIFF
--- a/dist/openshift/cincinnati-deployment.yaml
+++ b/dist/openshift/cincinnati-deployment.yaml
@@ -187,17 +187,26 @@ objects:
         name: cincinnati
       minReplicas: ${{MIN_REPLICAS}}
       maxReplicas: ${{MAX_REPLICAS}}
+      # Using 'Pods' type, as the metric is summed by pod, so each pod has its own value and
+      # HPA will automatically average across all pods. We want to scale based on per-pod metrics
       metrics:
         - type: Pods
           pods:
             metric:
-              name: cincinnati_pe_graph_incoming_requests_per_second
-              selector:
-                matchLabels:
-                  uri_path: "/api/upgrades_info/v1/graph"
+              name: cincinnati_policy_engine_graph_incoming_requests_rate
             target:
               type: AverageValue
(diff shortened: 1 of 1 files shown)


dist/openshift/cincinnati-deployment.yaml
```diff
@@ -187,17 +187,26 @@ objects:
...
+              averageValue: ${{PE_REQ_AVG}}
+  - apiVersion: monitoring.coreos.com/v1
+    kind: PrometheusRule
+    metadata:
+      name: cincinnati-recording-rule
+    spec:
+      groups:
+        - name: cincinnati.rules
+          rules:
+            - record: cincinnati_request_rate
```
wking: nit: `cincinnati_policy_engine_incoming_request_rate` or something more qualified, in case we need other metrics in this space later?

dist/openshift/cincinnati-deployment.yaml
```diff
@@ -187,17 +187,26 @@ objects:
...
+              expr: sum by (pod) (rate(cincinnati_pe_graph_incoming_requests_total[1m]))
```
wking: [Default `scrapeInterval` is 30s][1], so `1m` only gives time for 2 (and very rarely 3?) scrapes.  We'd get a more reliable rate estimate that could handle occasional scrape jitter or failures if we used `2m` or similar.

[1]: https://docs.openshift.com/container-platform/4.16/rest_api/monitoring_apis/prometheus-monitoring-coreos-com-v1.html#spec

dist/openshift/cincinnati-deployment.yaml
wking: 78a4a71eaf3b8 uses `cincinnati_policy_engine_graph_incoming_requests_rate`, marking this thread resolved.

wking (approved):
/lgtm

PratikMahajan:

/override ci/prow/cargo-test

PR #946, opened 2024-10-03
OTA-1348: Increment `http_upstream_reqs` on errors

petr-muller:
**increment `http_upstream_reqs` on errors**

Previously the `http_upstream_reqs` was only incremented when `cached_graph`
call was successful (and result was not cached). When `cached_graph` actually
made the call internally but it was not successful, the `.await?` shortcut
returned the error right away and the counter was never incremented, even
when the call was actually made.

This is actually tested in the `fetch_upstream_failure_test` macro:

```rust
assert_eq!(1, plugin.http_upstream_reqs.get() as usize);
```

These tests are failing for some time and it is hard to imagine how they
were ever passing.

---

**cincinnati_graph_fetch: simplify tests**

The `clone()` calls are `Arc::clone()` tests under the hood, so they
result in returning smart pointers. This is legit (we still `get()` on a
correct thing when testing) but it does not seem to be necessary and it
is confusing (how are we expecting a clone to be incremented? aha, they
are not actually clones under the hood!) so we can remove them

---

**cincinnati-graph-fetch: add comments about reliability**

During pair programming with David we discovered these tests are flaking:

There is a check that blocks on the asychnonous IO operation and asserts
that its result was an error (failure tests) or not (success test):

```rust
assert!(runtime.block_on(future_result).is_err());
```

We have seen these flaking as if individual tests interacted with each
other, such as if the plugin actually communicated with a graph-builder
mock created in a different test. This can be reproduced easily by running
all tests locally (some will randomly fail), then comment out the success
ones, suddenly all failure tests pass. This still indicates the tests
are not working correctly (they still communicate with a differently
failing mock server).

```
assertion failed: runtime.block_on(future_result).is_err()
thread 'plugins::internal::cincinnati_graph_fetch::tests::fetch_fail_request_fails_with_404' panicked at cincinnati/src/plugins/internal/cincinnati_graph_fetch.rs:338:5:
assertion failed: runtime.block_on(future_result).is_err()
```

### DIFF
--- a/cincinnati/src/plugins/internal/cincinnati_graph_fetch.rs
+++ b/cincinnati/src/plugins/internal/cincinnati_graph_fetch.rs
@@ -161,18 +161,26 @@ impl CincinnatiGraphFetchPlugin {
         }
 
         trace!("getting graph from upstream at {}", self.upstream);
-        let call_result = cached_graph(&self.client, &self.upstream, headers).await?;
-        // Increase request counter only if actual call was made
-        if !call_result.was_cached {
-            self.http_upstream_reqs.inc();
+        let result = cached_graph(&self.client, &self.upstream, headers).await;
+        match result {
+            Ok(call) => {
+                // Increase request counter only if actual call was made
+                if !call.was_cached {
+                    self.http_upstream_reqs.inc();
+                }
+                get_active_span(|span| {
+                    span.set_attribute(Key::new("cached").bool(call.was_cached));
+                });
+                Ok(InternalIO {
+                    graph: call.value,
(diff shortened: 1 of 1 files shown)


DavidHurta (approved):

petr-muller:
It is quite interesting that CI does not seem to suffer from the gotcha that makes the `cincinnati-graph-fetch` tests fail (the thing discussed by the comment added in the third commit). But the PR at least improves the state and the comment documents the gotcha for the reader so :shipit:

petr-muller:
/override ci/prow/cargo-test

4 fixed, five to go

PR #948, opened 2024-10-08
NO-JIRA: deps: remove unused `failure` dependency

petr-muller:
It seems that this crate was used in the past to provide `Error` traits implementation, but was eventually replaced with `thiserror::Error` (see the removed `use thiserror::Error as Fail` rename). The `Fail` trait was only derived at several places, so I dropped the rename and used `thiserror::Error` directly in derives. The `Fail` from the commons package was also used in the cincinnati package, relying on a transitive dependency when the code is actually directly used. I dropped this indirection and made cincinnati package depend on thiserror (same version like commons)

Also opportunistically bump `thiserror` from 1.0.49 to 1.0.55

### DIFF
--- a/cincinnati/Cargo.toml
+++ b/cincinnati/Cargo.toml
@@ -44,6 +44,7 @@ zeroize = "=1.3.0"
 hamcrest2 = "0.3.0"
 cached = "^0.44.0"
 similar = { version = "2.6.0", optional = true }
+thiserror = "1.0"
 
 [dev-dependencies]
 mockito = "0.31.1"
--- a/commons/src/errors.rs
+++ b/commons/src/errors.rs
@@ -11,9 +11,6 @@ pub mod prelude {
     pub use anyhow::Context;
     pub use anyhow::Error;
     pub use anyhow::Result as Fallible;
-
-    // Macro imports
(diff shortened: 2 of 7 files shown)


commons/src/errors.rs
```diff
@@ -11,9 +11,6 @@ pub mod prelude {
     pub use anyhow::Context;
     pub use anyhow::Error;
     pub use anyhow::Result as Fallible;
-
-    // Macro imports
-    pub use thiserror::Error as Fail;
```
petr-muller, line 16: This is `thiserror` pretenting to be `failure`

cincinnati/Cargo.toml
```diff
@@ -44,6 +44,7 @@ zeroize = "=1.3.0"
 hamcrest2 = "0.3.0"
 cached = "^0.44.0"
 similar = { version = "2.6.0", optional = true }
+thiserror = "1.0"
```
petr-muller, line 47: We use this in `commons` too, so it is not a new dep. Previously we used under its renamed `Fail` name through `commons`, the rename is no longer there so we need to use the dependency directly.

petr-muller:
/override ci/prow/cargo-test

Known five failures

PR #950, opened 2024-10-08
OTA-1348: Fix the failing `compare_quay_result_fixture::_20211124_083150_` test

DavidHurta:
**Fix metadata parser test fixture**
The commit [1] has removed conditional edges from normal edges. This was
done using the `io.openshift.upgrades.graph.previous.remove_regex`
metadata key. The `from` regex field of a conditional risk is appended
to the key's value of a respective `to` version. In other words, the
`OpenshiftSecondaryMetadataParser` parser signals using the metadata
key to the `EdgeAddRemove` plugin what edges to remove from the normal
edges thanks to its knowledge of conditional risks.

Unfortunately, a test fixture was not modified to adhere to the new
changes. The `graph-gb-with-quay-metadata.json` file is supposed to
represent graph with the metadata to act as control data.
This PR fixes that.

[1] https://github.com/openshift/cincinnati/commit/72ed35630cfcb0796473deffdc6c6c2c31ed1155

---

**cincinnati/src/lib.rs: Declare openly Graph's == operator limitations**
Even though conditional edges are part of the Graph type and are
equally as important as the graph itself, they are not taken into
account when two Graps are compared using the `==` operator. Declare
this fact more openly.

### DIFF
--- a/cincinnati/src/lib.rs
+++ b/cincinnati/src/lib.rs
@@ -691,6 +691,9 @@ impl Serialize for Graph {
 
 #[cfg(any(test, feature = "test"))]
 impl PartialEq for Graph {
+    /// Compares two Graphs.
+    ///
+    /// Does not take into account the `conditional_edges` field of Graphs.
     fn eq(&self, other: &Graph) -> bool {
         let mut releases = self
             .dag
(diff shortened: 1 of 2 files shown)


DavidHurta:
Let's be cautious. Modifying the control data to pass the test could be bad in the case I have misunderstood Cincinnati :eyes:

petr-muller (approved):
Looking at https://github.com/openshift/cincinnati/commit/72ed35630cfcb0796473deffdc6c6c2c31ed1155 the proposed fixture change makes sense to me as a right fix but it would probably be best if @PratikMahajan confirmed.

PratikMahajan (approved):
/lgtm

petr-muller:
/hold cancel
/override ci/prow/cargo-test

Pratik approved this so let's move this forward.

PR #953, opened 2024-10-09
replace HPA with keda ScaledObject

PratikMahajan:
HPA wont give us custom prometheus metrics in OCP instance,
thus we need to add keda ScaledObject which allows us to run
custom prometheus queries and in the background runs HPA.
OCP instance needs Keda for this to work.


PratikMahajan:

/override ci/prow/cargo-test

PR #956, opened 2024-10-11
ScaledObject threshold value should be a string 'integer'

PratikMahajan:
the threshold value is an integer without quotes, which is invalid. this fixes it


PratikMahajan:

/override ci/prow/cargo-test

PR #959, opened 2024-10-14
OTA-543: fix the PE_REQ_AVG templating

PratikMahajan:
use single {} for string value, {{}} is used for  JSON or YAML value

### DIFF
--- a/dist/openshift/cincinnati-deployment.yaml
+++ b/dist/openshift/cincinnati-deployment.yaml
@@ -189,7 +189,7 @@ objects:
           metadata:
             serverAddress: http://prometheus-app-sre.openshift-customer-monitoring.svc.cluster.local:9090
             metricName: cincinnati_policy_engine_graph_incoming_requests_rate
-            threshold: "${{PE_REQ_AVG}}"
+            threshold: "${PE_REQ_AVG}"
             query: avg(cincinnati_policy_engine_graph_incoming_requests_rate)
   - apiVersion: monitoring.coreos.com/v1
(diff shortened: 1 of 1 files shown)


dist/openshift/cincinnati-deployment.yaml
```diff
@@ -189,7 +189,7 @@ objects:
           metadata:
             serverAddress: http://prometheus-app-sre.openshift-customer-monitoring.svc.cluster.local:9090
             metricName: cincinnati_policy_engine_graph_incoming_requests_rate
-            threshold: "${{PE_REQ_AVG}}"
+            threshold: "${PE_REQ_AVG}"
```
petr-muller, line 192: Would the following be preferable for consistency with how we use e.g. MAX_REPLICAS/MIN_REPLICAS above on L185-186?

```suggestion
            threshold: ${{PE_REQ_AVG}}
```

PratikMahajan: we moved to `"${{PE_REQ_AVG}}"` in https://github.com/openshift/cincinnati/pull/956 because `${{PE_REQ_AVG}}` wont work. 
The actual solution according to app-sre is `"${PE_REQ_AVG}"`

PratikMahajan:

/override ci/prow/cargo-test

dist/openshift/cincinnati-deployment.yaml
petr-muller: :+1:

PR #967, opened 2024-10-22
bring back the replica field and set it to min-replicas

PratikMahajan:
bring back the replicas field and set it to minimum replicas. 
so that we dont end up with 1 replica while the scaledobject adjusts it.


PratikMahajan:
overriding all the tests. as we dont need the tests for it

PR #969, opened 2024-10-23
OTA-1378,OTA-1379: add retry logic for pulling images and less logs for sigs

PratikMahajan:
we're adding a retry logic that tries to pull layers if it fails due to any reason. We retry for 3 times before ultimately failing.

Also moved the warn log for sig pulling to debug and added a counter to tell how many sig images we've ignored

### DIFF
--- a/cincinnati/src/plugins/internal/graph_builder/release_scrape_dockerv2/registry/mod.rs
+++ b/cincinnati/src/plugins/internal/graph_builder/release_scrape_dockerv2/registry/mod.rs
@@ -32,6 +32,7 @@ use std::io::Read;
 use std::iter::Iterator;
 use std::path::{Path, PathBuf};
 use std::string::String;
+use std::sync::atomic::{AtomicUsize, Ordering};
 use std::sync::Arc;
 use tar::Archive;
 use url::Url;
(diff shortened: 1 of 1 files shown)


cincinnati/src/plugins/internal/graph_builder/release_scrape_dockerv2/registry/mod.rs
```diff
@@ -315,31 +318,31 @@ pub async fn fetch_releases(
...
                     Ok(result) => result,
                     Err(e) => {
-                        warn!(
-                            "error fetching manifest and manifestref for {}:{}: {}, ignoring this image",
-                            &repo,
-                            &tag,
-                            e
+                        if tag.contains(".sig") {
+                            debug!(
+                                "encountered a signature for {}:{}: {}, ignoring this image",
```
wking, line 339: we might encounter other releases we don't recognize which don't include the `.sig` tag pattern.  I'm also not all that worried about pulling out `.sig` as a specific counting bucket.  Can just increment the counter whenever we hit an error, and maybe hold on to the most recent error, and then log:

```rust
warn!("ignored {} tags including: {}", ignored_count, most_recent_error)
```

or something?

PratikMahajan: added another counter for invalid releases as we want to separate from the releases that we're deliberately ignoring.

cincinnati/src/plugins/internal/graph_builder/release_scrape_dockerv2/registry/mod.rs
```diff
@@ -315,31 +316,36 @@ pub async fn fetch_releases(
...
-                    // stopgap till we teach cincinnati to deal with signatures
+                match get_manifest_layers(tag.to_owned(), &repo, &registry_client).await {
                     Ok(result) => result,
                     Err(e) => {
-                        warn!(
-                            "error fetching manifest and manifestref for {}:{}: {}, ignoring this image",
-                            &repo,
-                            &tag,
-                            e
+                        if tag.contains(".sig") {
```
wking, line 337: [Tag-based discovery][1] is [one option][2] for finding Sigstore signatures.  In the future, we might move to [listing referrers][3] (containers/image#2030).  But if we do, the failure modes for this line:

* Misclassifying a non-sig as a Sigstore signature because it happens to use this tag structure, or
* Misclassifying a Sigstore signature as a non-sig, non-release ignored image,

both seem low, so :shrug:, I'm ok with this heuristic.

[1]: https://github.com/sigstore/cosign/blob/v2.4.0/specs/SIGNATURE_SPEC.md#tag-based-discovery
[2]: https://github.com/sigstore/cosign/blob/v2.4.0/specs/SIGNATURE_SPEC.md#discovery
[3]: https://github.com/opencontainers/distribution-spec/blob/v1.1.0/spec.md#listing-referrers

PratikMahajan: for misclassifying non-sig to sig will always be a risk if we do string comparison. but imo should be rare. 
if signature gets classifies as non-sig, the logs should bring that into notice. Not a lot worried about the mismatch.
We can also change this logic when we get listing referrers in dkregistry and pull it downstream to cincinnati.

cincinnati/src/plugins/internal/graph_builder/release_scrape_dockerv2/registry/mod.rs
```diff
@@ -527,17 +529,49 @@ async fn get_manifest_and_ref(
...
+                        e
+                    )
+                })
+                .await
+            {
+                Ok(manifest_and_ref) => break manifest_and_ref,
+                Err(e) => {
+                    // signatures are not identified by dkregistry and not useful for cincinnati graph, dont retry and return error
+                    if tag.contains(".sig") {
+                        return Err(e);
```
wking, line 557: and then this error bubbles up and is converted to a debug message via the `.sig` branch of `fetch_releases`'s `get_manifest_layers` handling in 022a8d651db44cffaf0aae3b908ef6d41cc721b6.

wking (approved):
/lgtm

PratikMahajan:
/override ci/prow/cargo-test

override known test failures

PR #970, opened 2024-10-23
OTA-1348: Fix registry URL parser tests

DavidHurta:
Fixes a currently failing CI test regarding parsing registry URLs.

### DIFF
--- a/cincinnati/src/plugins/internal/graph_builder/release_scrape_dockerv2/registry/mod.rs
+++ b/cincinnati/src/plugins/internal/graph_builder/release_scrape_dockerv2/registry/mod.rs
@@ -667,96 +667,164 @@ fn parse_url(
 mod tests {
     use super::*;
 
+    struct TestCase {
+        input: String,
+        expected_registry: Registry,
+        expected_registry_host_port_string: String,
+        expected_registry_host_port_namespaced_string: String,
+    }
+
+    fn setup_test_cases() -> Vec<TestCase> {
+        vec![
+            TestCase {
+                input: "localhost".to_string(),
+                expected_registry: Registry {
+                    scheme: "".to_string(),
+                    insecure: false,
+                    host: "localhost".to_string(),
+                    port: None,
+                    namespace: "/".to_string(),
+                },
+                expected_registry_host_port_string: "localhost".to_string(),
+                expected_registry_host_port_namespaced_string: "localhost/".to_string(),
+            },
+            TestCase {
+                input: "http://localhost:8080".to_string(),
+                expected_registry: Registry {
+                    scheme: "http".to_string(),
+                    insecure: true,
+                    host: "localhost".to_string(),
+                    port: Some(8080),
+                    namespace: "/".to_string(),
+                },
+                expected_registry_host_port_string: "http://localhost:8080".to_string(),
+                expected_registry_host_port_namespaced_string: "http://localhost:8080/".to_string(),
+            },
+            TestCase {
+                input: "http://localhost:8080/ns1/ns2".to_string(),
+                expected_registry: Registry {
+                    scheme: "http".to_string(),
+                    insecure: true,
+                    host: "localhost".to_string(),
+                    port: Some(8080),
+                    namespace: "/ns1/ns2".to_string(),
+                },
+                expected_registry_host_port_string: "http://localhost:8080".to_string(),
+                expected_registry_host_port_namespaced_string: "http://localhost:8080/ns1/ns2"
+                    .to_string(),
+            },
+            TestCase {
+                input: "127.0.0.1".to_string(),
+                expected_registry: Registry {
+                    scheme: "".to_string(),
+                    insecure: false,
+                    host: "127.0.0.1".to_string(),
+                    port: None,
+                    namespace: "/".to_string(),
+                },
+                expected_registry_host_port_string: "127.0.0.1".to_string(),
+                expected_registry_host_port_namespaced_string: "127.0.0.1/".to_string(),
+            },
+            TestCase {
+                input: "127.0.0.1/ns1/ns2".to_string(),
+                expected_registry: Registry {
+                    scheme: "".to_string(),
+                    insecure: false,
+                    host: "127.0.0.1".to_string(),
+                    port: None,
+                    namespace: "/ns1/ns2".to_string(),
+                },
+                expected_registry_host_port_string: "127.0.0.1".to_string(),
+                expected_registry_host_port_namespaced_string: "127.0.0.1/ns1/ns2".to_string(),
+            },
+            TestCase {
+                input: "sat-r220-02.lab.eng.rdu2.redhat.com:5000".to_string(),
+                expected_registry: Registry {
+                    scheme: "".to_string(),
+                    insecure: false,
+                    host: "sat-r220-02.lab.eng.rdu2.redhat.com".to_string(),
+                    port: Some(5000),
+                    namespace: "/".to_string(),
+                },
+                expected_registry_host_port_string: "sat-r220-02.lab.eng.rdu2.redhat.com:5000"
+                    .to_string(),
+                expected_registry_host_port_namespaced_string:
+                    "sat-r220-02.lab.eng.rdu2.redhat.com:5000/".to_string(),
+            },
+            TestCase {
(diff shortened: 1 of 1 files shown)


cincinnati/src/plugins/internal/graph_builder/release_scrape_dockerv2/registry/mod.rs
```diff
@@ -666,96 +667,365 @@ fn parse_url(
...
+    #[test]
+    fn registry_host_port_string() {
+        let tests = setup_test_cases();
+        for test in tests {
+            let registry: Registry = Registry::try_from_str(&test.input)
+                .unwrap_or_else(|_| panic!("could not parse {} to registry", &test.input));
+            assert_eq!(
+                test.expected_registry_host_port_string,
+                registry.host_port_string()
+            );
```
DavidHurta: Not sure if we should rather test this against the expected_registry.

```suggestion
            assert_eq!(
                test.expected_registry_host_port_string,
                test.expected_registry.host_port_string()
            );
```

DavidHurta:
/hold
QE should probably verify the changes. Let me analyze the code to decide whether should I test the changes myself first in a live cluster (or let me know).

cincinnati/src/plugins/internal/graph_builder/release_scrape_dockerv2/registry/mod.rs
```diff
@@ -667,96 +667,164 @@ fn parse_url(
...
-                "http://localhost:8080",
-                Registry {
+    struct TestCase {
+        input: String,
+        expected_registry: Registry,
+        expected_registry_host_port_string: String,
+        expected_registry_host_port_namespaced_string: String,
+    }
+
+    fn setup_test_cases() -> Vec<TestCase> {
```
petr-muller: I think this should have `#[cfg(test)]` so it only compiles when testing

petr-muller: also feels like a `make_test_cases` or `get_test_cases` rather than `setup_` which I'd expect to setup _something_ for the tests, not just make the list of testcases

cincinnati/src/plugins/internal/graph_builder/release_scrape_dockerv2/registry/mod.rs
petr-muller: Using `registry` as the tested value feels correct to me

cincinnati/src/plugins/internal/graph_builder/release_scrape_dockerv2/registry/mod.rs
```diff
@@ -667,96 +667,164 @@ fn parse_url(
...
+    #[test]
+    fn registry_host_port_namespaced_string() {
+        let tests = setup_test_cases();
+        for test in tests {
+            let registry: Registry = Registry::try_from_str(&test.input)
+                .unwrap_or_else(|_| panic!("could not parse {} to registry", &test.input));
+            assert_eq!(
+                test.expected_registry_host_port_namespaced_string,
+                registry.host_port_namespaced_string()
+            );
```
petr-muller: TBH I'm not sure whether this is worth a separate test, I'd just add that assertion to the previous one 

/shrug

cincinnati/src/plugins/internal/graph_builder/release_scrape_dockerv2/registry/mod.rs
```diff
@@ -667,96 +667,164 @@ fn parse_url(
...
+    struct TestCase {
+        input: String,
+        expected_registry: Registry,
+        expected_registry_host_port_string: String,
+        expected_registry_host_port_namespaced_string: String,
+    }
+
+    fn setup_test_cases() -> Vec<TestCase> {
+        vec![
+            TestCase {
```
petr-muller: personally I'm missing some kind of hint about the intent of each testcases - what is usually called "name" in Go tests ;)

I can figure it out if I look at all testcases but if every case had a comment like "hostname with port and path" it'd be easier to understand on a first glance

petr-muller (commented):
Mostly nits. I agree that doing a round of testing to make sure something does not accidentally depend on that path parsing difference could be useful if QE have a capacity for it. But if we cannot test it quickly then I'd just merge, would not block on testing.

cincinnati/src/plugins/internal/graph_builder/release_scrape_dockerv2/registry/mod.rs
DavidHurta: > also feels like a make_test_cases or get_test_cases rather than setup_ which I'd expect to setup something for the tests, not just make the list of testcases

True. No objections.

DavidHurta: > I think this should have #[cfg(test)] so it only compiles when testing

The module of the function already has the annotaion.

From the Rust documentation [1]:

> On the automatically generated tests module, the attribute cfg stands for configuration and tells Rust that the following item should only be included given a certain configuration option. In this case, the configuration option is test, which is provided by Rust for compiling and running tests. By using the cfg attribute, Cargo compiles our test code only if we actively run the tests with cargo test. This includes any helper functions that might be within this module, in addition to the functions annotated with #[test].

[1] https://doc.rust-lang.org/book/ch11-03-test-organization.html#the-tests-module-and-cfgtest

cincinnati/src/plugins/internal/graph_builder/release_scrape_dockerv2/registry/mod.rs
DavidHurta: As the `TestCase` struct has the field `input`, I am going to continue using the `registry` as the tested value. It does not matter *that* much.

cincinnati/src/plugins/internal/graph_builder/release_scrape_dockerv2/registry/mod.rs
DavidHurta: It is a different method, and the initialization of the test is only a few lines.

I do not see the downside as that bad. A new test for a new method.

/shrug

cincinnati/src/plugins/internal/graph_builder/release_scrape_dockerv2/registry/mod.rs
DavidHurta: Thanks for the review and the feedback. Let me improve this.

DavidHurta:
/unhold

The [host_port_namespaced_string](https://github.com/DavidHurta/cincinnati/blob/80e437248e6e32c1ac23ae6071ca2e2e54017b46/cincinnati/src/plugins/internal/graph_builder/release_scrape_dockerv2/registry/mod.rs#L139) function is only used in:

- [when passing its value to the registry:read_credentials in DkrV2OpenshiftSecondaryMetadataScraperPlugin](https://github.com/DavidHurta/cincinnati/blob/80e437248e6e32c1ac23ae6071ca2e2e54017b46/cincinnati/src/plugins/internal/graph_builder/dkrv2_openshift_secondary_metadata_scraper/plugin.rs#L200)
-  [when passing its value to the registry:read_credentials (and as an error for its unwrap_or_else) in ReleaseScrapeDockerv2Plugin](https://github.com/DavidHurta/cincinnati/blob/80e437248e6e32c1ac23ae6071ca2e2e54017b46/cincinnati/src/plugins/internal/graph_builder/release_scrape_dockerv2/plugin.rs#L129)
- tests

Its value is thus used only in the [read_credentials](https://github.com/DavidHurta/cincinnati/blob/80e437248e6e32c1ac23ae6071ca2e2e54017b46/cincinnati/src/plugins/internal/graph_builder/release_scrape_dockerv2/registry/mod.rs#L166) function (excluding tests and an error message). By analyzing its logic, the new changes (the addition of the  `/` forward slash to a `Registry` namespace when a scheme and namespace are not specified for a registry URL) should not affect it.

DavidHurta:
I will rebase the branch (squash some of the commits together) when changes are approved.

petr-muller (commented):
/lgtm
/override ci/prow/cargo-test

Changes LGTM

PR #973, opened 2024-10-24
OTA-543: remove prometheus rule and add query directly to scaledObject

PratikMahajan:
we added prometheusrule as we were not able to use complex
queries in HPA.
But with scaledObject we can directly use complex queries
for scaling replicas and dont need prometheusrule anymore.


petr-muller:
/uncc

Uncc-ing myself from a bunch of stale PRs to clean up my dashboards. If this PR becomes relevant again feel free to CC me again.

PR #991, opened 2024-11-25
OTA-1375: make readiness and liveness probe for cincinnati configurable

PratikMahajan:
make the liveness and readiness probe configurable so that we can dynamically adjust according to startup delays.


petr-muller:
/uncc

I am uncc-ing myself from some stale PRs that do not move. [OTA-1375](https://issues.redhat.com/browse/OTA-1375) is closed. Please `/cc` me if this PR is still relevant and needs my review.

PR #996, opened 2024-12-05
OTA-869: separate the cincinnati graph-builder and policy-engine deployments

PratikMahajan:
Separate the Graph Builder and Policy Engine deployments so that we can scale up PE quickly instead of waiting for graph to refresh. 
PE will use graph from existing graph-builder to get the complete graph

### DIFF
--- a/dist/openshift/cincinnati-deployment.yaml
+++ b/dist/openshift/cincinnati-deployment.yaml
@@ -212,13 +212,6 @@ objects:
                 requests:
                   cpu: ${PE_CPU_REQUEST}
                   memory: ${PE_MEMORY_REQUEST}
-          volumes:
-            - name: secrets
-              secret:
-                secretName: cincinnati-credentials
-            - name: configs
-              configMap:
-                name: cincinnati-configs
       triggers:
         - type: ConfigChange
 
@@ -300,7 +293,7 @@ objects:
       gb.rust_backtrace: "${RUST_BACKTRACE}"
       pe.address: "0.0.0.0"
       pe.status.address: "0.0.0.0"
-      pe.upstream: "${PE_UPSTREAM_URL}"
+      pe.upstream: "${PE_UPSTREAM_URL}${GB_PATH_PREFIX}/graph"
       pe.log.verbosity: ${{PE_LOG_VERBOSITY}}
       pe.mandatory_client_parameters: "channel"
       pe.rust_backtrace: "${RUST_BACKTRACE}"
@@ -465,7 +458,7 @@ parameters:
       [[plugin_settings]]
       name = "edge-add-remove"
   - name: PE_UPSTREAM_URL
-    value: "http://cincinnati-graph-builder.cincinnati.svc.cluster.local:8080${GB_PATH_PREFIX}/graph"
+    value: "http://cincinnati-graph-builder:8080"
     displayName: "Policy engine upstream URL"
     description: "URL for the policy engine to connect to graph builder"
   - name: RUST_BACKTRACE
@@ -475,8 +468,6 @@ parameters:
     value: "/etc/configs/gb.toml"
   - name: ENVIRONMENT_SECRETS
     value: '{ "CINCINNATI_GITHUB_SCRAPER_OAUTH_TOKEN_PATH": "/etc/secrets/github_token.key" }'
-  - name: GB_MIN_REPLICAS
-    value: "1"
   - name: PE_MIN_REPLICAS
     value: "1"
   - name: GB_MAX_REPLICAS
--- a/dist/openshift/cincinnati-deployment.yaml
+++ b/dist/openshift/cincinnati-deployment.yaml
@@ -8,13 +8,13 @@ objects:
     kind: Deployment
     metadata:
       labels:
-        app: cincinnati
-      name: cincinnati
+        app: cincinnati-graph-builder
+      name: cincinnati-graph-builder
     spec:
-      replicas: ${{MAX_REPLICAS}}
+      replicas: ${{GB_MAX_REPLICAS}}
       selector:
         matchLabels:
-          app: cincinnati
+          app: cincinnati-graph-builder
       strategy:
         type: RollingUpdate
         rollingUpdate:
@@ -23,7 +23,7 @@ objects:
       template:
         metadata:
           labels:
-            app: cincinnati
+            app: cincinnati-graph-builder
         spec:
           affinity:
             podAntiAffinity:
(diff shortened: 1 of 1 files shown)


dist/openshift/cincinnati-deployment.yaml
```diff
@@ -410,16 +464,24 @@ parameters:
...
+    description: "URL for the policy engine to connect to graph builder"
   - name: RUST_BACKTRACE
     value: "0"
     displayName: Set RUST_BACKTRACE env var
   - name: GB_CONFIG_PATH
     value: "/etc/configs/gb.toml"
   - name: ENVIRONMENT_SECRETS
     value: '{ "CINCINNATI_GITHUB_SCRAPER_OAUTH_TOKEN_PATH": "/etc/secrets/github_token.key" }'
-  - name: MIN_REPLICAS
+  - name: GB_MIN_REPLICAS
```
DavidHurta: `GB_MIN_REPLICAS` is not used anywhere. Do we plan to autoscale the GB? Otherwise, we could omit the parameter for simplicity.

dist/openshift/cincinnati-deployment.yaml
```diff
@@ -410,16 +464,24 @@ parameters:
 
       [[plugin_settings]]
       name = "edge-add-remove"
+  - name: PE_UPSTREAM_URL
+    value: "http://cincinnati-graph-builder.cincinnati.svc.cluster.local:8080${GB_PATH_PREFIX}/graph"
```
DavidHurta: I am not sure whether we can assume that the application will be deployed in the `cincinnati` namespace. For example, we use the `cincinnati-production` namespace in production. This template is used in https://gitlab.cee.redhat.com/service/app-interface/-/blob/master/data/services/cincinnati/cicd/ci-int/saas.yaml, which ends up deploying Cincinnati in various namespaces.

DavidHurta: We might use just the service name given the pods are deployed in the same namespace. For example:

```suggestion
    value: "http://cincinnati-graph-builder:8080${GB_PATH_PREFIX}/graph"
```

dist/openshift/cincinnati-deployment.yaml
```diff
@@ -298,6 +332,26 @@ objects:
...
+        name: cincinnati-policy-engine
+      maxReplicaCount: ${{PE_MAX_REPLICAS}}
+      minReplicaCount: ${{PE_MIN_REPLICAS}}
+      triggers:
+        - type: prometheus
+          metadata:
+            serverAddress: http://prometheus-app-sre.openshift-customer-monitoring.svc.cluster.local:9090
+            metricName: cincinnati_policy_engine_graph_incoming_requests_rate
+            threshold: "${PE_REQ_AVG}"
+            query: sum by (pod) (rate(cincinnati_pe_graph_incoming_requests_total[2m]))
```
DavidHurta, line 346: Is there a specific reason why do we change the query? I was not expecting such a change in a commit to separate the GB and PE.

DavidHurta: I have noticed that the PrometheusRule cincinnati-recording-rule is now removed and thus the query is updated. Could we move this change to a separate commit and explain why the change is introduced?

dist/openshift/cincinnati-deployment.yaml
```diff
@@ -298,6 +332,26 @@ objects:
...
+    spec:
+      scaleTargetRef:
+        name: cincinnati-policy-engine
+      maxReplicaCount: ${{PE_MAX_REPLICAS}}
+      minReplicaCount: ${{PE_MIN_REPLICAS}}
+      triggers:
+        - type: prometheus
+          metadata:
+            serverAddress: http://prometheus-app-sre.openshift-customer-monitoring.svc.cluster.local:9090
+            metricName: cincinnati_policy_engine_graph_incoming_requests_rate
```
DavidHurta, line 344: Reference of a removed metric as the PrometheusRule cincinnati-recording-rule is no longer present. I am not that knowledgeable regarding the APIs, feel free to correct me.

dist/openshift/cincinnati-deployment.yaml
DavidHurta: It seems like a cross-reference of parameters is not working. Note the literal `"GB_PATH_PREFIX"` string.

```sh
$ oc process --local -f dist/openshift/cincinnati-deployment.yaml | jq '.items[] | select(.metadata.name=="cincinnati")'
{
  "apiVersion": "v1",
  "data": {
    "gb.rust_backtrace": "0",
    "pe.address": "0.0.0.0",
    "pe.log.verbosity": "vv",
    "pe.mandatory_client_parameters": "channel",
    "pe.rust_backtrace": "0",
    "pe.status.address": "0.0.0.0",
    "pe.upstream": "http://cincinnati-graph-builder.cincinnati.svc.cluster.local:8080${GB_PATH_PREFIX}/graph"
  },
  "kind": "ConfigMap",
  "metadata": {
    "name": "cincinnati"
  }
}
```

Have you tested this template on a live cluster?

dist/openshift/cincinnati-deployment.yaml
```diff
@@ -171,34 +221,7 @@ objects:
                 name: cincinnati-configs
       triggers:
```
DavidHurta, line 215: This review comment reference lines 215-221:

```yaml
          volumes:
            - name: secrets
              secret:
                secretName: cincinnati-credentials
            - name: configs
              configMap:
                name: cincinnati-configs
```

The volumes seem to not be used in the `cincinnati-policy-engine` deployment and thus could be omitted.

DavidHurta: Or they are supposed to be mounted, and they are not :eyes:

DavidHurta (changes requested):

dist/openshift/cincinnati-deployment.yaml
PratikMahajan: fixed

dist/openshift/cincinnati-deployment.yaml
PratikMahajan: this is a combination of prometheus rule and previoud keda file.
Saw that prometheus rule is redundant. so merged everything into keda config

DavidHurta: Got it, thanks.

dist/openshift/cincinnati-deployment.yaml
DavidHurta: Shouldn't `cincinnati_pe_graph_incoming_requests_total` be used here as it is used in the query?

```suggestion
            metricName: cincinnati_pe_graph_incoming_requests_total
```

petr-muller:
/uncc

David has it covered :)

PR #997, opened 2024-12-09
OTA-1348: dist/cargo_test.sh: Do not fail when testing `rh-manifest-generator` using nextest

DavidHurta:
The `rh-manifest-generator` package does not have any tests. By default, nextest fails if there are no tests to run.

As it is expected for the `rh-manifest-generator` package to have no tests at the moment, pass when running `cargo nextest run` on the package. Fail on other packages, as it is expected for them to contain tests.

`cargo test` passes on empty tests. To be more consistent across the test runners, we should pass if there are no tests to run. However, as we utilize the nextest in CI, I would like to be more strict when using the runner if easily possible.

### DIFF
--- a/dist/cargo_test.sh
+++ b/dist/cargo_test.sh
@@ -12,6 +12,16 @@ cargo_test_flags["rh-manifest-generator"]=""
 cargo_test_flags["prometheus-query"]=""
 cargo_test_flags["quay"]="--features test-net"
 
+declare -A cargo_nextest_flags
+cargo_nextest_flags["cincinnati"]=""
+cargo_nextest_flags["commons"]=""
+cargo_nextest_flags["graph-builder"]=""
+cargo_nextest_flags["policy-engine"]=""
+cargo_nextest_flags["metadata-helper"]=""
+cargo_nextest_flags["rh-manifest-generator"]="--no-tests=pass"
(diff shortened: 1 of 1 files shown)


dist/cargo_test.sh
```diff
@@ -12,6 +12,16 @@ cargo_test_flags["rh-manifest-generator"]=""
...
 
+declare -A cargo_nextest_flags
+cargo_nextest_flags["cincinnati"]=""
+cargo_nextest_flags["commons"]=""
+cargo_nextest_flags["graph-builder"]=""
+cargo_nextest_flags["policy-engine"]=""
+cargo_nextest_flags["metadata-helper"]=""
+cargo_nextest_flags["rh-manifest-generator"]="--no-tests=pass"
+cargo_nextest_flags["prometheus-query"]=""
+cargo_nextest_flags["quay"]=""
```
DavidHurta, line 23:
```suggestion
declare -A cargo_nextest_flags
cargo_nextest_flags["rh-manifest-generator"]="--no-tests=pass"
```

works as well; however, I could not find in documentation the specified behavior when a key does not exist in an associative array.

petr-muller:
/test e2e
/override ci/prow/cargo-test

Failures in `cargo test` adressed in #978

petr-muller (approved):

PR #1023, opened 2025-05-13
release_scrape_dockerv2: Log cache misses

wking:
To reduce the load on the registry, we want to avoid pulling images while walking the tag list.  With this commit, we'll log the number of cache misses, so we can see when we need to bump verbosity to find gaps in our caching logic.


petr-muller:
Lol that Copilot review is useless :D

petr-muller (approved):

PR #1025, opened 2025-05-15
[WIP] OCPBUGS-56269: Remove registry cache placeholder on scrape failure

PratikMahajan:
replace the placeholders with `None`
so no more 0.0.0

and

remove the placeholder from the cache of we fail to get any release
metadata

### DIFF
--- a/cincinnati/src/plugins/internal/graph_builder/release_scrape_dockerv2/registry/mod.rs
+++ b/cincinnati/src/plugins/internal/graph_builder/release_scrape_dockerv2/registry/mod.rs
@@ -478,42 +478,43 @@ async fn lookup_or_fetch(
         }
         None => {
             cache_misses.fetch_add(1, Ordering::SeqCst);
-            let placeholder = Option::from(Metadata {
-                kind: MetadataKind::V0,
-                version: Version::new(0, 0, 0),
-                previous: vec![],
-                next: vec![],
-                metadata: Default::default(),
-            });
-            cache.write().await.insert(manifestref.clone(), placeholder);
+            cache.write().await.insert(manifestref.clone(), None);
 
-            let metadata = find_first_release_metadata(
+            let metadata = match find_first_release_metadata(
                 layer_digests,
                 registry_client,
                 repo.clone(),
                 tag.clone(),
             )
             .await
-            .context("failed to find first release")?
-            .map(|mut metadata| {
-                // Attach the manifestref this release was found in for further processing
-                metadata
-                    .metadata
-                    .insert(manifestref_key, manifestref.clone());
-
-                // Process the manifest architecture if given
-                if let Some(arch) = arch {
-                    // Encode the architecture as SemVer information
-                    metadata.version.build = vec![semver::Identifier::AlphaNumeric(arch.clone())];
-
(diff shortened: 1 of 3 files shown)


cincinnati/src/plugins/internal/graph_builder/release_scrape_dockerv2/registry/mod.rs
```diff
@@ -651,16 +661,21 @@ async fn find_first_release_metadata(
...
+                "[{}] Could not assemble metadata from layer ({})",
+                &tag, &layer_digest
+            ))? {
             Ok(metadata) => {
                 return Ok(Some(metadata));
             }
             Err(e) => {
-                debug!(
+                return Err(format_err!(
                     "[{}] Could not assemble metadata from layer ({}): {}",
```
wking, line 674: Hey look, there's some CI coverage in [cargo-test][1]:

```
cincinnati: plugins::internal::graph_builder::release_scrape_dockerv2::plugin::network_tests::scrape_public_with_first_empty_tag_must_succeed expand_less	1s
{Error: failed to fetch all release metadata from registry.ci.openshift.org/cincinnati-ci-public/cincinnati-test-emptyfirsttag-public-manual test failure with exit code 101 Error: failed to fetch all release metadata from registry.ci.openshift.org/cincinnati-ci-public/cincinnati-test-emptyfirsttag-public-manual

Caused by:
    [0.0.0] Could not assemble metadata from layer (sha256:4ca545ee6d5db5c1170386eeb39b2ffe3bd46e5d4a73a9acbebc805f19607eb3): 'release-manifests/release-metadata' not found}

cincinnati: plugins::internal::graph_builder::release_scrape_dockerv2::plugin::network_tests::scrape_public_multiarch_manual_succeeds expand_less	1s
{Error: failed to fetch all release metadata from registry.ci.openshift.org/cincinnati-ci-public/cincinnati-test-public-multiarch-manual test failure with exit code 101 Error: failed to fetch all release metadata from registry.ci.openshift.org/cincinnati-ci-public/cincinnati-test-public-multiarch-manual

Caused by:
    [0.0.3-amd64] Could not assemble metadata from layer (sha256:4ca545ee6d5db5c1170386eeb39b2ffe3bd46e5d4a73a9acbebc805f19607eb3): 'release-manifests/release-metadata' not found}
```

I'm not clear yet on whether that's turning up fixtures/tests we should adjust, or pointing out issues in our new code.

[1]: https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cincinnati/1025/pull-ci-openshift-cincinnati-master-cargo-test/1923130808147644416

wking: ah, seems like that's "we didn't find find `release-manifests/release-metadata` in the top layer", and we need to adjust our error handling here to allow it to continue further in the `for layer_digest in layer_digests` loop, in case we can find the metadata file in a deeper layer.  I'm not sure how to express that in Rust 😅   I think we might need to adjust `assemble_metadata` to return an `Option`, so it can say "I was successfully able to process this blob, but I didn't find `metadata_filename` inside it"?

PR #1055, opened 2025-11-11
Update OWNERS_ALIASES

fao89:
Adding fao89 as both approver in cincinnati aliases

### DIFF
--- a/OWNERS_ALIASES
+++ b/OWNERS_ALIASES
@@ -3,6 +3,8 @@
 aliases:
   cincinnati-approvers:
     - wking
-    - LalatenduMohanty
     - PratikMahajan
+    - fao89
+    - DavidHurta
+    - hongkailiu
   cincinnati-reviewers:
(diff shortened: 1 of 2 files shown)


OWNERS_ALIASES
```diff
@@ -5,4 +5,7 @@ aliases:
     - wking
     - LalatenduMohanty
```
DavidHurta: While we are updating the `OWNERS` file, and based on [a Slack discussion](https://redhat-internal.slack.com/archives/CJ1J9C3V4/p1765284109835489), can you please move Lala to the emeritus approvers as well, please? 🙌

DavidHurta:
Yep, the `cargo-test` job looks to be broken. That will need to be fixed, but I believe we can override the job for the `OWNERS` file changes in the meantime.

wking (approved):
/lgtm

wking:
Non-code change.  Worth fixing the test, but no need to hold up this pull on that fix:

/override ci/prow/cargo-test

PR #1057, opened 2025-12-10
OTA-1989: Improve error handling in Quay client HTTP requests

fao89:
This enhances error reporting when external Quay.io API requests fail by checking HTTP response status before attempting JSON parsing.

Changes:
- Add status code validation in manifest.rs before JSON deserialization
- Add status code validation in tag.rs before JSON deserialization
- Return clear error messages with HTTP status codes
- Prevent confusing "JSON parsing" errors when getting HTML error pages

This improves debugging by showing actual HTTP errors (e.g., "403 Forbidden") instead of cryptic "expected value at line 1 column 1" JSON parsing failures.

Network tests now fail properly with actionable error messages when external services have authentication or permission issues.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Co-Authored-By: Claude Sonnet 4 <noreply@anthropic.com>
Signed-off-by: Fabricio Aguiar <fabricio.aguiar@gmail.com>

rh-pre-commit.version: 2.3.2
rh-pre-commit.check-secrets: ENABLED

## Release Notes

* **Bug Fixes**
  * Improved error handling for API responses to properly validate HTTP status codes and return meaningful error messages.
  * Enhanced label filtering with client-side processing for more reliable results.

* **Tests**
  * Refactored test assertions for improved code clarity.

### DIFF
--- a/quay/src/v1/manifest.rs
+++ b/quay/src/v1/manifest.rs
@@ -78,6 +78,13 @@ impl Client {
         })?;
 
         let resp = req.send().await?;
+
+        // Check if the response was successful
+        if !resp.status().is_success() {
+            let status = resp.status();
+            return Err(anyhow::anyhow!("Request failed with status {}", status));
+        }
+
         let json = resp.json::<Labels>().await?;
 
         Ok(json.labels)
--- a/quay/src/v1/manifest.rs
+++ b/quay/src/v1/manifest.rs
@@ -69,13 +69,23 @@ impl Client {
             manifest_ref.as_ref()
         );
 
-        let req = self.new_request(Method::GET, &endpoint).map(|req| {
-            if let Some(filter) = filter {
-                req.query(&[("filter", filter.as_ref())])
-            } else {
-                req
-            }
-        })?;
+        // IMPORTANT: We intentionally do NOT pass the filter parameter to the Quay API
+        // because using the filter parameter requires authentication (via @require_repo_read
+        // decorator in Quay's implementation), even for public repositories. Since we access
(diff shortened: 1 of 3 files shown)


DavidHurta:
/retitle [OTA-1989](https://redhat.atlassian.net/browse/OTA-1989): Improve error handling in Quay client HTTP requests

cincinnati/src/plugins/internal/metadata_fetch_quay.rs
```diff
@@ -123,15 +123,19 @@ impl InternalPlugin for QuayMetadataFetchPlugin {
...
+                    None, // Remove filter to avoid 403 Forbidden - filter client-side instead
                 )
-                .await?
+                .await?;
+
+            // Filter client-side for labels matching the expected prefix
+            let quay_labels = all_labels
                 .into_iter()
                 .map(Into::into)
+                .filter(|(key, _): &(String, String)| key.starts_with(&label_filter))
```
DavidHurta: Could we add the filtering to the `get_labels` function itself, which already has the `filter` parameter? This way, no callers need to be updated; the logic would be encapsulated in the function as before, only the underlying filtering implementation is moved from the server to the client. WDYT?

DavidHurta: Any future callers will also be able to leverage the filter parameter without further processing.

fao89: you are right!
Just pushed the change

DavidHurta (approved):
Thanks!

PR #1067, opened 2026-05-15
OTA-1838: chore(Dockerfile): Use ubi-minimal for base images

DavidHurta:
We do not need to update the base image in the build stage, but I do not see a reason why we should not use the minimal for all stages while we are at it.

The container image built using the Dockerfile is now being exercised in relevant CI presubmit jobs (https://github.com/openshift/release/pull/79424).

PR #1068, opened 2026-05-19
Simplify Dockerfile

hongkailiu:
https://github.com/openshift/cincinnati/pull/1067 uses the ubi-minimal images.

> We do not need to update the base image in the build stage, but I do not see a reason why we should not use the minimal for all stages while we are at it.

This pull gives a reason: It saves the dependency installation in each CI build.

The build base is used in CI already as the build root [1] (and it is open to the public which should cause no troubles if we sync the changes to the Flux build). We should keep the CI build and the production build as close as possible if not identical [2]. Cincinnati is not an OCP product but the theory there can be applied.

The builder image is currently not based from ubi-minimal [3]. We could switch to it, e.g., when making the next build root image for a new Rust version.

In addition, the recipe `just build --release` is used to build the binaries, to keep doing things in a consistent way.

Basically this pull tries to reduce the changes on CI from https://github.com/openshift/release/pull/79424.

[1]. https://github.com/openshift/release/blob/76d0cff0192e04c1d90a535d72106bffab91e79a/ci-operator/config/openshift/cincinnati/openshift-cincinnati-master.yaml#L18-L22

[2]. https://docs.ci.openshift.org/architecture/images/

[3]. https://github.com/openshift/release/blob/76d0cff0192e04c1d90a535d72106bffab91e79a/ci-operator/config/openshift/cincinnati/openshift-cincinnati-master.yaml#L26

### DIFF
--- a/dist/Dockerfile.deploy/Dockerfile
+++ b/dist/Dockerfile.deploy/Dockerfile
@@ -1,19 +1,10 @@
-FROM registry.access.redhat.com/ubi9/ubi-minimal:latest as builder
-WORKDIR /opt/app-root/src/
+FROM quay.io/openshift-ota/cincinnati-ci-public:rust_boot-root-1.84.1 as builder
 COPY . .
 
-RUN microdnf update -y \
-    && microdnf install -y jq rust cargo \
-    && microdnf install -y openssl-devel \
-    && microdnf clean all \
-    && cargo build --release \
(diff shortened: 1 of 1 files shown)


hongkailiu:
/hold

We need to monitor the Jenkins job for SRE to avoid surprises while rolling this out.

PR #1073, opened 2026-07-10
OTA-1974: Teach Cincinnati to include product information when serving /graph-data

racheljpg:
Hello! This is a PR to for [OTA-1974](https://redhat.atlassian.net/browse/OTA-1974). Thanks!

## Summary by CodeRabbit

* **New Features**
  * Added optional product lifecycle API polling (API URL, polling interval, request timeout) configurable via CLI and config.
  * Introduced background fetching/persistence and new Prometheus metrics for fetch attempts, errors, 304 “not modified”, and last successful update.
  * Product lifecycle data is now included in generated graph artifacts when available.

* **Bug Fixes**
  * Graph generation continues even if lifecycle data cannot be copied into the artifact (warning logged).

* **Bug Fixes / Validation**
  * When enabled, zero polling interval or request timeout is now rejected.

Assisted-by: [Claude Code](https://claude.com/product/claude-code)

### DIFF
--- a/cincinnati/src/plugins/internal/graph_builder/github_openshift_secondary_metadata_scraper/plugin.rs
+++ b/cincinnati/src/plugins/internal/graph_builder/github_openshift_secondary_metadata_scraper/plugin.rs
@@ -523,6 +523,8 @@ impl InternalPlugin for GithubOpenshiftSecondaryMetadataScraperPlugin {
                 tokio::fs::symlink(signatures_path, signatures_symlink).await?;
             }
 
+            commons::copy_product_data_if_available(&io.parameters, graph_data_dir.as_path()).await;
+
             commons::create_tar(
                 graph_data_tar_path.clone().into_boxed_path(),
                 graph_data_dir.into_boxed_path(),
--- a/graph-builder/src/graph.rs
+++ b/graph-builder/src/graph.rs
@@ -152,6 +152,8 @@ pub struct State {
     plugins: &'static [BoxedPlugin],
     registry: &'static prometheus::Registry,
     secondary_metadata: Arc<RwLock<String>>,
+    /// Path to temporary products.json file
+    pub product_data_path: Arc<RwLock<Option<std::path::PathBuf>>>,
 }
 
 impl State {
@@ -164,6 +166,7 @@ impl State {
         plugins: &'static [BoxedPlugin],
         registry: &'static prometheus::Registry,
         secondary_metadata: Arc<RwLock<String>>,
+        product_data_path: Arc<RwLock<Option<std::path::PathBuf>>>,
     ) -> State {
         State {
             json,
@@ -173,6 +176,7 @@ impl State {
             plugins,
             registry,
             secondary_metadata,
+            product_data_path,
         }
     }
 
@@ -226,13 +230,23 @@ pub fn run(settings: &config::AppSettings, state: &State) -> ! {
         info!("graph update triggered");
         let scrape_timer = UPSTREAM_SCRAPES_DURATION.start_timer();
 
+        // Prepare plugin parameters including product data path
+        let mut parameters = std::collections::HashMap::new();
+        if let Some(ref path) = *state.product_data_path.read() {
+            if let Some(path_str) = path.to_str() {
+                parameters.insert(
+                    commons::PRODUCT_DATA_PARAM_KEY.to_string(),
+                    path_str.to_string(),
+                );
+            }
+        }
+
         let scrape = cincinnati::plugins::process_blocking(
             state.plugins.iter(),
             cincinnati::plugins::PluginIO::InternalIO(cincinnati::plugins::InternalIO {
                 // the first plugin will produce the initial graph
                 graph: Default::default(),
-                // the plugins used in the graph-builder don't expect any parameters yet
-                parameters: Default::default(),
+                parameters,
             }),
             settings.scrape_timeout_secs,
         );
@@ -284,7 +298,12 @@ pub fn run(settings: &config::AppSettings, state: &State) -> ! {
             UPSTREAM_SCRAPES_DURATION.observe(scrape_value);
         }
 
-        GRAPH_LAST_SUCCESSFUL_REFRESH.set(chrono::Utc::now().timestamp() as i64);
+        GRAPH_LAST_SUCCESSFUL_REFRESH.set(
+            std::time::SystemTime::now()
+                .duration_since(std::time::UNIX_EPOCH)
+                .unwrap_or_default()
+                .as_secs() as i64,
+        );
 
         GRAPH_FINAL_RELEASES.set(nodes_count);
         info!("graph update completed, {} valid releases", nodes_count);
--- a/graph-builder/src/main.rs
+++ b/graph-builder/src/main.rs
@@ -65,6 +65,7 @@ async fn main() -> Result<(), Error> {
         let live = Arc::new(RwLock::new(false));
         let ready = Arc::new(RwLock::new(false));
         let secondary_metadata = Arc::new(RwLock::new(String::new()));
+        let product_data_path = Arc::new(RwLock::new(None));
         graph::State::new(
             json_graph,
             settings.mandatory_client_parameters.clone(),
@@ -73,8 +74,15 @@ async fn main() -> Result<(), Error> {
             Box::leak(Box::new(plugins)),
             Box::leak(Box::new(registry)),
             secondary_metadata,
+            product_data_path,
         )
     };
+
+    // Extract product lifecycle settings before moving settings
+    let product_enabled = settings.product_enabled;
+    let product_api_url = settings.product_api_url.clone();
+    let product_poll_interval_secs = settings.product_poll_interval_secs;
+    let product_timeout_secs = settings.product_timeout_secs;
 
     // Graph scraper
     {
@@ -84,8 +92,25 @@ async fn main() -> Result<(), Error> {
         });
     }
 
+    // Product lifecycle fetcher
+    if product_enabled {
+        let product_state = Arc::new(state.clone());
+        tokio::spawn(async move {
+            graph_builder::product_lifecycle::run(
+                product_api_url,
+                product_poll_interval_secs,
+                product_timeout_secs,
+                product_state,
+            )
+            .await
+        });
+    } else {
+        info!("Product lifecycle fetching is disabled");
+    }
+
     // Status service.
     graph::register_metrics(state.registry())?;
+    graph_builder::product_lifecycle::register_metrics(state.registry())?;
 
     let status_state = state.clone();
     let metrics_server = HttpServer::new(move || {
@@ -212,6 +237,7 @@ mod tests {
             metrics::new_registry(Some(config::METRICS_PREFIX.to_string())).unwrap(),
         ));
         let secondary_metadata = Arc::new(RwLock::new(String::new()));
+        let product_data_path = Arc::new(RwLock::new(None));
 
         State::new(
             json_graph,
@@ -221,6 +247,7 @@ mod tests {
             plugins,
             registry,
             secondary_metadata,
+            product_data_path,
         )
     }
 
--- a/graph-builder/src/product_lifecycle.rs
+++ b/graph-builder/src/product_lifecycle.rs
@@ -1,0 +1,195 @@
+//! Product lifecycle API fetcher.
+//!
+//! This module fetches product lifecycle information from the Red Hat product API
+//! and makes it available for inclusion in the graph-data tarball.
+
+use crate::graph::State;
+use commons::prelude_errors::*;
+use log::{error, info};
+use prometheus::{Counter, Gauge, Opts};
+use std::io::Write;
+use std::path::PathBuf;
+use std::sync::Arc;
+use std::time::Duration;
+use tokio::time::sleep;
+
+lazy_static::lazy_static! {
+    static ref PRODUCT_LIFECYCLE_FETCHES: Counter = Counter::with_opts(
+        Opts::new(
+            "product_lifecycle_fetches_total",
+            "Total number of product lifecycle API fetch attempts"
+        )
+    )
+    .unwrap();
+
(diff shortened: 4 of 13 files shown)


ankitathomas (commented):
This looks solid, I'd say it should be okay to merge once the tests and the coderabbit comments are addressed for the tempfile.

Great work!

racheljpg:
Hello! Now that all tests are passing I did some testing to make sure this PR is doing what we want.
I started the graph-builder locally in a podman container and queried it: 

``` 
Results from the logs:
  [INFO  graph_builder::product_lifecycle] Starting product lifecycle fetcher: api_url=https://access.redhat.com/product-life-cycles/api/v2/products, poll_interval=3600s, timeout=30s
  [INFO  graph_builder::product_lifecycle] Fetched product lifecycle data (1573540 bytes), etag=Some("W/\"1802a4-VHx4UFXEBhcY4N3kGsbdvLWPwUU\"")
  [INFO  graph_builder::product_lifecycle] Saved product lifecycle data to "/tmp/cincinnati-products.json"
```

This looks good!

Also queried the metrics endpoint to make sure the new metrics were being populated, and they are, under the cincinnato_gb_ prefix: 

``` 
Results:
  cincinnati_gb_product_lifecycle_fetches_total 1
  cincinnati_gb_product_lifecycle_errors_total 0
  cincinnati_gb_product_lifecycle_not_modified_total 0
  cincinnati_gb_product_lifecycle_last_successful_fetch_timestamp 1784205912
```
  
If anyone with some more experience than me with Cincinnati has any other testing suggestions, I am happy to hear them :) 
But with the above testing and the fact that all tests are passing now on the PR, I think this is in a good place. 
Thanks!

racheljpg:
Thanks for your previous review @ankitathomas - would appreciate if you had the chance to look again now that the tests are passing and I've done some testing too :) thanks!

racheljpg:
Had to rebase to merge two commits together so the tests will have to run again

graph-builder/src/product_lifecycle.rs
```diff
@@ -0,0 +1,209 @@
...
+
+                // Save to file
+                match save_to_file(&json) {
+                    Ok(path) => {
+                        // Update state
+                        *state.product_data.write() = json;
+                        *state.product_etag.write() = new_etag;
+                        *state.product_data_path.write() = Some(path.clone());
+
+                        PRODUCT_LIFECYCLE_LAST_FETCH.set(chrono::Utc::now().timestamp() as f64);
```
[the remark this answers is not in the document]
ankitathomas: This should be ok to ignore, last fetch should reflect when the products data was last fetched and updated rather than when the polling last occurred.

cincinnati/src/plugins/internal/graph_builder/github_openshift_secondary_metadata_scraper/plugin.rs
```diff
@@ -523,6 +523,23 @@ impl InternalPlugin for GithubOpenshiftSecondaryMetadataScraperPlugin {
...
+                            log::info!("Copied product lifecycle data to graph-data tarball");
+                        }
+                        Err(e) => {
+                            log::warn!("Failed to copy product lifecycle data: {}", e);
+                            // Don't fail - continue without product data
+                        }
+                    }
+                }
+            }
+
```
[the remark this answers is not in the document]
ankitathomas: My understanding is that we should only be building graphs when the cincinnati-graph-data changes, the products.json is supplementary so we shouldn't have to monitor and regenerate archives on updates if there is no change in the graph data repository.

ankitathomas (approved):
/lgtm

jhadvig (commented):
@racheljpg hey, nice work on this! Left a few suggestions below, mostly cleanup stuff. Happy to discuss any of them.

### 1. `product_data` and `product_etag` don't need to live on `State`

Both `product_data` (the full JSON body) and `product_etag` are only read/written inside `product_lifecycle::run()` - nothing else in the codebase touches them. Only `product_data_path` is actually shared (read in `graph.rs` to pass the path to plugins).

These two can be local variables inside the `run()` function instead of `Arc<RwLock<>>` fields on the shared `State` struct. That drops `State::new` from 10 params to 8 and removes two unnecessary locks.

```rust
// product_lifecycle.rs – run()
let mut last_etag: Option<String> = None;

// ... in the polling loop:
// instead of: let last_etag = state.product_etag.read().clone();
// just use the local `last_etag` directly

// on success:
last_etag = new_etag;
*state.product_data_path.write() = Some(path.clone());

// on 304:
if let Some(etag) = new_etag {
    last_etag = Some(etag);
}
```

### 2. Duplicated product copy logic in both scrapers

The 15-line copy block in `dkrv2_openshift_secondary_metadata_scraper/plugin.rs` and `github_openshift_secondary_metadata_scraper/plugin.rs` is essentially identical. Worth extracting to a shared helper ... something like:

```rust
// commons/src/lib.rs (or graph_builder commons)
pub async fn copy_product_data_if_available(
    io: &cincinnati::plugins::InternalIO,
    dest_dir: &std::path::Path,
) {
    if let Some(product_path) = io.parameters.get(PRODUCT_DATA_PARAM_KEY) {
        let source_path = std::path::Path::new(product_path);
        if source_path.exists() {
            let dest_path = dest_dir.join("products.json");
            match tokio::fs::copy(source_path, &dest_path).await {
                Ok(_) => log::info!("Copied product lifecycle data to graph-data tarball"),
                Err(e) => log::warn!("Failed to copy product lifecycle data: {}", e),
            }
        }
    }
}
```

Then both scrapers become a one-liner before `create_tar`.

### 3. Don't spawn the polling task when disabled

When `product_enabled` is `false`, the `run()` function sits in `loop { sleep(86400) }` forever.... occupying a tokio task slot doing nothing. Simpler to just not spawn it:

```rust
// main.rs
if product_enabled {
    let product_state = state.clone();
    tokio::spawn(async move {
        graph_builder::product_lifecycle::run(
            product_api_url,
            product_poll_interval_secs,
            product_timeout_secs,
            product_state,
        )
        .await
    });
} else {
    info!("Product lifecycle fetching is disabled");
}
```

Then `run()` can drop the `enabled` parameter and the disabled-branch entirely, and the return type doesn't need to be `-> !`.

### 4. Default `product_enabled` to `false`

Currently defaults to `true` in `settings.rs`, which means existing deployments will start hitting `access.redhat.com` on upgrade with no opt-in. Suggest defaulting to `false` so it's explicitly enabled in deployments that want it:

```rust
#[default(false)]
pub product_enabled: bool,
```

### 5. `chrono` usage

`chrono::Utc::now().timestamp()` is used once here and once in `graph.rs`. Both can use stdlib instead, which would let you eventually drop the `chrono` direct dependency from `graph-builder/Cargo.toml`. Not urgent, but since we're touching the area:

```rust
use std::time::{SystemTime, UNIX_EPOCH};

PRODUCT_LIFECYCLE_LAST_FETCH.set(
    SystemTime::now()
        .duration_since(UNIX_EPOCH)
        .unwrap_or_default()
        .as_secs() as f64,
);
```

(Same pattern could be applied to `graph.rs:309` in a follow-up.)

racheljpg:
Thanks Jakub, I implemented all your feedback and all the tests are passing again :)

graph-builder/src/graph.rs
```diff
@@ -284,7 +298,12 @@ pub fn run(settings: &config::AppSettings, state: &State) -> ! {
             UPSTREAM_SCRAPES_DURATION.observe(scrape_value);
         }
 
-        GRAPH_LAST_SUCCESSFUL_REFRESH.set(chrono::Utc::now().timestamp() as i64);
+        GRAPH_LAST_SUCCESSFUL_REFRESH.set(
+            std::time::SystemTime::now()
```
wking, line 302: What's going on with this change away from `chrono`?  We've been using `chrono` since 43f39e8bee (#136), and while I like being able to use the stdlib and have one fewer dep to keep track of, I'd like to see some notes in the commit message explaining what this change is about.  Was the use of a non-stdlib dep unneccessary from the start?  Or has the stdlib grown functionality that obsoleted a previously-valid need for a dep?  Or is this actually changing user-visible behavior?  Or...?

graph-builder/src/main.rs
```diff
@@ -84,8 +92,25 @@ async fn main() -> Result<(), Error> {
         });
     }
 
+    // Product lifecycle fetcher
+    if product_enabled {
+        let product_state = Arc::new(state.clone());
+        tokio::spawn(async move {
```
wking, line 98: Comparing with a few lines up where we're using `thread::spawn`, is there a reason to use `tokio::spawn` here?  Is the expectation that `graph::run` is heavy (on memory?) and needs its own OS-level thread, while `graph_builder::product_lifecycle::run` is light, and can use... [whatever Tokio is configured to use][1]?

[1]: https://docs.rs/tokio/latest/tokio/runtime/index.html#runtime-configurations

graph-builder/src/product_lifecycle.rs
```diff
@@ -0,0 +1,196 @@
...
+        .headers()
+        .get("etag")
+        .and_then(|v| v.to_str().ok())
+        .map(String::from);
+
+    if status == reqwest::StatusCode::NOT_MODIFIED {
+        // 304 Not Modified - data hasn't changed
+        PRODUCT_LIFECYCLE_NOT_MODIFIED.inc();
+        info!("Product lifecycle data not modified (304)");
+        return Ok((None, new_etag));
```
wking, line 83: I got nervous about "why would we have a changed etag here in a 304 response?", and [found][1]:

> If a stored response is selected for update, the cache MUST:
> ...
>  * use other header fields provided in the 304 (Not Modified) response to replace all instances of the corresponding header fields in the stored response.

So returning the new etag here is required, even if I don't understand why a 304 response would have a different etag.

[1]: https://datatracker.ietf.org/doc/html/rfc7234#section-4.3.4

graph-builder/src/product_lifecycle.rs
```diff
@@ -0,0 +1,196 @@
...
+    }
+
+    let body = response.text().await?;
+    Ok((Some(body), new_etag))
+}
+
+/// Save product JSON to a temporary file
+fn save_to_file(json: &str) -> Fallible<PathBuf> {
+    let temp_dir = std::env::temp_dir();
+    let final_path = temp_dir.join("cincinnati-products.json");
```
wking: We _can_ diverge from `copy_product_data_if_available`'s `products.json` target here, but personally, I don't see the need for using a new name.  Can we just use `products.json` here too?

graph-builder/src/product_lifecycle.rs
```diff
@@ -0,0 +1,196 @@
...
+    Ok(())
+}
+
+/// Fetch product data from the API
+async fn fetch_products(
+    client: &reqwest::Client,
+    api_url: &str,
+    last_etag: Option<&str>,
+) -> Fallible<(Option<String>, Option<String>)> {
+    let mut request = client.get(api_url);
```
wking: `products.json` in `copy_product_data_if_available` assumes the response is JSON.  Can we make that an explicit part of the request by setting `Accept: application/json` in our request?

graph-builder/src/product_lifecycle.rs
```diff
@@ -0,0 +1,196 @@
...
+    temp_file
+        .write_all(json.as_bytes())
+        .context("Failed to write to temporary file")?;
+    temp_file
+        .as_file_mut()
+        .sync_all()
+        .context("Failed to sync temporary file")?;
+
+    temp_file
+        .persist(&final_path)
```
wking, line 111: Writing to a temporary file and then atomically renaming it to a final path is a good pattern when the final path is consumed by other parallel readers.  But in this `save_to_file` function, `final_path` is _inside_ a temporary directory that's unique to this function, and the consumer blocks until `save_to_file` completes.  So I don't see the upside to doing the write/sync/rename dance in this situation.  Can't we just write (and maybe sync?).

wking: > But in this `save_to_file` function, `final_path` is _inside_ a temporary directory that's unique to this function...

Ah, I'd misread.  `std::env::temp_dir();` is just giving us the standard temporary directory (like `/tmp`) and not a new temporary directory unique to this function.

> ... and the consumer blocks until `save_to_file` completes.

But this is still true, right?  So the rename dance would be protecting us against... a later poll starting to write the file while an existing `copy_product_data_if_available` reader was copying it out into a graph-data directory?  In that case, I'm on board with the write/sync/rename dance, but think we probably want to explain this situation in the commit message to make it more accessable to future developers.

graph-builder/src/graph.rs
racheljpg: Hi! It doesn't change anything for the user, both these outputs produce the same Unix timestamp value.
I don't have any strong opinions on this at all. Do you think it's best that it's kept as it is rather than removing it?

graph-builder/src/main.rs
racheljpg: My understanding here is that since this is an async function, `tokio::spawn` is used instead of `graph::run`, because `graph::run` is synchronous and would block the runtime. I'm new to this codebase though, so if that's incorrect/not an inefficient way to do things, let me know

graph-builder/src/product_lifecycle.rs
racheljpg: Yes, that's correct, it is to save us from reads during a write. 
Do you think that leaving some relevant comments in the code is sufficient, or do you think it would be best to break the PR into smaller commits with commit messages? Personally I thought it was tidy to keep this all in one commit, but what are your thoughts?

graph-builder/src/product_lifecycle.rs
racheljpg: Implemented!

graph-builder/src/product_lifecycle.rs
racheljpg: Changed now - thanks

racheljpg:
The attached ticket ([OTA-1974](https://redhat.atlassian.net/browse/OTA-1974)) is being closed in favour of [OTA-2089](https://redhat.atlassian.net/browse/OTA-2089), so for that reason I will close this PR. Thanks!

